PR #18, opened 2021-06-30
Sort TopK results by docId

ashtul:
related #19


DvirDukhan (changes requested):
please have it as a dedicated function on the API, meaning the flow should be something like:
```c
VecSimQueryResult*  res = VecSimIndex_TopKQuery(index, query);
res = VecSimIndex_SortResByID(res);
```

ashtul:
While it can be done on RediSearch side, this is a driver to connect the two libraries so it should return the results as RediSearch expects them.

DvirDukhan:
closing as it is done by #25

PR #33, opened 2021-08-09
Add hnsw with new deletion

alonre24:
Removed `hnswlib` submodule, and add the necessary parts from it it as part of this repo (surrounded with `hnswlib` namespace).
The integrated version of `hnswlib` contains in-place delete of vectors (instead of the mark delete of the original version)


src/python_bindings/bindings.cpp
```diff
@@ -46,12 +49,45 @@ class PyVecSimIndex {
...
 
+    VecSimIndex *getIndex() { return index; }
+
     virtual ~PyVecSimIndex() { VecSimIndex_Free(index); }
 
   private:
     VecSimIndex *index;
 };
 
+class PyHNSWLibIndex {
```
DvirDukhan, line 60: just do public inheritance from PyVecSimIndex

src/python_bindings/bindings.cpp
```diff
@@ -46,12 +49,45 @@ class PyVecSimIndex {
...
+
+    py::object knn(py::object input, size_t k, size_t ef) {
+        VecSimQueryParams queryParams = {.hnswRuntimeParams = {.efRuntime = ef}};
+        return hnsw_index->knn(input, k, &queryParams);
+    }
+
+    size_t indexSize() { return hnsw_index->indexSize(); }
+
+    // This function is unique for hnsw_index pointer only
+    void setDefaultEf(size_t ef) { HNSWLib_SetQueryParam(hnsw_index->getIndex(), ef); }
```
DvirDukhan, line 85: this is the only function needs to be here

src/python_bindings/bindings.cpp
```diff
@@ -86,10 +122,25 @@ PYBIND11_MODULE(VecSim, m) {
...
 
+    py::class_<VecSimQueryParams>(m, "VecSimQueryParams").def(py::init());
+
     py::class_<PyVecSimIndex>(m, "VecSimIndex")
         .def(py::init([](const VecSimParams &params) { return new PyVecSimIndex(params); }),
              py::arg("params"))
         .def("add_vector", &PyVecSimIndex::addVector)
-        .def("knn_query", &PyVecSimIndex::knn, py::arg("vector"), py::arg("id"))
+        .def("delete_vector", &PyVecSimIndex::deleteVector)
+        .def("knn_query", &PyVecSimIndex::knn, py::arg("vector"), py::arg("id"),
```
DvirDukhan, line 132: this should be `k` insted of `id` here

DvirDukhan (changes requested):

src/VecSim/algorithms/hnsw/visited_list_pool.h
```diff
@@ -0,0 +1,77 @@
...
+
+    ~VisitedList() { delete[] mass; }
+};
+///////////////////////////////////////////////////////////
+//
+// Class for multi-threaded pool-management of VisitedLists
+//
+/////////////////////////////////////////////////////////
+
+class VisitedListPool {
```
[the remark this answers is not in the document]
alonre24: This is hnsw implementation for multi-threaded insert/search. So actually we don't need the pool logic in single-threaded mode, I'll change the code so that it will be under ENABLE_PARALLELISATION build.

DvirDukhan (approved):

PR #44, opened 2021-09-01
fixed empty index bug

DvirDukhan:
This PR fixed a bug on HNSW wrapper. Now the wrapper creating the `VecSimQueryResult` array by iterate according to the query output size and not `k`.


ashtul (approved):
:+1:

tests/unit/test_bruteforce.cpp
```diff
@@ -429,3 +429,55 @@ TEST_F(BruteForceTest, brute_force_vector_search_test_blocksize_1) {
...
+        bfParams : {initialCapacity : 200},
+        type : VecSimType_FLOAT32,
+        size : 4,
+        metric : VecSimMetric_L2,
+        algo : VecSimAlgo_BF
+    };
+    size_t n = 100;
+    size_t k = 11;
+    VecSimIndex *index = VecSimIndex_New(&params);
+
```
alonre24, line 444: Why not adding a search here instead of having another test?

tests/unit/test_hnswlib.cpp
```diff
@@ -453,3 +453,55 @@ TEST_F(HNSWLibTest, test_query_runtime_params_user_build_args) {
...
+    VecSimParams params = {
+        bfParams : {initialCapacity : 200},
+        type : VecSimType_FLOAT32,
+        size : 4,
+        metric : VecSimMetric_L2,
+        algo : VecSimAlgo_HNSWLIB
+    };
+    size_t n = 100;
+    size_t k = 11;
+    VecSimIndex *index = VecSimIndex_New(&params);
```
alonre24, line 467: Why not adding a search here instead of having another test?

alonre24 (approved):

PR #46, opened 2021-09-01
fix inf bug

DvirDukhan:
This PR handles a bug where `inf` is a value of comparing two vectors

### DIFF
--- a/src/VecSim/algorithms/brute_force.cpp
+++ b/src/VecSim/algorithms/brute_force.cpp
@@ -118,6 +118,22 @@ static void BruteForce_UpdateVector(BruteForceIndex *bfIndex, idType id, const v
     memcpy(destinaion, vector_data, bfIndex->base.dim);
 }
 
+static size_t BruteForce_IndexOfMin(float *scores, size_t n) {
+    float max_float = std::numeric_limits<float>::max();
+    float inf = numeric_limits<float>::infinity();
+    float min = std::numeric_limits<float>::max();
+    size_t min_index = 0;
+    for (size_t i = 0; i < n; i++) {
+        if (scores[i] == max_float)
+            continue;
+        if (scores[i] < min || scores[i] == inf) {
+            min = scores[i];
+            min_index = i;
+        }
+    }
+    return min_index;
+}
+
 static void VectorBlock_AddVector(VectorBlock *vectorBlock, VectorBlockMember *vectorBlockMember,
                                   const void *vectorData, size_t vectorDim) {
     // Mutual point both structs on each other.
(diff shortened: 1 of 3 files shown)


src/VecSim/algorithms/brute_force.cpp
```diff
@@ -118,6 +118,22 @@ static void BruteForce_UpdateVector(BruteForceIndex *bfIndex, idType id, const v
...
 
+static size_t BruteForce_IndexOfMin(float *scores, size_t n) {
+    float max_float = std::numeric_limits<float>::max();
+    float inf = numeric_limits<float>::infinity();
+    float min = std::numeric_limits<float>::max();
+    size_t min_index = 0;
+    for (size_t i = 0; i < n; i++) {
+        if (scores[i] == max_float)
+            continue;
+        if (scores[i] < min || scores[i] == inf) {
```
alonre24, line 129: Why we set `scores[i]` as the minimum score if `scores[i] == inf`?

tests/unit/test_bruteforce.cpp
```diff
@@ -465,3 +465,32 @@ TEST_F(BruteForceTest, brute_force_search_empty_index) {
     VecSimQueryResult_Free(res);
     VecSimIndex_Free(index);
 }
+
+TEST_F(BruteForceTest, brute_force_test_RS_integration_test) {
+    VecSimParams params = {
```
alonre24, line 470: What does this test do? How does this scenario related to `inf` score?

tests/unit/test_hnswlib.cpp
```diff
@@ -487,3 +487,32 @@ TEST_F(HNSWLibTest, hnsw_search_empty_index) {
...
     VecSimIndex_Free(index);
 }
+
+TEST_F(HNSWLibTest, hnsw_test_RS_integration_test) {
+    VecSimParams params = {
+        bfParams : {initialCapacity : 4},
+        type : VecSimType_FLOAT32,
+        size : 2,
+        metric : VecSimMetric_L2,
+        algo : VecSimAlgo_HNSWLIB
```
alonre24, line 497: Why do you send `bfParams` when `algo` is `VecSimAlgo_HNSWLIB`?

src/VecSim/algorithms/brute_force.cpp
```diff
@@ -247,7 +263,7 @@ extern "C" VecSimQueryResult *BruteForce_TopKQuery(VecSimIndex *index, const voi
         std::fill_n(scores, bfIndex->vectorBlockSize, 1.0);
         bfIndex->distanceCalculationFunction(dim, vectorBlock, queryBlob, scores);
         for (int i = 0; i < MIN(vectorBlock->size, k); i++) {
-            size_t min_index = cblas_ismin(vectorBlock->size, scores, 1);
+            size_t min_index = BruteForce_IndexOfMin(scores, vectorBlock->size);
```
ashtul, line 266: Can we run this outside of the loop?

tests/unit/test_bruteforce.cpp
DvirDukhan: added comment

src/VecSim/algorithms/brute_force.cpp
DvirDukhan: this is awkward logic:
since there is an `OR` evaluation here, if the first condition is `true`, so the entire condition is `true`. The first test is for actual numbers. Actual numbers can be either the minimal number in the array, or not.
For a number smaller from the current `min`, the LHS of the condition is `true`, otherwise, the entire condition is `false`.
After a number is found is the current minimum fo the array, it is replaced with `std::numeric_limits<float>::max()`. Once we replaced all normal numbers we are left with `inf` or `std::numeric_limits<float>::max()`, so in case we have `inf`, RHS of this condition will hold, and will be replaced with `std::numeric_limits<float>::max()`

src/VecSim/algorithms/brute_force.cpp
DvirDukhan: I'm trying to come up with a better approach. For now, we can't do it outside the loop as we need top-k

PR #49, opened 2021-09-13
Build modifications for RediSearch

rafie:


src/CMakeLists.txt
```diff
@@ -1,14 +1,22 @@
...
+	set(binroot ${root}/build)
+endif()
+message("VectorSimilarity: " ${binroot})
+
 option(VECSIM_STATIC "Build as static library" OFF)
 
 include_directories(${CMAKE_CURRENT_SOURCE_DIR})
-include_directories(${CMAKE_CURRENT_SOURCE_DIR}/../deps)
+
+include_directories(${root}/deps)
```
DvirDukhan, line 19: I think we can remove it, no? The library is now self contained

rafie: you do have pybind11 there.
anyway, it will support anything that's added under `deps/` in the future.

DvirDukhan (approved):

PR #59, opened 2021-10-11
main library is built with x86-64

DvirDukhan:


src/CMakeLists.txt
```diff
@@ -32,7 +32,7 @@ endif()
 
 add_subdirectory(VecSim/spaces)
 
-add_compile_options(-march=native)
+add_compile_options(-march=x86-64)
```
rafie, line 35: just to be sure: it this better than not specifying `-march` at all?

PR #65, opened 2021-10-26
Redis_memory

DvirDukhan:
This PR introduces the ability for the Vector similarity library to have its own allocator. This allows the library to choose in compile time if it will use `malloc` or the Redis allocator as its allocator of choice, wrapped with `vecsim_malloc` functionality.
As such, in `vecsim_malloc.h` file there are 3 new additions:
1. `vecsim_malloc` function family where `*` is in `{malloc, calloc, realloc, strdup, strndup, free}`
2. `VecsimAllocator`: an object that uses `vecsim_malloc` function family and can be use in order to allocate other objects (see next)
3. 'VecsimSTLAllocator': a wrapper template around `VecsimAllocator` that meant to be used on STL data structures, which support custom allocators out of the box.

An additional file is `vecsim_base.h` which introduces a base object, that all the objects in the vector similarity library should inherit from. The base object must be allocated with `VecsimAllocator` and own a `VecsimAllocator`. This will allow to new objects to share their allocator with nested objects. The idea is that the highest level object (index) can report its memory consumption. a `VecSimIndex` object can report its consumption by the `info` command (preparation for `FT.INFO`).

Additional file is `vecsim_stl.h` which is aliasing `std` namespace STL datastructure such as `vector`, `set` etc.. that are set to use 'VecsimSTLAllocator'.

This PR allows 3rd party memory functions to replace default `malloc`, `calloc`, `free`, `realloc` and `strdup` via the function `VecSim_SetMemoryFunctions`


tests/unit/test_allocator.cpp
```diff
@@ -0,0 +1,182 @@
...
+    NestedObject(VecSimAllocator *allocator)
+        : VecsimBaseObject(allocator), stl_object(allocator), simpleObject(allocator){};
+    NestedObject(std::shared_ptr<VecSimAllocator> allocator)
+        : VecsimBaseObject(allocator), stl_object(allocator), simpleObject(allocator){};
+};
+
+TEST_F(AllocatorTest, test_simple_object) {
+    std::shared_ptr<VecSimAllocator> allocator = std::make_shared<VecSimAllocator>();
+    SimpleObject *obj = new (allocator) SimpleObject(allocator);
+    ASSERT_GE(allocator->getAllocationSize(), sizeof(SimpleObject) + sizeof(VecSimAllocator));
```
alonre24, line 49: Why `ASSERT_GE`? Can't we know the expected allocation exactly?

tests/unit/test_allocator.cpp
```diff
@@ -0,0 +1,182 @@
...
+    info = hnswIndex->info();
+    ASSERT_EQ(allocator->getAllocationSize(), info.memory);
+    current_memory = info.memory;
+
+    hnswIndex->addVector(vec, 2);
+    ASSERT_GE(allocator->getAllocationSize(), current_memory);
+    info = hnswIndex->info();
+    ASSERT_EQ(allocator->getAllocationSize(), info.memory);
+
+    // TODO: commented out since hnsw does not recalim memory
```
alonre24, line 169: What do you mean by "hnsw does not reclaim memory"?

alonre24 (commented):
Looks good! Left some comments. Also:
- Why did you copy `redismodule.h` into the project `src`? I thought that it should only be part of the tests... 
- I guess that all of the methods of `VecsimBaseObject` are required for the custom allocator mechanism, but couldn't exactly understand why. Consider adding some more detailed descriptions.

src/VecSim/algorithms/brute_force/vector_block.cpp
```diff
@@ -1,17 +1,24 @@
...
-    this->vectors = new float[blockSize * vectorSize];
+VectorBlockMember::VectorBlockMember(std::shared_ptr<VecSimAllocator> allocator)
+    : VecsimBaseObject(allocator) {}
+
+VectorBlock::VectorBlock(size_t blockSize, size_t vectorSize,
+                         std::shared_ptr<VecSimAllocator> allocator)
+    : VecsimBaseObject(allocator), dim(vectorSize), size(0), blockSize(blockSize) {
+    this->members =
+        (VectorBlockMember **)this->allocator->allocate(sizeof(VectorBlockMember *) * blockSize);
+    this->vectors = (float *)this->allocator->allocate(sizeof(float) * blockSize * vectorSize);
```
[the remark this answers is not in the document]
DvirDukhan: I didn't override the system `new` operator. There is a need to allocated a new `float` array, and it is needed to be done explicitly by vec sim allocator

src/VecSim/algorithms/hnsw/hnswlib.h
```diff
@@ -94,17 +99,19 @@ class HierarchicalNSW {
     labeltype *getExternalLabelPtr(tableint internal_id) const;
     char *getDataByInternalId(tableint internal_id) const;
     int getRandomLevel(double reverse_size);
-    std::set<tableint> *getIncomingEdgesPtr(tableint internal_id, int level) const;
+    std::set<tableint, std::less<tableint>, VecsimSTLAllocator<tableint>> *
+    getIncomingEdgesPtr(tableint internal_id, int level) const;
```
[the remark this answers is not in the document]
DvirDukhan: STL definition of set is 
```
template<
    class Key,
    class Compare = std::less<Key>,
    class Allocator = std::allocator<Key>
> class set;
```
when needed to specify an allocator, the compare also needs to be specify

src/VecSim/algorithms/hnsw/hnswlib.h
```diff
@@ -94,17 +99,19 @@ class HierarchicalNSW {
...
     unsigned short int getListCount(const linklistsizeint *ptr) const;
     void setListCount(linklistsizeint *ptr, unsigned short int size);
-    void removeExtraLinks(linklistsizeint *node_ll, CandidatesQueue<dist_t> candidates,
-                          size_t Mcurmax, tableint *node_neighbors,
-                          const std::set<tableint> &orig_neighbors, tableint *removed_links,
-                          size_t *removed_links_num);
+    void removeExtraLinks(
+        linklistsizeint *node_ll, CandidatesQueue<dist_t> candidates, size_t Mcurmax,
+        tableint *node_neighbors,
+        const std::set<tableint, std::less<tableint>, VecsimSTLAllocator<tableint>> &orig_neighbors,
```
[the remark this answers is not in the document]
DvirDukhan: will do

src/VecSim/memory/vecsim_malloc.h
```diff
@@ -0,0 +1,94 @@
...
+static inline void *vecsim_malloc(size_t n) { return RedisModule_Alloc(n); }
+static inline void *vecsim_calloc(size_t nelem, size_t elemsz) {
+    return RedisModule_Calloc(nelem, elemsz);
+}
+static inline void *vecsim_realloc(void *p, size_t n) { return RedisModule_Realloc(p, n); }
+static inline void vecsim_free(void *p) { RedisModule_Free(p); }
+static inline char *vecsim_strdup(const char *s) { return RedisModule_Strdup(s); }
+
+static inline char *vecsim_strndup(const char *s, size_t n) {
+    char *ret = (char *)rm_malloc(n + 1);
```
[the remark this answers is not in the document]
DvirDukhan: wrong copy-paste, good catch

tests/unit/test_allocator.cpp
DvirDukhan: when you delete a vector from hnsw, not all the memory that was allocated per this vector is reclaimed

tests/unit/test_allocator.cpp
DvirDukhan: in this case you comment is right, good catch
however when using STL, this is not always hold as STL implementation should be a "black box" and we cannot assume anything regarding exact allocation

src/VecSim/memory/vecsim_malloc.h
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    T *allocate(size_t size) { return (T *)this->vecsim_allocator->allocate(size * sizeof(T)); }
+
+    void deallocate(T *ptr, size_t size) {
+        this->vecsim_allocator->deallocate(ptr, size * sizeof(T));
+    }
+};
+
+template <class T, class U>
```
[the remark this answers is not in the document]
DvirDukhan: if you have a two instances of VecSimSTL allocator, they should be considered equal, as they are just allocating, so they should be template type agnoistic. This is why the copy constructor and assignment operator takes `U` as type template, instead of the template's definition `T`.  I took this as from an example for using custom allocators https://en.cppreference.com/w/cpp/named_req/Allocator

src/VecSim/memory/vecsim_malloc.h
```diff
@@ -0,0 +1,94 @@
...
+bool operator==(const VecsimSTLAllocator<T> &, const VecsimSTLAllocator<U> &) {
+    return true;
+}
+template <class T, class U>
```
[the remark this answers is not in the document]
DvirDukhan: same as above

tests/unit/test_allocator.cpp
```diff
@@ -150,19 +204,21 @@ TEST_F(AllocatorTest, test_hnsw) {
 
     float vec[128] = {};
     HNSWIndex *hnswIndex = new (allocator) HNSWIndex(&params, allocator);
-    ASSERT_GE(allocator->getAllocationSize(), sizeof(VecSimAllocator) + sizeof(HNSWIndex));
+    expectedAllocationSize +=
+        sizeof(BruteForceIndex) + sizeof(InnerProductSpace) + 2 * vecsimAllocationOverhead;
```
alonre24, line 208: Should be `sizeof(HNSWIndex)` no?

tests/unit/test_allocator.cpp
```diff
@@ -150,19 +204,21 @@ TEST_F(AllocatorTest, test_hnsw) {
...
+    ASSERT_GE(allocator->getAllocationSize(), expectedAllocationSize);
```
alonre24, line 209: Can we use `ASSERT_EQ` here, since the capacity is 0?

alonre24 (commented):
Liked the new `vecsim_stl` namespace! Also, tests are much better now in my opinion. Only few more comments

tests/unit/test_allocator.cpp
DvirDukhan: nope, as there are STLs being build as well as memory allocations happens in HNSW Ctor

alonre24 (approved):

src/VecSim/algorithms/brute_force/brute_force.cpp
```diff
@@ -74,20 +77,22 @@ int BruteForceIndex::addVector(const void *vector_data, size_t label) {
...
         // No vector blocks, create new one.
-        vectorBlock = new VectorBlock(this->vectorBlockSize, this->dim);
+        vectorBlock =
+            new (this->allocator) VectorBlock(this->vectorBlockSize, this->dim, this->allocator);
         this->vectorBlocks.push_back(vectorBlock);
     } else {
         // Get the last vector block.
         vectorBlock = this->vectorBlocks[this->vectorBlocks.size() - 1];
-        if (vectorBlock->getSize() == this->vectorBlockSize) {
+        if (vectorBlock->getLength() == this->vectorBlockSize) {
```
MeirShpilraien, line 86: Any reason to change this?

src/VecSim/memory/redismodule.h
```diff
@@ -0,0 +1,1483 @@
+#ifndef REDISMODULE_H
```
MeirShpilraien, line 1: Why do you need redismodule.h here?

tests/module/memory_test.c
```diff
@@ -0,0 +1,213 @@
+
+#include "redismodule.h"
```
MeirShpilraien, line 2: Why do you need a module to test this functionality? I would argue that redismodule.h should not even be in this library.

MeirShpilraien (commented):
I still now follow why you need redismodule.h at all here, why you need a module to test this functionality?

src/VecSim/memory/vecsim_malloc.h
```diff
@@ -63,11 +42,77 @@ struct VecSimAllocator {
...
+
+/**
+ * @brief Global function copy a string (strdup style).
+ *
+ * @param s Source string.
+ * @return char* Copied string.
+ */
+inline char *vecsim_strdup(const char *s) {
+    return VecSimAllocator::memFunctions.strdupFunction(s);
+}
```
alonre24, line 97: Why did you decide that for `vecsim_strdup` you'll have a `strdupFunction`, but `vecsim_strndup` is using `vecsim_malloc`? Consider making it consistent (I would go for using `vecsim_malloc` in both)

src/VecSim/memory/vecsim_malloc.h
```diff
@@ -1,44 +1,23 @@
...
-
-#endif
 
 struct VecSimAllocator {
+    // Allow global vecsim memory functions to access this class.
+    friend inline void *vecsim_malloc(size_t n);
+    friend inline void *vecsim_calloc(size_t nelem, size_t elemsz);
+    friend inline void *vecsim_realloc(void *p, size_t n);
+    friend inline void vecsim_free(void *p);
+    friend inline char *vecsim_strdup(const char *s);
```
alonre24, line 13: Why are those declared as `friend` if they are part of the class? They can access the inner fields of `VecSimAllocator` anyway no?

alonre24 (commented):
The change from static do dynamic linkage of the allocator callbacks looks good and clean! Added 2 comments

alonre24 (approved):

alonre24 (approved):

PR #67, opened 2021-10-30
c++ implementation

DvirDukhan:
This PR moves the internal implementation from C style struct with function pointers to C++ style implementation (abstract class) 

enabler for #65


src/VecSim/algorithms/brute_force/brute_force.h
```diff
@@ -0,0 +1,34 @@
...
+
+private:
+    void updateVector(idType id, const void *vector_data);
+    std::unordered_map<labelType, idType> labelToIdLookup;
+    std::vector<VectorBlockMember *> idToVectorBlockMemberMapping;
+    std::set<idType> deletedIds;
+    std::vector<VectorBlock *> vectorBlocks;
+    size_t vectorBlockSize;
+    idType count;
+    std::unique_ptr<SpaceInterface<float>> space;
```
alonre24, line 32: Perhaps you can move `space` and `dist_func` to the base `VecSimIndex` class, as they are common to all inherited indices tyeps

src/VecSim/algorithms/brute_force/vector_block.cpp
```diff
@@ -0,0 +1,26 @@
...
+void VectorBlock::addVector(VectorBlockMember *vectorBlockMember, const void *vectorData) {
+    // Mutual point both structs on each other.
+    this->members[this->size] = vectorBlockMember;
+    vectorBlockMember->block = this;
+    vectorBlockMember->index = this->size;
+
+    // Copy vector data and update block size.
+    memcpy(this->vectors + (this->size * this->dim), vectorData, this->dim * sizeof(float));
+    this->size++;
+}
```
alonre24, line 26: new line

src/VecSim/algorithms/brute_force/vector_block.h
```diff
@@ -0,0 +1,44 @@
...
+
+typedef size_t labelType;
+typedef size_t idType;
+
+// Pre declaration
+
+struct VectorBlock;
+
+struct VectorBlockMember {
+public:
```
alonre24, line 12: The `public` is not required, since in struct all fields are public by default...

src/VecSim/algorithms/brute_force/vector_block.h
```diff
@@ -0,0 +1,44 @@
...
+
+    inline size_t getSize() { return size; }
+
+    inline VectorBlockMember *getMember(size_t index) { return this->members[index]; }
+
+    inline void setMember(size_t index, VectorBlockMember *member) {
+        this->members[index] = member;
+    }
+
+    virtual ~VectorBlock();
```
alonre24, line 37: Why `virtual`?

src/VecSim/vec_sim_index.h
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * @brief Abstract C++ class for vector index, delete and lookup
+ *
+ */
+class VecSimIndex {
+protected:
+    size_t dim;
+    VecSimType vecType;
+    VecSimMetric metric;
```
alonre24, line 14: Why are they `protected` and not `private`? You have getters to access these fields

src/VecSim/algorithms/hnsw/hnswlib_c.h
```diff
@@ -1,34 +1,24 @@
...
+                                             VecSimQueryParams *queryParams) override;
+    virtual VecSimIndexInfo info() override;
+    virtual VecSimBatchIterator *newBatchIterator(const void *queryBlob) override;
+
+    void setEf(size_t ef);
+
+private:
+    std::unique_ptr<SpaceInterface<float>> space;
+    hnswlib::HierarchicalNSW<float> hnsw;
+};
```
alonre24, line 24: new line

src/VecSim/algorithms/hnsw/hnswlib_c.cpp
```diff
@@ -12,59 +11,44 @@
...
-
-    VecSimIndex base;
-    unique_ptr<SpaceInterface<float>> space;
-    HierarchicalNSW<float> hnsw;
-};
-
-#ifdef __cplusplus
-extern "C" {
-#endif
+/******************** Ctor / Dtor **************/
```
alonre24, line 14: There is no destructor implementation here, it means that the default distructor is what we need here?

alonre24 (commented):
Nice! added some questions

src/VecSim/algorithms/brute_force/vector_block.h
DvirDukhan: for good practice. See https://www.geeksforgeeks.org/virtual-destructor/

src/VecSim/vec_sim_index.h
DvirDukhan: for convenience only

src/VecSim/algorithms/brute_force/brute_force.h
DvirDukhan: recall that we want to fall back to brute force once the new API will take place (allowing L2 search on IP HNSW index for example). as such, I don't see any need for brute force to have space/dist_func, so I want to keep state as-is and modify it after.

src/VecSim/algorithms/brute_force/vector_block.h
DvirDukhan: I know. readiablity and good practices >> common knowledge

src/VecSim/algorithms/hnsw/hnswlib_c.cpp
DvirDukhan: it is just a comment for this section, so in the future, when needed we can put it here.
we can use `pragma region` also but I'm not sure it is supported everywhere so I didn't use it.

src/VecSim/algorithms/brute_force/vector_block.h
alonre24: But `VectorBlock` is not a base class...

src/VecSim/vec_sim_index.h
alonre24: Don't you think that it is better use `private` instead? So we won't bread abstraction

alonre24 (approved):

PR #69, opened 2021-11-02
Fix_hnsw_empty_index

DvirDukhan:


tests/unit/test_hnswlib.cpp
```diff
@@ -516,3 +516,31 @@ TEST_F(HNSWLibTest, hnsw_inf_score) {
     runTopKSearchTest(index, "abcdefgh", k, verify_res);
     VecSimIndex_Free(index);
 }
+
+TEST_F(HNSWLibTest, test_empty_index) {
```
alonre24, line 520: Consider maybe adding this case to `hnsw_search_empty_index` (set `initial_capacity=0`) instead of having a new test which is very similar to others...

alonre24 (approved):

PR #71, opened 2021-11-07
Bf batch iterator

alonre24:
Implementation for brute force batch iterator. 

This PR introduces an abstract `batch_iterator` class for searching `top k` results in batches. 
BF iterator supports two search methods: heap based search (generalisation of the current `topK` search), and select-based search, which uses `select` algorithm. There is a trade-off between the two approaches, and the current implementation uses the following heuristics:
if the number of results to return is lower than 0.1% of the number of remaining vectors in the index - use heap-based search. Otherwise, use select based search (this is based on these experiments https://redislabs.atlassian.net/wiki/spaces/AI/pages/3080781915/Iterators+Design)


src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
```diff
@@ -0,0 +1,104 @@
...
+        scores[coordinates.first].erase(scores[coordinates.first].begin() +
+                                        (int)coordinates.second);
+    }
+    return results;
+}
+
+BF_BatchIterator::BF_BatchIterator(const void *query_vector, const BruteForceIndex *bf_index)
+    : VecSimBatchIterator(query_vector), id(BF_BatchIterator::next_id) {
+    index = bf_index;
+    BF_BatchIterator::next_id++;
```
DvirDukhan, line 58: can you do all of it in the initialization list?
also, see if you can pre-allocate `scores` to the right size (avoid re allocation while appendeing)

src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
```diff
@@ -0,0 +1,104 @@
...
+                labelType label = blocks[i]->getMember(j)->label;
+                labelToScoreCoordinates[label] = {i, j};
+            }
+            scores.push_back(block_scores);
+        }
+    }
+    // heuristics: decide if using heap or select search, based on the ratio between the
+    // number of remaining results and the index size.
+    VecSimQueryResult *results;
+    if ((getIndex()->indexSize() - getResultsCount()) / 1000 > n_res) {
```
DvirDukhan, line 83: please have the heuristic calculation in a dedicated function

src/VecSim/algorithms/brute_force/vector_block.h
```diff
@@ -34,6 +50,11 @@ struct VectorBlock {
         this->members[index] = member;
     }
 
+    std::vector<std::pair<float, labelType>> ComputeScores(DISTFUNC<float> DistFunc,
```
DvirDukhan, line 53:
```suggestion
    std::vector<std::pair<float, labelType>> computeBlockScores(DISTFUNC<float> DistFunc,
```
keep camelCase please

src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+VecSimQueryResult *BF_BatchIterator::selectBasedSearch(size_t n_res) {
+    // Flatten the score matrix, keep the labels to have stable sorting.
+    vector<pair<float, labelType>> allScores;
+    for (auto vectorBlockScores : scores) {
+        allScores.insert(allScores.end(), vectorBlockScores.begin(), vectorBlockScores.end());
+    }
+
+    auto n_th_element_pos = allScores.begin() + (int)n_res;
+    std::nth_element(allScores.begin(), n_th_element_pos, allScores.end());
```
DvirDukhan, line 40: please add comments. I guess this is finding the pivot, right?

src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
```diff
@@ -0,0 +1,104 @@
...
+    }
+    return results;
+}
+
+VecSimQueryResult *BF_BatchIterator::selectBasedSearch(size_t n_res) {
+    // Flatten the score matrix, keep the labels to have stable sorting.
+    vector<pair<float, labelType>> allScores;
+    for (auto vectorBlockScores : scores) {
+        allScores.insert(allScores.end(), vectorBlockScores.begin(), vectorBlockScores.end());
+    }
```
DvirDukhan, line 37: I believe we can avoid this by keeping scores as a flat list, to begin with. Let's discuss it face to face

src/VecSim/algorithms/brute_force/vector_block.h
```diff
@@ -34,6 +50,11 @@ struct VectorBlock {
...
+                                                           const void *queryBlob);
+
+    void heapBasedSearch(const std::vector<std::pair<float, labelType>> &scores, float &upperBound,
```
DvirDukhan, line 56: should be a part of a vector block?

DvirDukhan (commented):
looks really good
few comments

src/VecSim/utils/arr_cpp.h
```diff
@@ -50,7 +50,7 @@ T *array_grow(T *arr, size_t cap) {
 
 template <typename T>
 T *array_append(T *arr, T val) {
-    arr = array_grow(arr, ++array_hdr(arr)->len);
+    arr = array_grow(arr, array_hdr(arr)->len);
```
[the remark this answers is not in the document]
alonre24: There was a bug,  because in `array_grow` we do `++array_hdr(arr)->len`, so we actually increased the array length by 2 instead of 1

src/VecSim/query_result_struct.h
```diff
@@ -4,7 +4,7 @@
 #include <limits>
 
 #define INVALID_ID    -1
-#define INVALID_SCORE std::numeric_limits<float>::min()
+#define INVALID_SCORE std::numeric_limits<float>::infinity()
```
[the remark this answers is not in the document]
alonre24: When I thought of invalidating scores that has already returned in some iteration, I wanted to use this macro. However, we always consider the lower score as the better, so I thought that the invalid one should be the one that is _larger_ than all the rest.

src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
alonre24: Right, and this also make a "partial" in-place sorting: all the elements to the left of the n-th element are smaller than it, and all the elements to the right are larger. I'll add a documentation.

src/VecSim/algorithms/brute_force/vector_block.h
alonre24: It was designed that way to allow consolidation with the original `TopK` that also uses heap search

alonre24: It was designed that way to allow consolidation with the original `TopK` that also uses heap search

src/VecSim/query_result_struct.h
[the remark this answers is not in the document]
alonre24: used `std::numeric_limits<float>::quiet_NaN()`

src/VecSim/algorithms/brute_force/vector_block.h
```diff
@@ -34,6 +50,11 @@ struct VectorBlock {
...
+                         size_t nRes, CandidatesHeap &candidates);
```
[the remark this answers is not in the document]
alonre24: Added for `computeBlockScores`, removed `heapBasedSearch`

src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
alonre24: Added `scores.reserve(index_size)` call in the first iteration of `getNextResults`

src/VecSim/algorithms/brute_force/vector_block.h
alonre24: removed

src/CMakeLists.txt
```diff
@@ -52,6 +55,6 @@ install(TARGETS VectorSimilaritySpaces
     DESTINATION ${CMAKE_INSTALL_PREFIX})
 
 if(NOT VECSIM_STATIC)
-    set_target_properties(VectorSimilarity PROPERTIES PREFIX "")
+    set_target_properties(VectorSimilarity PROPERTIES PREFIX "lib")
```
DvirDukhan, line 58: why the add?

src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
```diff
@@ -0,0 +1,139 @@
...
+    return a.first < b.first;
+}
+
+// heuristics: decide if using heap or select search, based on the ratio between the
+// number of remaining results and the index size.
+VecSimQueryResult *BF_BatchIterator::searchByHeuristics(size_t n_res,
+                                                        VecSimQueryResult_Order order) {
+    if ((getIndex()->indexSize() - getResultsCount()) / 1000 > n_res) {
+        return heapBasedSearch(n_res);
```
DvirDukhan, line 29: you are missing ordering here

src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include <cmath>
+#include <functional>
+#include "bf_batch_iterator.h"
+#include "VecSim/utils/arr_cpp.h"
+#include "VecSim/utils/vec_utils.h"
+#include "VecSim/query_result_struct.h"
+
+unsigned char BF_BatchIterator::next_id = 0;
+
+// Compare function that handle NaN values (consider these as the highest, above inf)
```
DvirDukhan, line 13:
```suggestion
// Compare function that handles NaN values (consider this as the highest, above inf: NaN>inf)
```

src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
```diff
@@ -0,0 +1,139 @@
...
+    for (size_t i = 0; i < n_res; i++) {
+        if (isnan(scores[i].first)) {
+            continue;
+        }
+        results = array_append(results, VecSimQueryResult{});
+        VecSimQueryResult_SetId(results[array_len(results) - 1], scores[i].second);
+        VecSimQueryResult_SetScore(results[array_len(results) - 1], scores[i].first);
+    }
+    // Erase the returned results slice from the scores array
+    scores.erase(scores.begin(), n_th_element_pos);
```
DvirDukhan, line 97: didn't we say we are not erasing but moving the start of the array to `n_th_element_pos`?

src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
```diff
@@ -0,0 +1,139 @@
...
+    BF_BatchIterator::next_id++;
+}
+
+VecSimQueryResult_List BF_BatchIterator::getNextResults(size_t n_res,
+                                                        VecSimQueryResult_Order order) {
+    assert((order == BY_ID || order == BY_SCORE) &&
+           "Possible order values are only 'BY_ID' or 'BY_SCORE'");
+    // Only in the first iteration we need to compute all the scores
+    if (scores.empty() && getResultsCount() == 0) {
+        scores.reserve(getIndex()->indexSize());
```
DvirDukhan, line 112: please let's use the convention that everything that is owned/called by the local instance, will be called from `this`
so `this->getIndex()->indexSize()`.
Please maintain it throughout the file

src/VecSim/algorithms/brute_force/brute_force.cpp
```diff
@@ -147,44 +140,36 @@ int BruteForceIndex::deleteVector(size_t label) {
...
-            scores[i] = this->dist_func(vectorBlock->getVector(i), queryBlob, &dim);
-        }
-        size_t vec_count = vectorBlock->getSize();
-        for (int i = 0; i < vec_count; i++) {
-
-            if (knn_res.size() < k) {
-                labelType label = vectorBlock->getMember(i)->label;
-                knn_res.emplace(scores[i], label);
-                upperBound = knn_res.top().first;
+        std::vector<std::pair<float, labelType>> scores =
```
DvirDukhan, line 151: let's discuss this change

src/VecSim/query_result_struct.h
```diff
@@ -4,7 +4,7 @@
 #include <limits>
 
 #define INVALID_ID    -1
-#define INVALID_SCORE std::numeric_limits<float>::min()
+#define INVALID_SCORE std::numeric_limits<float>::quiet_NaN()
```
DvirDukhan, line 7: please add a comment

tests/unit/CMakeLists.txt
```diff
@@ -55,3 +55,7 @@ target_link_libraries(test_bruteforce PUBLIC gtest_main VectorSimilarity)
 include(GoogleTest)
 gtest_discover_tests(test_hnswlib)
 gtest_discover_tests(test_bruteforce)
+
+find_package(benchmark REQUIRED)
```
DvirDukhan, line 59: where is the benchmark package coming from?

src/CMakeLists.txt
alonre24: When I used scratch file for the benchmark, I couldn't link our VectorSimilarity shared library from command line. Then I understood from here https://www.cprogramming.com/tutorial/shared-libraries-linux-gcc.html that GCC assumes that all libraries start with lib and end with .so

src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
alonre24: What do you mean? the heap based search always returns the results sorted by their scores.

tests/unit/CMakeLists.txt
alonre24: Should be a dependency if we'll decide using google-benchmark

src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
alonre24: You're right. Changing this

src/VecSim/algorithms/brute_force/brute_force.cpp
```diff
@@ -147,44 +140,43 @@ int BruteForceIndex::deleteVector(size_t label) {
...
-            if (knn_res.size() < k) {
-                labelType label = vectorBlock->getMember(i)->label;
-                knn_res.emplace(scores[i], label);
-                upperBound = knn_res.top().first;
+        for (int i = 0; i < scores.size(); i++) {
+            // Always choose the current candidate if we have less than k.
+            if (TopCandidates.size() < k) {
+                TopCandidates.emplace(scores[i], vectorBlock->getMember(i)->label);
+                upperBound = TopCandidates.top().first;
+                // otherwise, try greedily to improve the top candidates with a vector that
```
DvirDukhan, line 162: `Otherwise`
also, move this comment to the `else` clause please

src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
```diff
@@ -0,0 +1,146 @@
...
+    // This will perform an in-place partition of the elements in the slice of the array that
+    // contains valid results, based on the n-th element as the pivot - every element with a lower
+    // will be placed before it, and all the rest will be placed after.
+    std::nth_element(valid_begin_it, n_th_element_pos, this->scores.end(), cmp);
+
+    auto *results = array_new<VecSimQueryResult>(n_res);
+    for (size_t i = this->scores_valid_start_pos; i < this->scores_valid_start_pos + n_res; i++) {
+        if (isnan(this->scores[i].first)) {
+            continue;
+        }
```
DvirDukhan, line 95: is this a valid scenario?

DvirDukhan (commented):
looks good, minor comment

src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
alonre24: Made an update on Thursday night that makes this check no longer required (+ improving performance for heap based search)

DvirDukhan (approved):

PR #72, opened 2021-11-11
Add benchmark tool

alonre24:
Add google benchmark to our tests (also in CI). Currently we use it to test BF batch iterator (this is waiting for  #71 to be merged).
By running `make benchmark` one can reproduce the benchmarks locally.


.circleci/config.yml
```diff
@@ -49,6 +49,10 @@ commands:
           name: Valgrind
           command: bash -l -c "make valgrind"
 
+      - run:
+          name: Benchmark
+          command: bash -l -c "make benchmark"
+
```
DvirDukhan, line 55: CI for sanity?

Makefile
```diff
@@ -93,3 +93,8 @@ format:
 
 lint:
 	./clang-check-all.sh
+
+benchmark:
+	./tests/benchmark.sh
+
+.PHONY: benchmark
```
DvirDukhan, line 100: please add it also to the help section

tests/unit/CMakeLists.txt
```diff
@@ -54,4 +54,4 @@ target_link_libraries(test_bruteforce PUBLIC gtest_main VectorSimilarity)
 
 include(GoogleTest)
 gtest_discover_tests(test_hnswlib)
-gtest_discover_tests(test_bruteforce)
+gtest_discover_tests(test_bruteforce)
```
DvirDukhan, line 57: empty line?

tests/benchmark/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+if(NOT CMAKE_BUILD_TYPE)
+    set(CMAKE_BUILD_TYPE Release)
+endif()
+
+include(FetchContent)
+FetchContent_Declare(
+        googletest
+        # Specify the commit you depend on and update it regularly.
+        URL https://github.com/google/googletest/archive/609281088cfefc76f9d0ce82e1ff6c30cc3591e5.zip
+)
```
DvirDukhan, line 29: I know google test is a dependency for google benchmark, but is this fetch required?

tests/benchmark/bf_google_benchmark.cpp
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    void TearDown(const ::benchmark::State &state) {}
+
+    ~BM_BatchIteratorBF() { VecSimIndex_Free(bf_index); }
+};
+
+BENCHMARK_DEFINE_F(BM_BatchIteratorBF, get_10000_total_results)(benchmark::State &st) {
+    for (auto _ : st) {
+        size_t n_res = st.range(0);
```
DvirDukhan, line 53: you can move `n_res` outside the benchmark scope

tests/benchmark/bf_google_benchmark.cpp
```diff
@@ -0,0 +1,74 @@
...
+            }
+        }
+        VecSimBatchIterator_Free(batchIterator);
+    }
+}
+
+// Register the function as a benchmark
+BENCHMARK_REGISTER_F(BM_BatchIteratorBF, get_10000_total_results)
+    ->Arg(100)
+    ->Arg(1000)
```
DvirDukhan, line 70: why the second arg?

alonre24: For every `->Arg()` the benchmark runs with different configurations. We access the given arg in the benchmark from `st.range(0)`. So this benchmark runs twice, once with `n_res=100` (this will use heap search, according to the heuristics), and in the second time with `n_res=1000` (this will use select search)

tests/benchmark/bf_google_benchmark.cpp
alonre24: see https://github.com/RedisLabsModules/VectorSimilarity/pull/72#discussion_r748861306

.circleci/config.yml
alonre24: Yes, and also for regression (we can compare the results before and after introducing new features)

tests/benchmark/CMakeLists.txt
alonre24: I guess so, google-benchmark did not built without it..

tests/benchmark/bf_google_benchmark.cpp
DvirDukhan: I meant, you can access it outside the loop

DvirDukhan (approved):

PR #74, opened 2021-11-18
Redis memory test

GuyAv46:
added a module for testing that memory allocating in redis-server is preformed by the server allocator.
added automation for running the module and added it to tox tests for testing the allocating process from the CI

### DIFF
--- a/tests/module/flow/test_vecsim_malloc.py
+++ b/tests/module/flow/test_vecsim_malloc.py
@@ -1,0 +1,82 @@
+import redis
+
+from functools import wraps
+
+'''
+python -m RLTest --test test_vecsim_malloc.py --module path/to/memory_test.so
+'''
+
+def with_test_module(f):
+    @wraps(f)
+    def wrapper(env, *args, **kwargs):
+        con = env.getConnection()
+        modules = con.execute_command("MODULE", "LIST")
+        if b'VecSim_memory' in [module[1] for module in modules]:
+            return f(env, *args, **kwargs)
+        try:
+            ret = con.execute_command('MODULE', 'LOAD', 'memory_test.so')
+            env.assertEqual(ret, b'OK')
+        except Exception as e:
+            env.assertFalse(True)
+            env.debugPrint(str(e), force=True)
+            return
+        return f(env, *args, **kwargs)
+    return wrapper
+
+@with_test_module
+def test_basic_check(env):
+
+    con = env.getConnection()
+    ret = con.execute_command('VecSim_memory.basic_check')
+    env.assertEqual(ret, b'OK')
+
+@with_test_module
+def test_create_index_check(env):
+
+    con = env.getConnection()
+
+    ret = con.execute_command('VecSim_memory.create_index_check', 'BF')
+    env.assertEqual(ret, b'OK')
+    ret = con.execute_command('VecSim_memory.create_index_check', 'HNSW')
+    env.assertEqual(ret, b'OK')
+
+@with_test_module
+def test_create_index_add_check_bf(env):
+
+    con = env.getConnection()
+
+    for n in range(1, 100000, 1000):
+        ret = con.execute_command('VecSim_memory.create_index_add_n_check', 'BF', n)
+        env.assertEqual(ret, b'OK')
+    
+@with_test_module
+def test_create_index_add_check_hnsw(env):
+
+    con = env.getConnection()
+
+    for n in range(1, 100000, 1000):
+        ret = con.execute_command('VecSim_memory.create_index_add_n_check', 'HNSW', n)
+        env.assertEqual(ret, b'OK')   
+ 
+@with_test_module
+def test_create_index_add_del_check_bf(env):
+
+    con = env.getConnection()
+
+    for n in range(1, 100000, 1000):
+        ret = con.execute_command('VecSim_memory.create_index_add_n_delete_m_check', 'BF', n, n/2)
+        env.assertEqual(ret, b'OK')
+        ret = con.execute_command('VecSim_memory.create_index_add_n_delete_m_check', 'BF', n, n)
+        env.assertEqual(ret, b'OK')
+    
+@with_test_module
+def test_create_index_add_del_check_hnsw(env):
+
+    con = env.getConnection()
+
+    for n in range(1, 100000, 1000):
+        ret = con.execute_command('VecSim_memory.create_index_add_n_delete_m_check', 'HNSW', n, n/2)
+        env.assertEqual(ret, b'OK')
+        ret = con.execute_command('VecSim_memory.create_index_add_n_delete_m_check', 'HNSW', n, n)
+        env.assertEqual(ret, b'OK')
+    
--- a/tests/module/flow/test_vecsim_malloc.py
+++ b/tests/module/flow/test_vecsim_malloc.py
@@ -53,11 +53,11 @@ def test_create_index_add_check_bf(env):
 def test_create_index_add_check_hnsw(env):
 
     con = env.getConnection()
+    vectors = [1000, 100000]
 
-    ret = con.execute_command('VecSim_memory.create_index_add_n_check', 'HNSW', 1000)
-    env.assertEqual(ret, b'OK')
-    ret = con.execute_command('VecSim_memory.create_index_add_n_check', 'HNSW', 100000)
-    env.assertEqual(ret, b'OK')
+    for n in vectors:
+        ret = con.execute_command('VecSim_memory.create_index_add_n_check', 'HNSW', n)
+        env.assertEqual(ret, b'OK')
  
 @with_test_module
 def test_create_index_add_del_check_bf(env):
@@ -74,9 +74,8 @@ def test_create_index_add_del_check_bf(env):
 def test_create_index_add_del_check_hnsw(env):
 
     con = env.getConnection()
+    vectors = [1000, 100000]
 
-    ret = con.execute_command('VecSim_memory.create_index_add_n_delete_m_check', 'HNSW', 1000, 1000)
-    env.assertEqual(ret, b'OK')
-    ret = con.execute_command('VecSim_memory.create_index_add_n_delete_m_check', 'HNSW', 100000, 100000)
-    env.assertEqual(ret, b'OK')
-    
+    for n in vectors:
+        ret = con.execute_command('VecSim_memory.create_index_add_n_delete_m_check', 'HNSW', n, n)
+        env.assertEqual(ret, b'OK')
--- a/tests/module/flow/test_vecsim_malloc.py
+++ b/tests/module/flow/test_vecsim_malloc.py
@@ -24,7 +24,7 @@ def with_test_module(f):
     return wrapper
 
 @with_test_module
-def test_create_index_check(env):
+def test_create_bf_and_hnsw_indexes_check(env):
 
     con = env.getConnection()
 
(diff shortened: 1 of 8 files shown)


tests/module/flow/test_vecsim_malloc.py
```diff
@@ -0,0 +1,82 @@
...
+@with_test_module
+def test_create_index_add_check_hnsw(env):
+
+    con = env.getConnection()
+
+    ret = con.execute_command('VecSim_memory.create_index_add_n_check', 'HNSW', 1000)
+    env.assertEqual(ret, b'OK')
+    ret = con.execute_command('VecSim_memory.create_index_add_n_check', 'HNSW', 100000)
+    env.assertEqual(ret, b'OK')
+
```
DvirDukhan, line 61: you can merge both function into one:
```python
@with_test_module
def test_create_index_add_check(env):
    con = env.getConnection()
    algorithms = ['HNSW', 'BF']
    vectors = [1000, 100000]
    for algo in algorithms:
        for v in vectors:
            res = con.execute_command('VecSim_memory.create_index_add_n_check', algo, v)
            env.assertEqual(ret, b'OK')
```
if you decide not to do it, please stay consistent with iteration over an array of different amounts of vectors, as in the BF function

tests/module/flow/test_vecsim_malloc.py
```diff
@@ -0,0 +1,82 @@
...
+@with_test_module
+def test_create_index_add_del_check_hnsw(env):
+
+    con = env.getConnection()
+
+    ret = con.execute_command('VecSim_memory.create_index_add_n_delete_m_check', 'HNSW', 1000, 1000)
+    env.assertEqual(ret, b'OK')
+    ret = con.execute_command('VecSim_memory.create_index_add_n_delete_m_check', 'HNSW', 100000, 100000)
+    env.assertEqual(ret, b'OK')
+
```
DvirDukhan, line 82: note here: always add an empty line in the end of the file

tests/module/flow/test_vecsim_malloc.py
```diff
@@ -0,0 +1,82 @@
...
+    
+@with_test_module
+def test_create_index_add_del_check_hnsw(env):
+
+    con = env.getConnection()
+
+    ret = con.execute_command('VecSim_memory.create_index_add_n_delete_m_check', 'HNSW', 1000, 1000)
+    env.assertEqual(ret, b'OK')
+    ret = con.execute_command('VecSim_memory.create_index_add_n_delete_m_check', 'HNSW', 100000, 100000)
+    env.assertEqual(ret, b'OK')
```
DvirDukhan, line 81: same comment as above

tests/module/memory_test.c
```diff
@@ -0,0 +1,191 @@
+
+#include "VecSim/memory/redismodule.h"
+#include "VecSim/vec_sim.h"
+#include <stdbool.h>
+#include <errno.h>
+#include <string.h>
+
+#define DIMENSION 100
+
+// get memory usage from redis-server
```
DvirDukhan, line 10: note here: a comment sentence starts with a capital letter and ends with a period. please apply it to the entire file.

tests/module/memory_test.c
```diff
@@ -0,0 +1,191 @@
...
+}
+
+// adds 'amount' vectors to the index. could be 0
+void _add_vectors(VecSimIndex *index, long long amount) {
+    VecSimIndexInfo indexInfo = VecSimIndex_Info(index);
+    int i;
+    int64_t vec[indexInfo.d];
+    for (i = 0; i < indexInfo.d; i++)
+        vec[i] = i;
+    for (i = 0; i < amount; i++)
```
DvirDukhan, line 28: for the sake of consistency in our code, please don't re-use the iteration variables in different scope. 
have `for(int i=0; i < n; i++)`, unless you really must use shared variables

tests/module/memory_test.c
```diff
@@ -0,0 +1,191 @@
...
+    param.type = VecSimType_INT64;
+    param.size = DIMENSION;
+    param.metric = VecSimMetric_L2;
+
+    return VecSimIndex_New(&param);
+}
+
+// creates an index, add n vectors, deletes m vectors and then compares redis-server memory usage
+// report to the index memory usage report uncomment the 3 lines close to the end to get the values
+// along with the comparison result.
```
DvirDukhan, line 67: please rephrase this comment to be more readable.

tests/module/memory_test.c
```diff
@@ -0,0 +1,191 @@
...
+int VecSim_memory_create_index_add_n_delete_m_check(RedisModuleCtx *ctx, RedisModuleString **argv,
+                                                    int argc) {
+
+    long long n, m, i;
+    long long startMemory, endMemory;
+    VecSimAlgo algo = -1;
+    RedisModuleString *al;
+
+    if (argc != 4 || RedisModule_StringToLongLong(argv[2], &n) == REDISMODULE_ERR ||
+        RedisModule_StringToLongLong(argv[3], &m) == REDISMODULE_ERR) {
```
DvirDukhan, line 77: please split the conversion check from the arity check. the wrong arity reply is where the expected parameters count does not match what your function expects.
you should use RedisModule_ReplyWithError of any error scenario

tests/module/memory_test.c
```diff
@@ -0,0 +1,191 @@
...
+        algo = VecSimAlgo_BF;
+    RedisModule_FreeString(ctx, al);
+    al = RedisModule_CreateString(ctx, "HNSW", 4);
+    if (!RedisModule_StringCompare(argv[1], al))
+        algo = VecSimAlgo_HNSWLIB;
+    RedisModule_FreeString(ctx, al);
+    if (algo == -1) {
+        RedisModule_WrongArity(ctx);
+        return REDISMODULE_OK;
+    }
```
DvirDukhan, line 92: here, you can extract the cstring from `argv[1]` using `RedisModule_StringPtrLen` and just use `strcmp`, instead of creating and deleting RedisModuleStrings.

tests/module/memory_test.c
```diff
@@ -0,0 +1,191 @@
...
+// creates an index, add n vectors, deletes m vectors and then compares redis-server memory usage
+// report to the index memory usage report uncomment the 3 lines close to the end to get the values
+// along with the comparison result.
+int VecSim_memory_create_index_add_n_delete_m_check(RedisModuleCtx *ctx, RedisModuleString **argv,
+                                                    int argc) {
+
+    long long n, m, i;
+    long long startMemory, endMemory;
+    VecSimAlgo algo = -1;
+    RedisModuleString *al;
```
DvirDukhan, line 74: please declare variables as soon as you need them, not before.

tests/module/memory_test.c
```diff
@@ -0,0 +1,191 @@
...
+
+// calls to VecSim_memory_create_index_add_n_delete_m_check with m=0
+int VecSim_memory_create_index_add_n_check(RedisModuleCtx *ctx, RedisModuleString **argv,
+                                           int argc) {
+    int i, res;
+    RedisModuleString *newArgv[argc + 1];
+    for (i = 0; i < argc; i++)
+        newArgv[i] = argv[i];
+    newArgv[argc] = RedisModule_CreateString(ctx, "0", 1);
+    res = VecSim_memory_create_index_add_n_delete_m_check(ctx, newArgv, argc + 1);
```
DvirDukhan, line 125: I see what you meant here, but try do it cleaner. Consider having 
`VecSim_memory_check_impl` or something similar, where you can call it from both `VecSim_memory_create_index_add_n_delete_m_check` and `VecSim_memory_create_index_add_n_check` after you parsed `argv` and without creating new `RedisModuleString`

tests/module/memory_test.c
```diff
@@ -0,0 +1,191 @@
...
+}
+
+// calls to VecSim_memory_create_index_add_n_check with n=0
+int VecSim_memory_create_index_check(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
+    int i, res;
+    RedisModuleString *newArgv[argc + 1];
+    for (i = 0; i < argc; i++)
+        newArgv[i] = argv[i];
+    newArgv[argc] = RedisModule_CreateString(ctx, "0", 1);
+    res = VecSim_memory_create_index_add_n_check(ctx, newArgv, argc + 1);
```
DvirDukhan, line 137: the same comment as above. have a dedicated implementation function shared between those callbacks.

tests/module/memory_test.c
```diff
@@ -0,0 +1,191 @@
...
+    RedisModule_FreeString(ctx, newArgv[argc]);
+    return res;
+}
+
+int VecSim_memory_basic_check(RedisModuleCtx *ctx, RedisModuleString **argv, int argc) {
```
DvirDukhan, line 142: this is redundant function IMO. and it is not testing any memory consumption

tests/module/memory_test.c
```diff
@@ -0,0 +1,191 @@
...
+
+    endMemory = _get_memory_usage(ctx); // get memory usage after creating the index
+
+    VecSimIndexInfo indexInfo = VecSimIndex_Info(index);
+
+    // RedisModule_ReplyWithArray(ctx, 3);
+    // RedisModule_ReplyWithLongLong(ctx,endMemory-startMemory);
+    // RedisModule_ReplyWithLongLong(ctx,indexInfo.memory);
+
+    if (indexInfo.memory <= endMemory - startMemory)
```
DvirDukhan, line 108: This is probably the most important line in this file. Please add a comment on what is being tested here.

tests/module/memory_test.c
```diff
@@ -0,0 +1,191 @@
...
+        return REDISMODULE_ERR;
+
+    // usage: VecSim_memory.create_index_add_n_delete_m_check <algorithm:BF/HNSW> <number> <number>
+    if (RedisModule_CreateCommand(ctx, "VecSim_memory.create_index_add_n_delete_m_check",
+                                  VecSim_memory_create_index_add_n_delete_m_check, "", 0, 0,
+                                  0) == REDISMODULE_ERR)
+        return REDISMODULE_ERR;
+
+    return REDISMODULE_OK;
+}
```
DvirDukhan, line 191: empty line

DvirDukhan (changes requested):
very nice!
try to learn from the comments about our coding and comments style.

tests/module/flow/test_vecsim_malloc.py
GuyAv46: BF is much faster in these tests, so I allowed myself to run more checks on it.

GuyAv46: I think it is better to keep these tests separated. I'll make a vectors list in the hnsw test

tests/module/flow/test_vecsim_malloc.py
GuyAv46: line 82 is not empty?

tests/module/memory_test.c
GuyAv46: cool, much simpler

tests/module/memory_test.c
GuyAv46: You're right. This suppose to be a sanity check to verify we manage to create an index at all.
If you think this is useless I delete it.

DvirDukhan: yes, please remove this

DvirDukhan (changes requested):
looks good
small fix and we are good to merge

DvirDukhan (approved):

PR #76, opened 2021-11-19
Bindings_batch_iterator_from_index

DvirDukhan:


alonre24 (commented):
Thanks for this!! Now I am able to debug the C++ in the flow tests. I went over your changes and added some questions for me to better understand. Also, what was the reason that the previous code didn't work?

PR #77, opened 2021-12-03
CLang sanitizers revisited

rafie:


Makefile
```diff
@@ -49,52 +89,72 @@ all: build
...
+	@make -C build
 
 clean:
 ifneq ($(ALL),1)
-	make -C build clean
+	@make -C build clean
+	@make -C tests/unit/build clean
 else
-	rm -rf build
+	@rm -rf build tests/unit/build
```
DvirDukhan, line 116: also the build folder itself, no?

tests/unit/CMakeLists.txt
```diff
@@ -47,15 +46,12 @@ add_executable(test_hnswlib test_hnswlib.cpp test_utils.cpp)
...
-
-# add_test(NAME test_hnswlib COMMAND test_hnswlib)
-# add_test(NAME test_bruteforce COMMAND test_bruteforce)
-
 target_link_libraries(test_hnswlib PUBLIC gtest_main VectorSimilarity)
 target_link_libraries(test_bruteforce PUBLIC gtest_main VectorSimilarity)
 target_link_libraries(test_allocator PUBLIC gtest_main VectorSimilarity)
 
-include(GoogleTest)
+include(${root}/cmake/gtest.cmake)
```
DvirDukhan, line 53: please move the include up, after the project definition

src/VecSim/spaces/space_interface.h
```diff
@@ -1,6 +1,9 @@
+
```
DvirDukhan, line 1: remove empty line please

sbin/check-format.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+ROOT=$HERE/..
+READIES=$ROOT/deps/readies
+. $READIES/shibumi/functions
```
DvirDukhan, line 6: why do we need this?

DvirDukhan (commented):
thanks
see comments please. Looks good overall

.circleci/config.yml
```diff
@@ -15,85 +16,77 @@ commands:
...
+      - run:
+          name: System setup
+          command: ./sbin/system-setup.py
 
+  build-steps:
+    steps:
+      - checkout-all
       - restore_cache: # Download and cache dependencies
           keys:
           - v1-dependencies-{{ checksum "pyproject.toml" }}
```
rafie, line 31: v2

src/VecSim/spaces/CMakeLists.txt
```diff
@@ -1,23 +1,35 @@
...
+	set(VECSIM_ARCH "x86-64-v4")
 endif()
 message("VecSim/spaces VECSIM_ARCH: " ${VECSIM_ARCH})
 
-add_compile_options(-march=${VECSIM_ARCH})
+if (DEFINED VECSIM_ARCH)
+	add_compile_options(-march=${VECSIM_ARCH})
+endif()
+
+if(ARCH STREQUAL "x64")
```
rafie, line 13: comment for SSE/AVX/AVX512

setup.py
```diff
@@ -6,13 +6,11 @@
...
-    "win32": "Win32",
-    "win-amd64": "x64",
-    "win-arm32": "ARM",
-    "win-arm64": "ARM64",
-}
+HERE = os.path.abspath(os.path.dirname(__file__))
+ROOT = HERE
+READIES = os.path.join(ROOT, "deps/readies")
+sys.path.insert(0, READIES)
+import paella
```
rafie, line 13: comment

setup.py
```diff
@@ -26,6 +24,12 @@ def __init__(self, name, sourcedir=""):
 
 class CMakeBuild(build_ext):
     def build_extension(self, ext):
+        # plat = paella.Platform()
+        # bindir = os.path.join(os.getenv("BINDIR", f"bin/{plat.os}-{plat.arch}-release"), "pybind")
+        # paella.mkdir_p(bindir)
+        # self.build_lib = bindir
+        # self.build_temp = bindir + ".tmp"
+
```
rafie, line 32: comment

tests/flow/test_hnswlib.py
```diff
@@ -1,12 +1,10 @@
...
-import numpy as np
+from common import *
 import hnswlib
-from scipy import spatial
-from  numpy.testing import assert_allclose
 
 
 # compare results with the original version of hnswlib - do not use elements deletion.
 def test_sanity_hnswlib_index_L2():
+    BB()
```
rafie, line 7: habayta

sbin/check-format.sh
rafie: Yes. It's actually `. $READIES/shibumi/defs`
It "includes" all relevant automation code.

Makefile
```diff
@@ -8,95 +50,240 @@ include $(ROOT)/deps/readies/mk/main
...
+#----------------------------------------------------------------------------------------------
 
-benchmark:
-	./tests/benchmark.sh
+ifneq ($(wildcard /w/*),)
+SANBOX_ARGS += -v /w:/w
+endif
+
+sanbox:
+	@docker run -it -v $(PWD):/vecsim -w /vecsim --cap-add=SYS_PTRACE --security-opt seccomp=unconfined $(SANBOX_ARGS) redisfab/clang:13-x64-bullseye bash
```
DvirDukhan, line 287:
```suggestion
	@docker run -it --rm -v $(PWD):/vecsim -w /vecsim --cap-add=SYS_PTRACE --security-opt seccomp=unconfined $(SANBOX_ARGS) redisfab/clang:13-x64-bullseye bash
```

Makefile
```diff
@@ -8,95 +50,240 @@ include $(ROOT)/deps/readies/mk/main
...
 
-lint:
-	./clang-check-all.sh
+#----------------------------------------------------------------------------------------------
 
-benchmark:
-	./tests/benchmark.sh
+ifneq ($(wildcard /w/*),)
+SANBOX_ARGS += -v /w:/w
+endif
```
DvirDukhan, line 284: what is the purpose of this check?

Makefile
```diff
@@ -8,95 +50,240 @@ include $(ROOT)/deps/readies/mk/main
...
+		python3 -m RLTest \
+		--module $(BINDIR)/module_tests//memory_test.so \
+		--clear-logs \
+		$(RLTEST_ARGS)
 
 #----------------------------------------------------------------------------------------------
 
-test:
-	./tests/test.sh
+benchmark:
```
DvirDukhan, line 249: call build first, please, as we will use it during benchmark development 
also, there might be more than one benchmark, so please modify it with the ability to call multiple benchmarks

DvirDukhan (changes requested):

Makefile
```diff
@@ -1,3 +1,45 @@
...
+else ifeq ($(SAN),leak)
+else ifeq ($(SAN),thread)
+else
+$(error SAN=mem|addr|leak|thread)
+endif
+
+export SAN
+endif # SAN
+
+#----------------------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
rafie: this is the next phase - we'll have common definitions for all modules (and this will reside in `mk/readies`).
same mess exists in RediSearch - so let's take it as it is now and I'll clean it up soon for everyone.

Makefile
```diff
@@ -8,95 +50,240 @@ include $(ROOT)/deps/readies/mk/main
...
   ALL=1            # remove binary directories
 
 make all           # build all libraries and packages
 
-make test          # run tests
-
-make valgrind		   # run valgrind
-
-make benchmark	   # run micro benchmarks
+make unit_test     # run unit tests
```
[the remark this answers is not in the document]
rafie: `make valgrind` == `make VG=1 build unit_test`?
If so, I'll add it.

Makefile
```diff
@@ -8,95 +50,240 @@ include $(ROOT)/deps/readies/mk/main
...
 endef
 
 #----------------------------------------------------------------------------------------------
 
-MK_CUSTOM_CLEAN=1
 BINDIR=$(BINROOT)
+TARGET=$(BINDIR)/libVectorSimilarity.so
+SRCDIR=src
+
+MK_CUSTOM_CLEAN=1
```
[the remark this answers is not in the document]
rafie: It tells `readies/mk` that we're handling `make clean` in the scope of our Makefile.

Makefile
```diff
@@ -8,95 +50,240 @@ include $(ROOT)/deps/readies/mk/main
...
+
+ifeq ($(SLOW),1)
+MAKE_J=
+else
+MAKE_J:=-j$(shell nproc)
+endif
```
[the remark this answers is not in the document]
rafie: same as above.

Makefile
rafie: You don't want to do that because if you accidently exit the container you lose and ability to commit and run it again.

Makefile
rafie: It will mount `/w` on `/w` if such a directory exists. It allows one to mount all their repos on `/w` and get access to them.

Makefile
rafie: I usually separate build and test because when you run tests repeatedly it's annoying to see the build run in vain. The most frequent changes will be to the python files anyway.

Makefile
```diff
@@ -8,95 +50,240 @@ include $(ROOT)/deps/readies/mk/main
...
+
 ifeq ($(VERBOSE),1)
-CMAKE_FLAGS+=-DCMAKE_VERBOSE_MAKEFILE=on
+_CTEST_ARGS += -V
+endif
+
+ifeq ($(VALGRIND),1)
+_CTEST_ARGS += \
+	-T memcheck \
+	--overwrite MemoryCheckCommandOptions="--leak-check=full"
```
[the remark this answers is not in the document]
rafie: I'll check.

Makefile
```diff
@@ -8,95 +50,240 @@ include $(ROOT)/deps/readies/mk/main
...
+
+#----------------------------------------------------------------------------------------------
+
+.PHONY: __force
+
+$(BINDIR)/Makefile: __force
+	$(SHOW)cd $(BINDIR) && cmake $(CMAKE_FLAGS) $(CMAKE_DIR)
+
+$(TARGET): $(BINDIR)/Makefile
+	@echo Building $(TARGET) ...
```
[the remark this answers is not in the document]
rafie: This should also move into `readies/mk`: it's what happens when you `make build`.

.gitignore
```diff
@@ -1,3 +1,14 @@
+/tests/flow/logs/
+/tests/module/flow/logs
+
+/.tox/
+/bin/
+/build/
+/dist/
+/venv/
+/deps/readies/
+/1/
```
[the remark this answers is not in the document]
rafie: Yep. `1` is something I use like an attic (a place into which I move things before removing them completely). A sort of convention.

Makefile
```diff
@@ -8,95 +50,240 @@ include $(ROOT)/deps/readies/mk/main
...
+VECSIM_MARCH ?= skylake-avx512
+endif
+
+CMAKE_VECSIM=-DVECSIM_MARCH=$(VECSIM_MARCH)
+
+else # ARCH != x64
+
+CMAKE_VECSIM=
+
+endif # ARCH
```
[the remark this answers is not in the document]
rafie: These definitions are also used in RediSearch so I need to figure out the right way to include them into both modules. The standard way is to do it via `Makefile.defs`. I need to figure it out.

setup.py
```diff
@@ -26,6 +24,15 @@ def __init__(self, name, sourcedir=""):
 
 class CMakeBuild(build_ext):
     def build_extension(self, ext):
+        platform = paella.Platform()
```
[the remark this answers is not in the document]
rafie: Why? You need to communicate osnick/arch to cmake.
The real problem is we're doing a non-standard build here. I prefer to find a way of building via `make` and then going something like `python3 setup.py sdist ...` to take the built .so files and just make wheels out of them.

tests/unit/CMakeLists.txt
```diff
@@ -1,61 +1,35 @@
...
-	set(binroot ${root}/build)
-endif()
-message("VectorSimilarity_UnitTest: " ${binroot})
-
-if(NOT CMAKE_BUILD_TYPE)
-  set(CMAKE_BUILD_TYPE Release)
+	get_filename_component(root ${CMAKE_CURRENT_LIST_DIR}/../.. ABSOLUTE)
 endif()
+message("# VectorSimilarity/tests/unit root: " ${root})
+message("# VectorSimilarity/tests/unit binroot: " ${binroot})
```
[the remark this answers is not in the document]
rafie: This is just a sanity test, we should have defined root/binroot at a higher level. The printouts here are to verify we're not missing anything and we'll eventually remove them.

tests/flow/common.py
```diff
@@ -0,0 +1,12 @@
...
+from scipy import spatial
+from  numpy.testing import assert_allclose
+import time
+import os, sys
+
+HERE = os.path.abspath(os.path.dirname(__file__))
+ROOT = os.path.abspath(os.path.join(HERE, "../.."))
+READIES = os.path.join(ROOT, "deps/readies")
+sys.path.insert(0, READIES)
+import paella
```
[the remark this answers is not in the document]
rafie: This enables debugging with the `BB=1` idiom.
The "we don't use" approach is a little bit offensive IMO.

DvirDukhan (approved):

PR #81, opened 2021-12-08
Delete enter point bug

GuyAv46:
in the case when we delete the index entry point:
if we don't find a neighbor of the entry point to replace it,
we need to check if there is any other vector in the entry point top level,
and only then check the level below.

### DIFF
--- a/src/VecSim/algorithms/hnsw/hnswlib.h
+++ b/src/VecSim/algorithms/hnsw/hnswlib.h
@@ -806,40 +806,35 @@ bool HierarchicalNSW<dist_t>::removePoint(const labeltype label) {
     // replace the entry point with another one, if we are deleting the current entry point.
     if (element_internal_id == enterpoint_node_) {
         assert(element_top_level == maxlevel_);
-        linklistsizeint *top_level_list = get_linklist_at_level(element_internal_id, maxlevel_);
-        unsigned short list_len = getListCount(top_level_list);
-        while (list_len == 0) {
-            size_t elements = cur_element_count;
-            tableint cur_id = 0;
-            while (elements > 0) {
-                if (available_ids.find(&cur_id) == available_ids.end()) {
-                    elements--;
-                    top_level_list = get_linklist_at_level(cur_id, maxlevel_);
-                    if ((list_len = getListCount(top_level_list)))
+        // Sets the (arbitrary) new entry point.
+        while (element_internal_id == enterpoint_node_) {
+            linklistsizeint *top_level_list = get_linklist_at_level(element_internal_id, maxlevel_);
+
+            if (getListCount(top_level_list)) {
+                // Tries to set the (arbitrary) first neighbor as the entry point.
+                enterpoint_node_ = ((tableint *)(top_level_list + 1))[0];
+            } else {
+                // If there is no neighbors in the current level, check for any vector at
+                // this level to be the new entry point.
+                for (tableint cur_id = 0; cur_id <= max_id; cur_id++) {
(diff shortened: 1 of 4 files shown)


src/VecSim/algorithms/hnsw/hnswlib.h
```diff
@@ -806,27 +806,35 @@ bool HierarchicalNSW<dist_t>::removePoint(const labeltype label) {
...
-        linklistsizeint *top_level_list = get_linklist_at_level(element_internal_id, maxlevel_);
-        unsigned short list_len = getListCount(top_level_list);
-        while (list_len == 0) {
-            maxlevel_--;
-            if (maxlevel_ < 0) {
+        // Sets the (arbitrary) new entry point.
+        while (element_internal_id == enterpoint_node_) {
+            linklistsizeint *top_level_list = get_linklist_at_level(element_internal_id, maxlevel_);
+
+            if (getListCount(top_level_list)) {
```
DvirDukhan, line 813: please be verbose here
```suggestion
            if (getListCount(top_level_list) > 0) {
```

src/VecSim/algorithms/hnsw/hnswlib.h
```diff
@@ -806,27 +806,35 @@ bool HierarchicalNSW<dist_t>::removePoint(const labeltype label) {
...
+                for (tableint cur_id = 0; cur_id <= max_id; cur_id++) {
+                    if (element_levels_[cur_id] == maxlevel_ && cur_id != element_internal_id) {
+                        enterpoint_node_ = cur_id;
+                        break;
+                    }
+                }
+            }
+            // If we didn't find any vector at the top level, decrease the maxlevel_ and try again,
+            // until we find a new enter point, or the index is empty.
+            if (element_internal_id == enterpoint_node_ && --maxlevel_ < 0) {
```
DvirDukhan, line 828: make this condition clearer

tests/unit/test_hnswlib.cpp
```diff
@@ -544,3 +544,28 @@ TEST_F(HNSWLibTest, hnsw_bad_params) {
...
+    ASSERT_TRUE(index != NULL);
+
+    int64_t vec[dim];
+    for (int i = 0; i < dim; i++)
+        vec[i] = i;
+    for (size_t j = 0; j < n; j++)
+        VecSimIndex_AddVector(index, vec, j);
+
+    for (size_t j = 0; j < n; j++)
+        ASSERT_NO_THROW(VecSimIndex_DeleteVector(index, j));
```
DvirDukhan, line 570: delete the entry point

src/VecSim/vec_sim_common.h
```diff
@@ -76,6 +76,7 @@ typedef struct {
             size_t efConstruction; // EF parameter for HNSW graph accuracy/latency for indexing.
             size_t efRuntime;      // EF parameter for HNSW graph accuracy/latency for search.
             size_t levels;         // Number of graph levels.
+            size_t entry;          // Entry vector label.
```
DvirDukhan, line 79:
```suggestion
            size_t entrypoint;          // Entrypoint vector label.
```

src/VecSim/algorithms/hnsw/hnswlib.h
```diff
@@ -130,6 +130,7 @@ class HierarchicalNSW : VecsimBaseObject {
     size_t getEfConstruction() const;
     size_t getM() const;
     size_t getMaxLevel() const;
+    size_t getEnterPointLabel() const;
```
DvirDukhan, line 133:
```suggestion
    size_t getEntryPointLabel() const;
```

DvirDukhan (approved):

PR #83, opened 2021-12-12
fix memory issues

DvirDukhan:


src/VecSim/algorithms/hnsw/hnswlib.h
```diff
@@ -728,9 +727,8 @@ void HierarchicalNSW<dist_t>::resizeIndex(size_t new_max_elements) {
     if (new_max_elements < cur_element_count)
         throw std::runtime_error(
             "Cannot resize, max element is less than the current number of elements");
-    delete visited_list_pool_;
-    visited_list_pool_ =
-        new (this->allocator) VisitedListPool(1, (int)new_max_elements, this->allocator);
+    visited_list_pool_ = std::unique_ptr<VisitedListPool>(
```
alonre24, line 730: Consider using `make_unique`, I think it maybe cleaner...

src/VecSim/memory/vecsim_base.cpp
```diff
@@ -18,6 +18,20 @@ void VecsimBaseObject::operator delete[](void *p, size_t size) {
...
+    allocator->free_allocation(p);
+}
+
+void operator delete(void *p, std::shared_ptr<VecSimAllocator> allocator) {
+    allocator->free_allocation(p);
+}
+void operator delete[](void *p, std::shared_ptr<VecSimAllocator> allocator) {
+    allocator->free_allocation(p);
+}
+
```
alonre24, line 34: Consider removing the unused functions, or at least document why it is still here even though we're not using it...

alonre24 (commented):
small comments/suggestions. Also, maybe we should try find a way of disabling the sanitiser error in tests where we are allocating too much memory on purpose...

alonre24 (approved):

PR #86, opened 2021-12-16
Fix HNSW deletion tie bug

alonre24:
Deletion of a node in HNSW index is followed by a procedure of repairing edges for other nodes that were pointing to the deleted node. In this procedure, we select neighbours to the repaired node by the heuristics, and compare these to the original set of its neighbours. We assume that these 2 "nodes queues" are sorted in a unified way - by their distance to the repaired node. However, when there is tie, the order of the nodes is undefined, and may result inconsistency when going over the two queues. In some case, we miss a node that is selected to be linked to the repaired node because of that, and it leads to the disconnection of the entire graph. 

In this PR we add a seconder sorting by the id, so that the order of the nodes in the 2 queues is well defined. This also fixes an incorrect check in `knn` function of HNSW where we use `ef_construction` instead of `ef_runtime`.

### DIFF
--- a/src/VecSim/algorithms/brute_force/vector_block.h
+++ b/src/VecSim/algorithms/brute_force/vector_block.h
@@ -16,7 +16,7 @@ struct VectorBlock;
 struct CompareByFirst {
     constexpr bool operator()(std::pair<float, labelType> const &a,
                               std::pair<float, labelType> const &b) const noexcept {
-        return a.first < b.first;
+        return (a.first != b.first) ? a.first < b.first : a.second < b.second;
     }
 };
 
--- a/src/VecSim/algorithms/brute_force/vector_block.h
+++ b/src/VecSim/algorithms/brute_force/vector_block.h
@@ -4,6 +4,7 @@
 #include "VecSim/utils/vecsim_stl.h"
 
 #include "VecSim/spaces/space_interface.h"
+#include "VecSim/utils/vec_utils.h"
 
 typedef size_t labelType;
 typedef size_t idType;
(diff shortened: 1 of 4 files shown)


src/VecSim/algorithms/hnsw/hnswlib.h
```diff
@@ -31,7 +31,7 @@ template <typename dist_t>
 struct CompareByFirst {
```
GuyAv46, line 31: I noticed there is an identical function in vector_block.h. Maybe this should be implemented there as well?

GuyAv46 (commented):
Nice catch!
one question.

src/VecSim/algorithms/hnsw/hnswlib.h
alonre24: Right, I'll add this addition to `vector_block.h` as well, thanks!

src/VecSim/algorithms/brute_force/vector_block.h
```diff
@@ -16,7 +16,7 @@ struct VectorBlock;
 struct CompareByFirst {
     constexpr bool operator()(std::pair<float, labelType> const &a,
                               std::pair<float, labelType> const &b) const noexcept {
-        return a.first < b.first;
+        return (a.first != b.first) ? a.first < b.first : a.second < b.second;
```
DvirDukhan, line 19: why don't we move this comparator to utils/common helper file?
this is the same implementation here and in hnswlib?

alonre24: This isn't exactly the same implementation, since in hnswlib this comparator is generic. If and when we refactor our hnswlib implementation to use `float` as data type instead of being generic, we should definitely move this to some utils file...

DvirDukhan (approved):

PR #98, opened 2021-12-26
Hnsw batch iterator benchmark+flow test

alonre24:
Add benchmark and flow test to the new hnsw batch iterator feature. This also includes:

- Add the ability to build to project with `PROFILE=1`. This will allow us to see the stack trace whenever we run benchmarks with perf.
- Change the inner `hnsw` and `space` fields in `HNSWIndex` to shared pointers. This is needed since we use these objects in the HNSW batch iterator as well, and in case that we release the index before we release the batch iterator, we get a crash (this happens in the python bindings for example)
- Fix warning (add `override` to `indexSize()` method)


tests/flow/test_hnswlib.py
```diff
@@ -142,3 +142,90 @@ def test_recall_for_hnswlib_index_with_deletion():
...
+        dists = [(spatial.distance.euclidean(target_vector, vec), key) for key, vec in vectors]
+        dists = sorted(dists)
+        accumulated_labels = []
+        while batch_iterator.has_next():
+            iterations += 1
+            labels, distances = batch_iterator.get_next_results(batch_size, BY_SCORE)
+            accumulated_labels.extend(labels[0])
+            returned_results_num = len(accumulated_labels)
+            if returned_results_num == total_res:
+                returned_results_num = len(accumulated_labels)
```
DvirDukhan, line 210: duplicate line?

tests/flow/test_hnswlib.py
```diff
@@ -142,3 +142,90 @@ def test_recall_for_hnswlib_index_with_deletion():
...
+                keys = [key for _, key in dists[:returned_results_num]]
+                correct += len(set(accumulated_labels).intersection(set(keys)))
+                break
+        assert iterations == np.ceil(total_res/batch_size)
+    recall = float(correct) / (total_res*num_queries)
+    print(f'\nRecall for {total_res} results in index of size {num_elements} with dim={dim} is: ', recall)
+    assert recall > 0.9
```
DvirDukhan, line 217: I think we should test recall per query, not per a collection of queries

tests/flow/test_hnswlib.py
```diff
@@ -142,3 +142,90 @@ def test_recall_for_hnswlib_index_with_deletion():
...
+    iterations = 0
+    accumulated_labels = set()
+
+    while batch_iterator.has_next():
+        iterations += 1
+        labels, distances = batch_iterator.get_next_results(batch_size, BY_SCORE)
+        # Verify that we got new scores in each iteration.
+        assert len(accumulated_labels.intersection(set(labels[0]))) == 0
+        accumulated_labels = accumulated_labels.union(set(labels[0]))
+    assert len(accumulated_labels) >= 0.95*num_elements
```
DvirDukhan, line 230: this is true always, no?

Makefile
```diff
@@ -159,6 +160,10 @@ CMAKE_BUILD_TYPE=RelWithDebInfo
 endif
 CMAKE_FLAGS += -DCMAKE_BUILD_TYPE=$(CMAKE_BUILD_TYPE)
 
+ifeq ($(PROFILE),1)
+CMAKE_FLAGS += -DUSE_PROFILE=on
```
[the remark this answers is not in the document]
alonre24: PROFILE and DEBUG are independent... We usually use profiling to run the RELEASE build (faster) without loosing the stack frame

src/VecSim/algorithms/hnsw/hnswlib_c.h
```diff
@@ -22,10 +22,10 @@ class HNSWIndex : public VecSimIndex {
...
-    inline hnswlib::HierarchicalNSW<float> *getHNSWIndex() { return &hnsw; }
-    inline std::unique_ptr<SpaceInterface<float>> getSpace() { return std::move(space); }
+    inline std::shared_ptr<hnswlib::HierarchicalNSW<float>> getHNSWIndex() { return hnsw; }
+    inline std::shared_ptr<SpaceInterface<float>> getSpace() { return space; }
 
 private:
-    std::unique_ptr<SpaceInterface<float>> space;
-    hnswlib::HierarchicalNSW<float> hnsw;
+    std::shared_ptr<SpaceInterface<float>> space;
+    std::shared_ptr<hnswlib::HierarchicalNSW<float>> hnsw;
```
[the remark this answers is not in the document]
alonre24: We talked about it...

tests/flow/test_hnswlib.py
alonre24: Yes, thanks

tests/flow/test_hnswlib.py
alonre24: Not necessarily, there might be a case where the number of nodes that are reachable from the entry point in the HNSW is small (we should choose the right `M` and `ef_construction` to avoid that...)

tests/benchmark/CMakeLists.txt
```diff
@@ -7,13 +7,16 @@ project(VectorSimilarity_Benchmark)
 set(CMAKE_CXX_STANDARD 14)
 
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3 -g")
+IF (USE_PROFILE)
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-omit-frame-pointer")
+ENDIF()
```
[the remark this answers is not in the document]
alonre24: no, removed

DvirDukhan (approved):

PR #106, opened 2022-01-02
Resolve raw params

GuyAv46:
Added ResolveParams function to VecSimIndex, including the struct VecSimRawParam that holds a parameter in <string,string> key-value fashion.
Also added tests for hnsw and br indexes ResolveParams.

### DIFF
--- a/src/VecSim/algorithms/brute_force/brute_force.cpp
+++ b/src/VecSim/algorithms/brute_force/brute_force.cpp
@@ -2,7 +2,6 @@
 #include "brute_force.h"
 #include "VecSim/spaces/L2_space.h"
 #include "VecSim/spaces/IP_space.h"
-#include "VecSim/utils/arr_cpp.h"
 #include "VecSim/utils/vec_utils.h"
 #include "VecSim/query_result_struct.h"
 #include "VecSim/algorithms/brute_force/bf_batch_iterator.h"
@@ -154,9 +153,15 @@ int BruteForceIndex::deleteVector(size_t label) {
 
 size_t BruteForceIndex::indexSize() const { return this->count; }
 
-int BruteForceIndex::resolveParams(VecSimRawParam *rparams, VecSimQueryParams *qparams) {
+VecSimResolveCode BruteForceIndex::resolveParams(VecSimRawParam *rparams, int paramNum, VecSimQueryParams *qparams) {
+    if (!qparams) {
+        return VecSimErr_MissingParamStruct;
+    }
     bzero(qparams, sizeof(VecSimQueryParams));
-    return (array_len(rparams) == 0);
+    if (paramNum != 0) {
+        return VecSimErr_UnknownParam;
+    }
+    return VecSimErr_OK;
 }
 
 VecSimQueryResult_List BruteForceIndex::topKQuery(const void *queryBlob, size_t k,
--- a/src/VecSim/algorithms/brute_force/brute_force.cpp
+++ b/src/VecSim/algorithms/brute_force/brute_force.cpp
@@ -154,6 +154,11 @@ int BruteForceIndex::deleteVector(size_t label) {
 
 size_t BruteForceIndex::indexSize() const { return this->count; }
 
+int BruteForceIndex::resolveParams(VecSimRawParam *rparams, VecSimQueryParams *qparams) {
+    bzero(qparams, sizeof(VecSimQueryParams));
+    return (array_len(rparams) == 0);
+}
+
 VecSimQueryResult_List BruteForceIndex::topKQuery(const void *queryBlob, size_t k,
                                                   VecSimQueryParams *queryParams) {
 
--- a/src/VecSim/algorithms/brute_force/brute_force.cpp
+++ b/src/VecSim/algorithms/brute_force/brute_force.cpp
@@ -153,7 +153,8 @@ int BruteForceIndex::deleteVector(size_t label) {
 
 size_t BruteForceIndex::indexSize() const { return this->count; }
 
-VecSimResolveCode BruteForceIndex::resolveParams(VecSimRawParam *rparams, int paramNum, VecSimQueryParams *qparams) {
+VecSimResolveCode BruteForceIndex::resolveParams(VecSimRawParam *rparams, int paramNum,
+                                                 VecSimQueryParams *qparams) {
     if (!qparams) {
         return VecSimErr_MissingParamStruct;
     }
--- a/src/VecSim/algorithms/brute_force/brute_force.cpp
+++ b/src/VecSim/algorithms/brute_force/brute_force.cpp
@@ -155,8 +155,8 @@ size_t BruteForceIndex::indexSize() const { return this->count; }
 
 VecSimResolveCode BruteForceIndex::resolveParams(VecSimRawParam *rparams, int paramNum,
                                                  VecSimQueryParams *qparams) {
-    if (!qparams) {
-        return VecSimParamResolverErr_MissingParamStruct;
+    if (!qparams || (!rparams && (paramNum != 0))) {
+        return VecSimParamResolverErr_NullParam;
     }
     bzero(qparams, sizeof(VecSimQueryParams));
     if (paramNum != 0) {
--- a/src/VecSim/algorithms/brute_force/brute_force.cpp
+++ b/src/VecSim/algorithms/brute_force/brute_force.cpp
@@ -156,13 +156,13 @@ size_t BruteForceIndex::indexSize() const { return this->count; }
 VecSimResolveCode BruteForceIndex::resolveParams(VecSimRawParam *rparams, int paramNum,
                                                  VecSimQueryParams *qparams) {
     if (!qparams) {
-        return VecSimErr_MissingParamStruct;
+        return VecSimParamResolverErr_MissingParamStruct;
     }
     bzero(qparams, sizeof(VecSimQueryParams));
     if (paramNum != 0) {
-        return VecSimErr_UnknownParam;
+        return VecSimParamResolverErr_UnknownParam;
     }
-    return VecSimErr_OK;
+    return (VecSimResolveCode)VecSim_OK;
 }
 
 VecSimQueryResult_List BruteForceIndex::topKQuery(const void *queryBlob, size_t k,
--- a/src/VecSim/algorithms/hnsw/hnswlib_c.cpp
+++ b/src/VecSim/algorithms/hnsw/hnswlib_c.cpp
@@ -51,8 +51,8 @@ int HNSWIndex::deleteVector(size_t id) { return this->hnsw->removePoint(id); }
 
 VecSimResolveCode HNSWIndex::resolveParams(VecSimRawParam *rparams, int paramNum,
                                            VecSimQueryParams *qparams) {
-    if (!qparams) {
-        return VecSimParamResolverErr_MissingParamStruct;
+    if (!qparams || (!rparams && (paramNum != 0))) {
+        return VecSimParamResolverErr_NullParam;
     }
     bzero(qparams, sizeof(VecSimQueryParams));
     for (int i = 0; i < paramNum; i++) {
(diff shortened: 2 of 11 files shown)


src/VecSim/vec_sim_index.h
```diff
@@ -48,6 +48,15 @@ class VecSimIndex : public VecsimBaseObject {
      */
     virtual size_t indexSize() const = 0;
 
+    /**
+     * @brief Resolves VecSimRawParam array and generate VecSimQueryParams struct.
+     * @param index the index whose size is requested.
+     * @param rparams array of raw params to resolve.
```
DvirDukhan, line 54: you choose to take the assumption that you will get an instance of `arr.h` this should not be the basic assumption. In C API, you give a pointer and a length of the array

src/VecSim/algorithms/hnsw/hnswlib_c.cpp
```diff
@@ -50,6 +50,30 @@ int HNSWIndex::addVector(const void *vector_data, size_t id) {
 
 int HNSWIndex::deleteVector(size_t id) { return this->hnsw->removePoint(id); }
 
+int HNSWIndex::resolveParams(VecSimRawParam *rparams, VecSimQueryParams *qparams) {
+    bzero(qparams, sizeof(VecSimQueryParams));
```
DvirDukhan, line 54: validate null

src/VecSim/algorithms/brute_force/brute_force.cpp
```diff
@@ -154,6 +154,11 @@ int BruteForceIndex::deleteVector(size_t label) {
 
 size_t BruteForceIndex::indexSize() const { return this->count; }
 
+int BruteForceIndex::resolveParams(VecSimRawParam *rparams, VecSimQueryParams *qparams) {
+    bzero(qparams, sizeof(VecSimQueryParams));
```
DvirDukhan, line 158: validate NULL
also, what about the index type if you are setting to zeros?

tests/unit/test_bruteforce.cpp
```diff
@@ -837,3 +838,28 @@ TEST_F(BruteForceTest, brute_force_batch_iterator_corner_cases) {
...
+                                        .metric = VecSimMetric_L2,
+                                        .initialCapacity = 0,
+                                        .blockSize = 5}};
+    VecSimIndex *index = VecSimIndex_New(&params);
+
+    VecSimQueryParams qparams;
+
+    VecSimRawParam *rparams = array_new<VecSimRawParam>(1);
+
+    ASSERT_TRUE(VecSimIndex_ResolveParams(index, rparams, &qparams));
```
DvirDukhan, line 857: please validate qparams

tests/unit/test_hnswlib.cpp
```diff
@@ -913,3 +914,51 @@ TEST_F(HNSWLibTest, hnsw_batch_iterator_advanced) {
...
+                                          .M = M,
+                                          .efConstruction = ef,
+                                          .efRuntime = ef}};
+    VecSimIndex *index = VecSimIndex_New(&params);
+
+    VecSimQueryParams qparams;
+
+    VecSimRawParam *rparams = array_new<VecSimRawParam>(2);
+
+    ASSERT_TRUE(VecSimIndex_ResolveParams(index, rparams, &qparams));
```
DvirDukhan, line 937: why this is true?

tests/unit/test_hnswlib.cpp
```diff
@@ -913,3 +914,51 @@ TEST_F(HNSWLibTest, hnsw_batch_iterator_advanced) {
...
+
+    array_append(rparams, (VecSimRawParam){
+                              .name = "ef_runtime", .nameLen = 10, .value = "100", .valLen = 3});
+    ASSERT_TRUE(VecSimIndex_ResolveParams(index, rparams, &qparams));
+
+    array_append(rparams, (VecSimRawParam){
+                              .name = "ef_runtime", .nameLen = 10, .value = "100", .valLen = 3});
+    ASSERT_FALSE(VecSimIndex_ResolveParams(index, rparams, &qparams));
+
+    array_hdr(rparams)->len--;
```
DvirDukhan, line 947: array_pop? array_del_fast? don't break abstractions. If you need a function, write it

tests/unit/test_hnswlib.cpp
```diff
@@ -913,3 +914,51 @@ TEST_F(HNSWLibTest, hnsw_batch_iterator_advanced) {
...
+
+    VecSimQueryParams qparams;
+
+    VecSimRawParam *rparams = array_new<VecSimRawParam>(2);
+
+    ASSERT_TRUE(VecSimIndex_ResolveParams(index, rparams, &qparams));
+
+    array_append(rparams, (VecSimRawParam){
+                              .name = "ef_runtime", .nameLen = 10, .value = "100", .valLen = 3});
+    ASSERT_TRUE(VecSimIndex_ResolveParams(index, rparams, &qparams));
```
DvirDukhan, line 941: validate qparams?

src/VecSim/algorithms/brute_force/brute_force.cpp
GuyAv46: VecSimQueryParams doesn't hold type field. Each index is looking for its part in the union.
Right now VecSimQueryParams is just size_t for hnsw.

tests/unit/test_hnswlib.cpp
GuyAv46: I assume calling VecSimIndex_ResolveParams without params is OK, So empty array should return true.

tests/unit/test_hnswlib.cpp
GuyAv46: the value of qparams is not defined if rparams has bad params. should I zero up qparams?

src/VecSim/algorithms/hnsw/hnswlib_c.cpp
```diff
@@ -50,6 +50,30 @@ int HNSWIndex::addVector(const void *vector_data, size_t id) {
...
+    for (int i = 0; i < array_len(rparams); i++) {
+        if (!strncasecmp(rparams[i].name, VecSimCommonStrings::HNSW_EF_RUNTIME_STRING,
+                         rparams[i].nameLen)) {
+            if (qparams->hnswRuntimeParams.efRuntime != 0) { // Already been set.
+                return false;
+            } else {
+                char *ep;
```
alonre24, line 61: Consider adding documentation for what `ep` is used for

alonre24 (commented):
Looks good, added few comments. Also, consider return an error message in case that `VecSimIndex_ResolveParams` fails (@DvirDukhan ?)

src/VecSim/algorithms/hnsw/hnswlib_c.cpp
```diff
@@ -50,6 +50,30 @@ int HNSWIndex::addVector(const void *vector_data, size_t id) {
 
 int HNSWIndex::deleteVector(size_t id) { return this->hnsw->removePoint(id); }
 
+int HNSWIndex::resolveParams(VecSimRawParam *rparams, VecSimQueryParams *qparams) {
+    bzero(qparams, sizeof(VecSimQueryParams));
+    for (int i = 0; i < array_len(rparams); i++) {
+        if (!strncasecmp(rparams[i].name, VecSimCommonStrings::HNSW_EF_RUNTIME_STRING,
+                         rparams[i].nameLen)) {
+            if (qparams->hnswRuntimeParams.efRuntime != 0) { // Already been set.
```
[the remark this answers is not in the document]
GuyAv46: If someone tries to set this parameter twice.

src/VecSim/algorithms/hnsw/hnswlib_c.cpp
```diff
@@ -50,6 +50,30 @@ int HNSWIndex::addVector(const void *vector_data, size_t id) {
...
+                return false;
+            } else {
+                char *ep;
+                errno = 0;
```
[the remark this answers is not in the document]
GuyAv46: IDK. maybe there was some error earlier and the value was not changed?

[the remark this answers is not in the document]
alonre24: Then I would just check if `errno==ERANGE`. But it's not crucial...

src/VecSim/vec_sim_common.h
```diff
@@ -26,6 +26,26 @@ typedef enum { VecSimAlgo_BF, VecSimAlgo_HNSWLIB } VecSimAlgo;
...
+    size_t valLen;
+} VecSimRawParam;
+
+typedef enum {
+    VecSimErr_OK = 0,
+    VecSimErr_MissingParamStruct,
+    VecSimErr_AlreadySet,
+    VecSimErr_UnknownParam,
+    VecSimErr_BadValue
+} VecSimResolveCode;
```
DvirDukhan, line 47: This is for the specific param resolve use case, right? if so:
1. please us specific param resolve prefix like `VecSimParamResolverErr_BadValue`
2. don't use contradicting words in the same code `VecSimErr_Ok`
3. take the above error code and have it as a define, so if in the future we will need additional error code usage, we will have `VecSim_OK` shared

src/VecSim/algorithms/brute_force/brute_force.cpp
```diff
@@ -154,6 +153,18 @@ int BruteForceIndex::deleteVector(size_t label) {
 
 size_t BruteForceIndex::indexSize() const { return this->count; }
 
+VecSimResolveCode BruteForceIndex::resolveParams(VecSimRawParam *rparams, int paramNum,
+                                                 VecSimQueryParams *qparams) {
+    if (!qparams) {
```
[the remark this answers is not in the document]
GuyAv46: if rparams is null I assume paramNum is 0

src/VecSim/algorithms/hnsw/hnswlib_c.cpp
```diff
@@ -50,6 +49,34 @@ int HNSWIndex::addVector(const void *vector_data, size_t id) {
...
+        if (!strncasecmp(rparams[i].name, VecSimCommonStrings::HNSW_EF_RUNTIME_STRING,
+                         rparams[i].nameLen)) {
+            if (qparams->hnswRuntimeParams.efRuntime != 0) {
+                return VecSimErr_AlreadySet;
+            } else {
+                char *ep; // for checking strtoll used all rparams[i].valLen chars.
+                errno = 0;
+                long long val = strtoll(rparams[i].value, &ep, 0);
+                if (val <= 0 || val == LLONG_MAX || errno != 0 ||
+                    (rparams[i].value + rparams[i].valLen) != ep) {
```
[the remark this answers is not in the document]
GuyAv46: there is no str to size_t func, so Im using str to long long. this way I can also make sure we got positive number, and also that it wasnt to large number.
the last test checks that the entire rparams[i].value was used. ep will hold the char the strtoll was stoped on, and the test tests that it is the entire word. we catch here input like "3.14", "123text" (this isnt possible with RediSearch parser) and so on.

tests/unit/test_hnswlib.cpp
```diff
@@ -913,3 +914,66 @@ TEST_F(HNSWLibTest, hnsw_batch_iterator_advanced) {
...
+              VecSimErr_OK);
+    ASSERT_EQ(memcmp(&qparams, &zero, sizeof(VecSimQueryParams)), 0);
+
+    array_append(rparams, (VecSimRawParam){
+                              .name = "ef_runtime", .nameLen = 10, .value = "100", .valLen = 3});
+    ASSERT_EQ(VecSimIndex_ResolveParams(index, rparams, array_len(rparams), &qparams),
+              VecSimErr_OK);
+    ASSERT_EQ(qparams.hnswRuntimeParams.efRuntime, 100);
+    qparams.hnswRuntimeParams.efRuntime = 0;
+    ASSERT_EQ(memcmp(&qparams, &zero, sizeof(VecSimQueryParams)), 0);
```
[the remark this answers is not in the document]
GuyAv46: it checks that the only field we set is `efRuntime`

src/VecSim/algorithms/brute_force/brute_force.cpp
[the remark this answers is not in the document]
DvirDukhan: this is a wrong assumption. you can easily test it

src/VecSim/algorithms/hnsw/hnswlib_c.cpp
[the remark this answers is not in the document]
DvirDukhan: but please add it in a comment :)

DvirDukhan (approved):

PR #107, opened 2022-01-02
Compiler flags

DvirDukhan:
this PR adds compiler flag `-Wall -Werror` and the resulting fixing due to those flags


src/CMakeLists.txt
```diff
@@ -46,6 +47,8 @@ endif()
 
 add_subdirectory(VecSim/spaces)
 
+
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror -Wall")
```
rafie, line 51: Don't you want to set those in a somewhat more global location? (`build/cmake/common.cmake` or something)
Same note for spaces.

DvirDukhan: no
as this will build the google deps with those flags, and there are warnings there.

src/VecSim/vec_sim_index.h
```diff
@@ -9,7 +9,7 @@
  * @brief Abstract C++ class for vector index, delete and lookup
  *
  */
-class VecSimIndex : public VecsimBaseObject {
+struct VecSimIndex : public VecsimBaseObject {
```
alonre24, line 12: Why this change? since everything is public?

src/VecSim/algorithms/hnsw/hnswlib.h
```diff
@@ -60,14 +63,14 @@ class HierarchicalNSW : public VecsimBaseObject {
 
     // Index state
     size_t cur_element_count;
-    int max_id;
-    int maxlevel_;
+    tableint max_id;
+    size_t maxlevel_;
```
alonre24, line 67: Why `maxlevel_` is `size_t` and not `unsigned int`? It shouldn't be larger than the number of ids, and we use `unsigned int` for that...

alonre24 (commented):
LGTM

alonre24 (approved):

PR #114, opened 2022-01-10
split spaces build to different x86 architectures

DvirDukhan:


rafie (approved):

rafie:
@DvirDukhan I'd like to see there something different than CMake's `CMAKE_SYSTEM_PROCESSOR`, but currently the alternative will probably break the python build, so I'll let it be, and adjust when we handle the python build properly.

PR #115, opened 2022-01-12
Macos build

DvirDukhan:


.circleci/config.yml
```diff
@@ -106,6 +106,45 @@ jobs:
...
+       - run:
+          name: get conda
+          command: |
+            brew install wget
+            wget https://repo.anaconda.com/archive/Anaconda3-2021.11-MacOSX-x86_64.sh -O ~/anaconda.sh
+            bash ~/anaconda.sh -b -p $HOME/anaconda
+            source ~/anaconda/bin/activate
+            conda init zsh
+            source ~/.zshrc
+            conda env create -f ./conda/macos.yml
```
rafie, line 123: PLEASE PLEASE PLEASE
Don't do it like that. Let's try to avoid putting real code into config.yml.
There's `getconda` in readies now - let's make it work through there.
I would also like very much to avoid zsh in automation - but let's take it offline, I'll lay out my arguments.

.circleci/config.yml
```diff
@@ -106,6 +106,45 @@ jobs:
...
+       - run:
+          name: unit test
+          command: |
+            source ~/.zshrc
+            conda activate vecsim
+            make unit_test
+       - run:
+          name: flow test
+          command: |
+            source ~/.zshrc
```
chayim, line 139: Let's add this to a script or make target -then you can call it directly. WDYT

.circleci/config.yml
```diff
@@ -106,6 +106,45 @@ jobs:
...
 
+  build-macos:
+      macos:
+        xcode: 11.6.0
+      steps:
+       - checkout-all
+       - run:
+          name: get conda
+          command: |
+            brew install wget
```
chayim, line 117: IMHO this can be a script

.circleci/config.yml
rafie: We already have a script: https://github.com/RedisLabsModules/readies/blob/master/bin/getconda
Just need to integrate the logic there: we can have `FULL=1` or `MINI=1` selectors to get the Anaconda we require.
Note that macOS/M1 should also be in mind (not for CircleCI but we shouldn't write things that work only in CircleCI anyway).

PR #126, opened 2022-01-26
check compiler version

DvirDukhan:


DvirDukhan:
This PR builds on the basic spaces for every compiler that is not GCC11

PR #137, opened 2022-02-24
Ad-hoc bf heuristics

alonre24:
Add the `VecSimIndex_PreferAdHocSearch` API to VecSim index that returns `true` if running hybrid query in RediSearch should be faster with using "ad-hoc" BF approach, instead of using the batch iterator.
Both indexes (BF and HNSW) implements a pre-trained decision tree that determines what is the better approach per query, based on the index and the query parameters. 
The python scripts that were used to train the models are also available under the `scripts` dir, that was renamed from `notebooks`.


tests/unit/test_hnswlib.cpp
```diff
@@ -1148,3 +1148,59 @@ TEST_F(HNSWLibTest, hnsw_get_distance) {
         VecSimIndex_Free(index[i]);
     }
 }
+
+TEST_F(HNSWLibTest, preferAdHocOptimization) {
```
DvirDukhan, line 1152: why this test is different then the flat index test?

scripts/HNSW_batches_clf.py
```diff
@@ -0,0 +1,320 @@
...
+
+from sklearn.linear_model import LinearRegression
+from sklearn.tree import DecisionTreeClassifier
+from sklearn import tree
+'''------------------- batches HNSW -----------------'''
+
+
+def print_tree(clf):
+    tree.plot_tree(clf)
+    plt.savefig('batches_vs_adhoc-HNSW.pdf')
```
DvirDukhan, line 12: please add the pdf files also

DvirDukhan (approved):

PR #140, opened 2022-03-01
Fix bug in BF batch search

alonre24:
replacing indexes in scores array after heap based search wasn't taking care of the case where results are placed in the head of the array which is going to be trimmed in the next iteration.

### DIFF
--- a/src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
+++ b/src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
@@ -25,6 +25,36 @@ VecSimQueryResult *BF_BatchIterator::searchByHeuristics(size_t n_res,
     return res;
 }
 
+void BF_BatchIterator::swapScores(const unordered_map<size_t, size_t> &TopCandidatesIndices,
+                                  size_t res_num) {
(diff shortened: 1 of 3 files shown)


src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
```diff
@@ -25,6 +25,36 @@ VecSimQueryResult *BF_BatchIterator::searchByHeuristics(size_t n_res,
     return res;
 }
 
+void BF_BatchIterator::swapScores(const unordered_map<size_t, size_t> &TopCandidatesIndices,
+                                  size_t res_num) {
+    // Create a set of the indices in the scores array for every results that we return.
+    set<size_t> indices;
```
DvirDukhan, line 31: should it be sorted set? in order to enforce right order of iteration when you compare `it` to `ind`

alonre24: As far as I know `set` in c++ is the basically a sorted set. `unsorted_set` is a set which is implemented with hash table...

DvirDukhan (approved):

PR #143, opened 2022-03-09
Get distance from normalize vector

GuyAv46:
Get distance from normalize vector without normalizing it again.

### DIFF
--- a/src/VecSim/algorithms/brute_force/brute_force.cpp
+++ b/src/VecSim/algorithms/brute_force/brute_force.cpp
@@ -150,6 +150,9 @@ int BruteForceIndex::deleteVector(size_t label) {
 }
 
 double BruteForceIndex::getDistanceFrom(size_t label, const void *vector_data) {
+    return BruteForceIndex::getDistanceFrom(label, vector_data, true);
+}
+double BruteForceIndex::getDistanceFrom(size_t label, const void *vector_data, bool normalize) {
     auto optionalId = this->labelToIdLookup.find(label);
     if (optionalId == this->labelToIdLookup.end()) {
         return INVALID_SCORE;
@@ -157,7 +160,7 @@ double BruteForceIndex::getDistanceFrom(size_t label, const void *vector_data) {
     idType id = optionalId->second;
     VectorBlockMember *vector_index = this->idToVectorBlockMemberMapping[id];
     float normalized_blob[this->dim]; // This will be use only if metric == VecSimMetric_Cosine
-    if (this->metric == VecSimMetric_Cosine) {
+    if (this->metric == VecSimMetric_Cosine && normalize) {
         // TODO: need more generic
         memcpy(normalized_blob, vector_data, this->dim * sizeof(float));
         float_vector_normalize(normalized_blob, this->dim);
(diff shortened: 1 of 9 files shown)


src/VecSim/algorithms/brute_force/brute_force.cpp
```diff
@@ -150,14 +150,17 @@ int BruteForceIndex::deleteVector(size_t label) {
 }
 
 double BruteForceIndex::getDistanceFrom(size_t label, const void *vector_data) {
+    return BruteForceIndex::getDistanceFrom(label, vector_data, true);
```
alonre24, line 153: Like on the syntactic sugar ;)

src/VecSim/vec_sim.cpp
```diff
@@ -45,6 +45,13 @@ extern "C" int VecSimIndex_DeleteVector(VecSimIndex *index, size_t id) {
 extern "C" double VecSimIndex_GetDistanceFrom(VecSimIndex *index, size_t id, const void *blob) {
     return index->getDistanceFrom(id, blob);
 }
+extern "C" double VecSimIndex_GetDistanceFrom_withFlag(VecSimIndex *index, size_t id,
+                                                       const void *blob, bool normalize) {
+    return index->getDistanceFrom(id, blob, normalize);
+}
+extern "C" void VecSimIndex_PrepareVector(VecSimIndex *index, const void *source, void *dest) {
+    index->prepareVector(source, dest);
+}
```
alonre24, line 54: I don't like the name `PrepareVector`... I think that it is better to just call the child at it's name - `NormalizeVector`.
Also, I don't think that this function should be a member of `VecSimIndex`, as it is more of a "static" function. Then the signature may become something like:
 `VecSim_NormalizeVector(const void *source, void *dest, size_t dim, VecSimType type)`
@DvirDukhan WDYT?

src/VecSim/vec_sim.h
```diff
@@ -55,6 +55,30 @@ int VecSimIndex_DeleteVector(VecSimIndex *index, size_t id);
...
+ * @param id the id of the vector in the index.
+ * @param blob binary representation of the second vector. Blob size should match the index data
+ * type and dimension.
+ * @param normalize flag for skipping normalization. useful when you intend to use this function
+ * many times, and your blob is already normalized.
+ * @return The distance (according to the index's distance metric) between `blob` and the vector
+ * with id `id`.
+ */
+double VecSimIndex_GetDistanceFrom_withFlag(VecSimIndex *index, size_t id, const void *blob,
+                                            bool normalize);
```
alonre24, line 71: When `normalize=True`, this API is equivalent to `VecSimIndex_GetDistanceFrom`, right? 
So why not calling it `VecSimIndex_GetDistanceFromNormalizedVector` (which means that we assume that `blob` is normalised) and avoid the `normalize` flag?

alonre24 (commented):
It looks OK, go over my comments and suggestions and let's discuss it afterwards

src/VecSim/vec_sim.cpp
GuyAv46: the idea was to use this function before using `getDistanceFrom_withFlag` without a need to check what type of space we have.
sometimes this function will not normalize so calling it like that will be misleading.

src/VecSim/vec_sim.h
GuyAv46: this way Ill need to write the function twice or use helper func. not a big deal just another approach :). 
@DvirDukhan ?

PR #144, opened 2022-03-10
Normalize query vector in batch search cosine

alonre24:
This include copying the blob when creating batch iterator (as currently we were relying on external buffer that may change).
Unit tests for Cosine were added to both indexes.


GuyAv46 (approved):
LGTM

DvirDukhan (approved):

PR #145, opened 2022-03-10
Vecsim normalize

GuyAv46:
exported normalization function and removed normalization step from `getDistanceFrom` function

### DIFF
--- a/src/VecSim/vec_sim.cpp
+++ b/src/VecSim/vec_sim.cpp
@@ -46,6 +46,12 @@ extern "C" double VecSimIndex_GetDistanceFrom(VecSimIndex *index, size_t id, con
     return index->getDistanceFrom(id, blob);
 }
 
+extern "C" void VecSim_Normalize(void *blob, size_t dim, VecSimType type) {
+    // TODO: need more generic
+    assert(type == VecSimType_FLOAT32);
+    float_vector_normalize((float *)blob, dim);
+}
+
 extern "C" size_t VecSimIndex_IndexSize(VecSimIndex *index) { return index->indexSize(); }
 
 extern "C" VecSimResolveCode VecSimIndex_ResolveParams(VecSimIndex *index, VecSimRawParam *rparams,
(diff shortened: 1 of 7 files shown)


src/VecSim/vec_sim.cpp
```diff
@@ -46,6 +46,12 @@ extern "C" double VecSimIndex_GetDistanceFrom(VecSimIndex *index, size_t id, con
     return index->getDistanceFrom(id, blob);
 }
 
+extern "C" void VecSim_Normalize(void *blob, size_t dim, VecSimType type) {
```
DvirDukhan, line 49: do you assume the buffer is mutable?

tests/unit/test_bruteforce.cpp
```diff
@@ -971,6 +971,8 @@ TEST_F(BruteForceTest, brute_get_distance) {
     }
 
     void *query = v1;
+    void *norm = v2;
```
DvirDukhan, line 974: please state it is a unit vector. maybe change it to {1, 1,}

src/VecSim/vec_sim.cpp
GuyAv46: Yes. either send the original vector or make a copy yourself

GuyAv46: although I can move `VecSimType_sizeof` to the library and then with the same args have the copy part also

src/VecSim/vec_sim.h
```diff
@@ -49,12 +49,21 @@ int VecSimIndex_DeleteVector(VecSimIndex *index, size_t id);
  * metric.
  * @param id the id of the vector in the index.
  * @param blob binary representation of the second vector. Blob size should match the index data
- * type and dimension.
+ * type and dimension, and pre-normalized if needed.
```
alonre24, line 52: `and pre-normalized if needed.`
This should be more emphasised in my opinion, as it is not trivial...

alonre24 (dismissed):
Much better and cleaner... nice job!

alonre24 (approved):

DvirDukhan (approved):

PR #147, opened 2022-03-13
Fix bug in batch iterators for the case where batch_size=0

alonre24:
In brute force - instead of asserting that we hadn't computed the scores when the number of returned results by the batch iterator is 0, we should assert that the number of returned results by the iterator is 0 if the scores hadn't been computed yet (otherwise, we crash after asking for 0 results).
In HNSW - we should insert the entry point node to the candidates heap only when both the returned results count is 0 and the `top_candidates_extra` queue is empty. Otherwise, we might get that the entry point will be returned twice (if we ask for 0 results, and moved the entry point to the `top_candidates_extra`, and then reinsert the entry point to the candidates).


GuyAv46 (approved):
THANK YOU

DvirDukhan (approved):

PR #150, opened 2022-03-22
Data block refactor

GuyAv46:


src/VecSim/utils/data_block.h
```diff
@@ -0,0 +1,60 @@
...
+        this->members[index] = member;
+        member->index = index;
+        member->block = this;
+    }
+
+    virtual ~DataBlock();
+
+private:
+    // Data unit size.
+    size_t unitSize;
```
DvirDukhan, line 51: maybe `valueSize` ?
also please add a comment that this is measured in bytes

DvirDukhan (approved):

PR #153, opened 2022-04-27
HNSW blocksize param

GuyAv46:
adding a new `blocksize` parameter, to set the number of new vectors we need to allocate memory for when we need to resize the index


alonre24 (approved):
LGTM

DvirDukhan (approved):

PR #154, opened 2022-04-27
Get estimated memory consumption of an element

GuyAv46:
added a function to return expected element memory consumption according to the given parameters, including metadata for it.

### DIFF
--- a/src/VecSim/algorithms/brute_force/brute_force.cpp
+++ b/src/VecSim/algorithms/brute_force/brute_force.cpp
@@ -49,6 +49,10 @@ size_t BruteForceIndex::estimateInitialSize(const BFParams *params) {
     return est;
 }
 
+size_t BruteForceIndex::estimateElementMemory(const BFParams *params) {
+    return params->dim * sizeof(float);
+}
+
 void BruteForceIndex::updateVector(idType id, const void *vector_data) {
     // Get the vector block
     VectorBlockMember *vectorBlockMember = this->idToVectorBlockMemberMapping[id];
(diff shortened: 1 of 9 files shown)


src/VecSim/vec_sim.h
```diff
@@ -27,6 +27,14 @@ VecSimIndex *VecSimIndex_New(const VecSimParams *params);
  */
 size_t VecSimIndex_EstimateInitialSize(const VecSimParams *params);
 
+/**
+ * @brief normalize the vector blob in place.
```
alonre24, line 31: Remove...

alonre24 (commented):
LGTM, one small fix

src/VecSim/algorithms/brute_force/brute_force.cpp
```diff
@@ -49,6 +49,10 @@ size_t BruteForceIndex::estimateInitialSize(const BFParams *params) {
     return est;
 }
 
+size_t BruteForceIndex::estimateElementMemory(const BFParams *params) {
+    return params->dim * sizeof(float);
```
alonre24, line 53: Actually, why not consider the meta-data as well? size of `VectorBlockMember`, and `labelToIdLookup`entry?

tests/unit/test_hnswlib.cpp
```diff
@@ -1379,10 +1379,13 @@ TEST_F(HNSWLibTest, testSizeEstimation) {
     size_t size_data_per_element = size_links_level0 + dim * sizeof(float) + sizeof(labeltype);
     test_estimation += n * size_data_per_element;
 
-    size_t estimation = VecSimIndex_EstimateInitialSize(&params);
-    ASSERT_GE(estimation, test_estimation);
+    size_t iestimation = VecSimIndex_EstimateInitialSize(&params);
+    ASSERT_GE(iestimation, test_estimation);
```
alonre24, line 1383: In second thought, maybe a better test will be comparing the estimation to the actual memory consumption as reported by the allocator (relevant for BF as well).

alonre24 (commented):
I suggest comparing the estimated memory consumption (both for the index initial memory and the element) to the actual memory report by the allocator.

DvirDukhan (approved):

DvirDukhan (approved):

PR #156, opened 2022-05-01
Optional runtime params for hybrid queries

alonre24:
Add support for 2 additional runtime params - "HYBRID_POLICY" and "BATCH_SIZE" - to allow users to override the hybrid query default configuration.

### DIFF
--- a/src/VecSim/algorithms/brute_force/brute_force.cpp
+++ b/src/VecSim/algorithms/brute_force/brute_force.cpp
@@ -179,14 +179,37 @@ size_t BruteForceIndex::indexSize() const { return this->count; }
 
 VecSimResolveCode BruteForceIndex::resolveParams(VecSimRawParam *rparams, int paramNum,
                                                  VecSimQueryParams *qparams) {
-    if (!qparams || (!rparams && (paramNum != 0))) {
-        return VecSimParamResolverErr_NullParam;
-    }
-    bzero(qparams, sizeof(VecSimQueryParams));
-    if (paramNum != 0) {
-        return VecSimParamResolverErr_UnknownParam;
-    }
-    return (VecSimResolveCode)VecSim_OK;
+	if (!qparams || (!rparams && (paramNum != 0))) {
+		return VecSimParamResolverErr_NullParam;
+	}
+	bzero(qparams, sizeof(VecSimQueryParams));
+	long long num_val;
+	for (int i = 0; i < paramNum; i++) {
+		if (!strcasecmp(rparams[i].name, VecSimCommonStrings::BATCH_SIZE_STRING)) {
+			if (qparams->batchSize != 0) {
+				return VecSimParamResolverErr_AlreadySet;
+			}
+			if (validate_numeric_param(rparams[i], &num_val) != VecSimParamResolver_OK) {
+				return VecSimParamResolverErr_BadValue;
+			}
+			qparams->batchSize = (size_t)num_val;
+		} else if (!strcasecmp(rparams[i].name, VecSimCommonStrings::HYBRID_POLICY_STRING)) {
+			if (!strcasecmp(rparams[i].value, "batches")) {
+				qparams->searchMode = HYBRID_BATCHES;
+			} else if (!strcasecmp(rparams[i].value, "adhoc_bf")) {
+				qparams->searchMode = HYBRID_ADHOC_BF;
+			} else {
+				return VecSimParamResolverErr_InvalidPolicy;
+			}
+		} else {
+			return VecSimParamResolverErr_UnknownParam;
+		}
+	}
+	// The combination of AD-HOC with batch_size is invalid, as there are no batches in this policy.
+	if (qparams->searchMode == HYBRID_ADHOC_BF && qparams->batchSize > 0) {
+		return VecSimParamResolverErr_InvalidPolicy;
+	}
+	return (VecSimResolveCode)VecSim_OK;
 }
 
 VecSimQueryResult_List BruteForceIndex::topKQuery(const void *queryBlob, size_t k,
--- a/src/VecSim/algorithms/brute_force/brute_force.cpp
+++ b/src/VecSim/algorithms/brute_force/brute_force.cpp
@@ -179,37 +179,40 @@ size_t BruteForceIndex::indexSize() const { return this->count; }
 
 VecSimResolveCode BruteForceIndex::resolveParams(VecSimRawParam *rparams, int paramNum,
                                                  VecSimQueryParams *qparams) {
-	if (!qparams || (!rparams && (paramNum != 0))) {
-		return VecSimParamResolverErr_NullParam;
-	}
-	bzero(qparams, sizeof(VecSimQueryParams));
-	long long num_val;
-	for (int i = 0; i < paramNum; i++) {
-		if (!strcasecmp(rparams[i].name, VecSimCommonStrings::BATCH_SIZE_STRING)) {
-			if (qparams->batchSize != 0) {
-				return VecSimParamResolverErr_AlreadySet;
-			}
-			if (validate_numeric_param(rparams[i], &num_val) != VecSimParamResolver_OK) {
-				return VecSimParamResolverErr_BadValue;
-			}
-			qparams->batchSize = (size_t)num_val;
-		} else if (!strcasecmp(rparams[i].name, VecSimCommonStrings::HYBRID_POLICY_STRING)) {
-			if (!strcasecmp(rparams[i].value, "batches")) {
-				qparams->searchMode = HYBRID_BATCHES;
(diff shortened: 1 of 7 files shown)


src/VecSim/utils/vec_utils.h
```diff
@@ -50,6 +52,8 @@ void sort_results_by_id(VecSimQueryResult_List results);
 
 void sort_results_by_score(VecSimQueryResult_List results);
 
+VecSimResolveCode validate_numeric_param(VecSimRawParam rawParam, long long *val);
```
DvirDukhan, line 55:
```suggestion
VecSimResolveCode validate_integer_param(VecSimRawParam rawParam, long long *val);
```

src/VecSim/algorithms/brute_force/brute_force.cpp
```diff
@@ -183,8 +183,34 @@ VecSimResolveCode BruteForceIndex::resolveParams(VecSimRawParam *rparams, int pa
         return VecSimParamResolverErr_NullParam;
     }
     bzero(qparams, sizeof(VecSimQueryParams));
-    if (paramNum != 0) {
-        return VecSimParamResolverErr_UnknownParam;
+    long long num_val;
+    for (int i = 0; i < paramNum; i++) {
+        if (!strcasecmp(rparams[i].name, VecSimCommonStrings::BATCH_SIZE_STRING)) {
```
DvirDukhan, line 188: you have shared logic here and on hnsw
why not define functions for them?

tests/unit/test_bruteforce.cpp
```diff
@@ -996,6 +996,73 @@ TEST_F(BruteForceTest, brute_force_resolve_params) {
...
+    ASSERT_EQ(qparams.searchMode, HYBRID_BATCHES);
+    ASSERT_EQ(qparams.batchSize, 10);
+
+    // Check for invalid batch sizes params.
+    rparams[0].value = "not_a_number";
+    rparams[0].valLen = strlen("not_a_number");
+    ASSERT_EQ(VecSimIndex_ResolveParams(index, rparams, array_len(rparams), &qparams),
+              VecSimParamResolverErr_BadValue);
+
+    rparams[0].value = "9223372036854775808"; // LLONG_MAX+1
```
DvirDukhan, line 1046: don't we want to limit the batch size?

tests/unit/test_hnswlib.cpp
```diff
@@ -1058,6 +1058,77 @@ TEST_F(HNSWLibTest, hnsw_resolve_params) {
...
+              VecSimParamResolverErr_BadValue);
+
+    rparams[1].value = "0";
+    rparams[1].valLen = strlen("0");
+    ASSERT_EQ(VecSimIndex_ResolveParams(index, rparams, array_len(rparams), &qparams),
+              VecSimParamResolverErr_BadValue);
+
+    rparams[1].value = "10f";
+    rparams[1].valLen = strlen("10f");
+    ASSERT_EQ(VecSimIndex_ResolveParams(index, rparams, array_len(rparams), &qparams),
```
DvirDukhan, line 1129: what about moving those functions to utils and having a single test for them instead of duplicating the logic in two files?

tests/unit/test_bruteforce.cpp
alonre24: I don't think so... The actual batch size will never exceeds the index size in practice... Of course that very large batch sizes will harm performance, but if a user asks for a very large batch for some reason, we don't we allow it?

src/VecSim/utils/vec_utils.cpp
```diff
@@ -70,6 +74,19 @@ void sort_results_by_score(VecSimQueryResult_List results) {
           (__compar_fn_t)cmpVecSimQueryResultByScore);
 }
 
+VecSimResolveCode validate_numeric_param(VecSimRawParam rawParam, long long *val) {
```
GuyAv46, line 77: This function can return only positive integers. we should address that

GuyAv46 (commented):
LGTM

GuyAv46 (dismissed):
Nice!
I would fix the coverage warning

GuyAv46 (dismissed):

GuyAv46 (dismissed):
LGTM

src/VecSim/vec_sim.cpp
```diff
@@ -75,8 +75,76 @@ extern "C" void VecSim_Normalize(void *blob, size_t dim, VecSimType type) {
...
+                                                       bool hybrid) {
+
+    if (!qparams || (!rparams && (paramNum != 0))) {
+        return VecSimParamResolverErr_NullParam;
+    }
+    VecSimAlgo index_type = index->info().algo;
+    bzero(qparams, sizeof(VecSimQueryParams));
+    long long num_val;
+    for (int i = 0; i < paramNum; i++) {
+        if (!strcasecmp(rparams[i].name, VecSimCommonStrings::HNSW_EF_RUNTIME_STRING)) {
```
DvirDukhan, line 88: readability - move every argument validation to its own function

src/VecSim/vec_sim.cpp
```diff
@@ -75,8 +75,76 @@ extern "C" void VecSim_Normalize(void *blob, size_t dim, VecSimType type) {
...
+    // Also, 'ef_runtime' is meaning less in AD-HOC policy, since it doesn't involve search in HNSW
+    // graph.
+    if (qparams->searchMode == HYBRID_ADHOC_BF && index_type == VecSimAlgo_HNSWLIB &&
+        qparams->hnswRuntimeParams.efRuntime > 0) {
+        return VecSimParamResolverErr_InvalidPolicy_AdHoc_With_EfRuntime;
+    }
+    // EF_RUNTIME cannot be lower than the batch_size.
+    if (index_type == VecSimAlgo_HNSWLIB && qparams->hnswRuntimeParams.efRuntime &&
+        qparams->batchSize && qparams->hnswRuntimeParams.efRuntime < qparams->batchSize) {
+        return VecSimParamResolverErr_InvalidPolicy_BatchSize_GT_EfRuntime;
```
DvirDukhan, line 142: hnsw calls
```c++
searchLayer(bottom_layer_ep, query_data, 0, std::max(ef_, k))
```
why should we not follow this?

src/python_bindings/bindings.cpp
```diff
@@ -93,10 +93,10 @@ class PyVecSimIndex {
 
     size_t indexSize() { return VecSimIndex_IndexSize(index); }
 
-    PyBatchIterator createBatchIterator(py::object &query_blob) {
+    PyBatchIterator createBatchIterator(py::object &query_blob, VecSimQueryParams *query_params) {
```
DvirDukhan, line 96: tests?

DvirDukhan (commented):
See comments, also:
Think about the input validation tests structure, can we unify them into a single test file?

src/python_bindings/bindings.cpp
```diff
@@ -185,7 +185,14 @@ PYBIND11_MODULE(VecSim, m) {
...
 
-    py::class_<VecSimQueryParams>(m, "VecSimQueryParams").def(py::init());
+    py::class_<VecSimQueryParams> queryParams(m, "VecSimQueryParams");
+
+    queryParams.def(py::init<>())
+        .def_readwrite("hnswRuntimeParams", &VecSimQueryParams::hnswRuntimeParams);
+
+    py::class_<HNSWRuntimeParams>(queryParams, "HNSWRuntimeParams")
+        .def(py::init<>())
+        .def_readwrite("efRuntime", &HNSWRuntimeParams::efRuntime);
```
DvirDukhan, line 195: what about other members of the query parameters?

src/VecSim/vec_sim.cpp
```diff
@@ -8,6 +8,63 @@
...
+        return VecSimParamResolverErr_AlreadySet;
+    }
+    if (validate_positive_integer_param(rparam, &num_val) != VecSimParamResolver_OK) {
+        return VecSimParamResolverErr_BadValue;
+    }
+    // Asking for EF_RUNTIME lower than k is useless for standard KNN, as HNSW sets EF_RUNTIME to be
+    // >= k
+    if (k > (size_t)num_val && !hybrid) {
+        return VecSimParamResolverErr_k_GT_EfRuntime;
+    }
```
DvirDukhan, line 29: breaking change?

alonre24: Yes... should avoid? If so, we'll loose the consistency from hybrid query (where we don't allow `ef_runtime < bacth_size`)

src/python_bindings/bindings.cpp
alonre24: They are irrelevant, as the hybrid policy and the batch size are parameters in the hybrid iterator level of`rediSearch`, and they don't mean anything in the level of the "lower" batch iterator API...

DvirDukhan (approved):

PR #161, opened 2022-05-25
Timeout Enabling - Brute Force

GuyAv46:
Adding a static (global) function to VecSimIndex class to have support for 3rd party timeout functions.

By default this function just returns 0 (not timed out ever).


GuyAv46:
for template use

src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
```diff
@@ -134,17 +157,23 @@ VecSimQueryResult_List BF_BatchIterator::getNextResults(size_t n_res,
         for (auto &block : blocks) {
             // compute the scores for the vectors in every block and extend the scores array.
             vecsim_stl::vector<std::pair<float, labelType>> block_scores =
-                block->computeBlockScores(getIndex()->distFunc(), getQueryBlob());
+                this->computeBlockScores(block);
+            if (block_scores.empty()) {
```
DvirDukhan, line 161: how about sending an `int*` as an error code to the function and checking it for `VecSim_QueryResult_TimedOut` instead of relying on an empty vector

src/VecSim/algorithms/brute_force/brute_force.cpp
```diff
@@ -1,4 +1,4 @@
-
+// #pragma GCC optimize("O0")
```
DvirDukhan, line 1: remove?

src/VecSim/algorithms/brute_force/brute_force.cpp
```diff
@@ -199,6 +202,11 @@ VecSimQueryResult_List BruteForceIndex::topKQuery(const void *queryBlob, size_t
         size_t block_size = vectorBlock->getLength();
         vecsim_stl::vector<float> scores(block_size, this->allocator);
         for (size_t i = 0; i < block_size; i++) {
+            if (__builtin_expect(VecSimIndex::timeoutCallback(timeoutCtx), 0)) {
```
DvirDukhan, line 205: why not be consistent with the batch iterator method?

src/VecSim/algorithms/brute_force/brute_force.h
```diff
@@ -12,6 +12,7 @@ class BruteForceIndex : public VecSimIndex {
     size_t dim;
     VecSimType vecType;
     VecSimMetric metric;
+    friend class BF_BatchIterator;
```
DvirDukhan, line 15: what for?

src/VecSim/query_results.h
```diff
@@ -28,10 +30,19 @@ int64_t VecSimQueryResult_GetId(const VecSimQueryResult *item);
  */
 float VecSimQueryResult_GetScore(const VecSimQueryResult *item);
 
+typedef enum {
```
DvirDukhan, line 33: don't you want to define it in `vec_sim_common`?

src/VecSim/vec_sim.cpp
```diff
@@ -1,3 +1,4 @@
+// #pragma GCC optimize("O0")
```
DvirDukhan, line 1: remove?

src/VecSim/vec_sim.cpp
```diff
@@ -8,6 +9,16 @@
 #include <cassert>
 #include "memory.h"
 
+timeoutCallbackFunction VecSimIndex::timeoutCallback = [](void *ctx) { return 0; };
+
+void VecSimIndex::setTimeoutCallbackFunction(timeoutCallbackFunction callback) {
```
DvirDukhan, line 14: not a part of vecsimIndex?

tests/unit/test_bruteforce.cpp
```diff
@@ -1293,3 +1293,80 @@ TEST_F(BruteForceTest, testSizeEstimation) {
...
+
+    VecSim_SetTimeoutCallbackFunction([](void *ctx) { return 1; }); // Always times out
+    rl = VecSimBatchIterator_Next(batchIterator, 1, BY_ID);
+    ASSERT_EQ(rl.code, VecSim_QueryResult_TimedOut);
+    VecSimQueryResult_Free(rl);
+
+    VecSimBatchIterator_Free(batchIterator);
+
+    // Fail on first batch (while calculating)
+    // Timeout callback function already always times out
```
DvirDukhan, line 1362:
```suggestion
    // Timeout callback function always times out
```

tests/unit/test_common.cpp
```diff
@@ -132,3 +132,31 @@ TEST_F(CommonTest, ResolveQueryRuntimeParams) {
...
+
+    // Actual test: sets the vecsimindex timeout callback to always return timeout
+    VecSim_SetTimeoutCallbackFunction([](void *ctx) { return 1; });
+
+    rl = VecSimIndex_TopKQuery(index, vec, 1, NULL, BY_ID);
+    ASSERT_EQ(rl.code, VecSim_QueryResult_TimedOut);
+    VecSimQueryResult_Free(rl);
+
+    VecSimIndex_Free(index);
+}
```
DvirDukhan, line 162: new line

src/VecSim/algorithms/brute_force/brute_force.cpp
GuyAv46: Not consistent how? in both cases I check the timeout before every distance calculation

src/VecSim/algorithms/brute_force/brute_force.h
GuyAv46: For using the protected member `dim`. I thought that it is reasonable that the batch iterator class of the BF algorithm will be friend of the algorithm. but I can add a `getDim()` function instead

src/VecSim/vec_sim.cpp
GuyAv46: It is, but there is no `vec_sim_index.cpp` file, and if I try to define it in `vec_sim_index.h` I get an error because it is abstract and the function is defined twice (in bf and in hnsw)

tests/unit/test_bruteforce.cpp
GuyAv46: `// Timeout callback function already set to always times out`

src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
```diff
@@ -55,11 +55,12 @@ void BF_BatchIterator::swapScores(const unordered_map<size_t, size_t> &TopCandid
...
 }
 
-VecSimQueryResult *BF_BatchIterator::heapBasedSearch(size_t n_res) {
+VecSimQueryResult_List BF_BatchIterator::heapBasedSearch(size_t n_res) {
+    VecSimQueryResult_List rl = {0};
     float upperBound = std::numeric_limits<float>::lowest();
     vecsim_stl::max_priority_queue<pair<float, labelType>> TopCandidates(this->allocator);
     // map vector's label to its index in the scores vector.
-    unordered_map<size_t, size_t> TopCandidatesIndices(n_res);
+    unordered_map<size_t, size_t> TopCandidatesIndices(n_res + 1);
```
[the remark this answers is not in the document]
GuyAv46: because if the map is full (we have `n_res` results already), we first enter another result and then pop out the worst one

src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
```diff
@@ -69,28 +70,29 @@ VecSimQueryResult *BF_BatchIterator::heapBasedSearch(size_t n_res) {
             if (this->scores[i].first >= upperBound) {
                 continue;
             } else {
-                TopCandidates.emplace(this->scores[i].first, this->scores[i].second);
-                TopCandidatesIndices[this->scores[i].second] = i;
                 // remove the furthest vector from the candidates and from the label->index mappings
                 TopCandidatesIndices.erase(TopCandidates.top().second);
                 TopCandidates.pop();
+                TopCandidatesIndices[this->scores[i].second] = i;
+                TopCandidates.emplace(this->scores[i].first, this->scores[i].second);
```
DvirDukhan, line 77: Can you explain why you changed of order?

src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
```diff
@@ -105,23 +107,45 @@ VecSimQueryResult *BF_BatchIterator::selectBasedSearch(size_t n_res) {
...
+vecsim_stl::vector<std::pair<float, labelType>>
+BF_BatchIterator::computeBlockScores(VectorBlock *block, VecSimQueryResult_Code *rc) {
+    const void *queryBlob = getQueryBlob();
+    DISTFUNC<float> DistFunc = getIndex()->distFunc();
+    size_t len = block->getLength();
+    vecsim_stl::vector<std::pair<float, labelType>> scores(len, this->allocator);
+    for (size_t i = 0; i < len; i++) {
+        if (__builtin_expect(VecSimIndex::timeoutCallback(this->getTimeoutCtx()), 0)) {
+            *rc = VecSim_QueryResult_TimedOut;
+            return vecsim_stl::vector<std::pair<float, labelType>>(this->allocator);
```
DvirDukhan, line 140: can we have a static empty list to return each time instead of creating a new one?

src/VecSim/algorithms/brute_force/brute_force.cpp
DvirDukhan: you created a new function `computeBlockScore`. why not using it here?

src/VecSim/query_results.h
```diff
@@ -31,7 +33,10 @@ float VecSimQueryResult_GetScore(const VecSimQueryResult *item);
 /**
  * @brief An opaque object from which results can be obtained via iterator.
  */
-typedef VecSimQueryResult *VecSimQueryResult_List;
+typedef struct {
+    VecSimQueryResult *results;
+    VecSimQueryResult_Code code;
+} VecSimQueryResult_List;
```
DvirDukhan, line 39: let's consider renaming it. It is not a list anymore

src/VecSim/vec_sim.cpp
DvirDukhan: so let's have `vec_sim_index.cpp`

src/VecSim/algorithms/brute_force/brute_force.cpp
alonre24, line 205: Can you explain what `__builtin_expect` does? Also, I would suggest renaming `timeoutCallback` to something like: `ReachedTimeout()`

src/VecSim/vec_sim_common.h
```diff
@@ -111,6 +111,7 @@ typedef struct {
     };
     size_t batchSize;
     VecSearchMode searchMode;
+    void *timeoutCtx;
```
alonre24, line 114: To document that this isn't passed by the user and therefore doesn't need to be resolved as the others

alonre24 (commented):
LOOKS GREAT!

src/VecSim/algorithms/brute_force/bf_batch_iterator.cpp
GuyAv46: I will return the current result instead, and the caller will decide what to do with it (aka ignore)
is this better?

src/VecSim/query_results.h
GuyAv46: A list with a header is a list in my opinion. how will you call it?

alonre24 (approved):

DvirDukhan (approved):

PR #165, opened 2022-05-31
ARM build configuration

GuyAv46:


.circleci/config.yml
```diff
@@ -115,7 +123,7 @@ commands:
           name: Flow tests
           command: |
             python3 -m pip install -r sbin/requirements.txt
-            python3 -m tox -e flowenv
+            python3 -m tox -e flowenv -r
```
DvirDukhan, line 126: Do we need this?

.circleci/config.yml
```diff
@@ -59,9 +59,17 @@ commands:
       - run:
           name: basic install
           command: |
-            apt-get update
+            apt-get update --fix-missing
```
DvirDukhan, line 62: Do we need this?

pyproject.toml
```diff
@@ -28,15 +28,15 @@ classifiers = [
 python = "^3.7"
 
 [tool.poetry.dev-dependencies]
-numpy = "1.21"
-tox = "^3.24.1"
+numpy = "^1.21"
```
DvirDukhan, line 31: If arm runs tiny build, do we need the modifications in this file and the requirements file?

sbin/system-setup.py
```diff
@@ -13,40 +13,42 @@
...
-
     def debian_compat(self):
         self.run("%s/bin/getgcc --modern" % READIES)
-        self.install("python3-dev clang-format valgrind")
+        self.install("python3-dev valgrind")
 
     def redhat_compat(self):
         self.install("redhat-lsb-core")
+        self.run("%s/bin/getepel" % READIES, sudo=True)
+        self.install("libatomic")
```
DvirDukhan, line 32: Why the additions here?

sbin/system-setup.py
```diff
@@ -13,40 +13,42 @@
...
+
         self.run("%s/bin/getgcc --modern" % READIES)
 
     def fedora(self):
+        self.install("libatomic")
```
DvirDukhan, line 37: Why the addition here?

.circleci/config.yml
GuyAv46: No.

pyproject.toml
GuyAv46: No, but why not upgrade?

sbin/system-setup.py
GuyAv46: the entire file was updated

.circleci/config.yml
chayim: Sometimes apt loses its mind during updates. An update with --fix-missing solves many of these issues, when possible It rebuilds the index

DvirDukhan (approved):

PR #167, opened 2022-06-12
Add support for range query in brute force index

alonre24:
This includes:
- unit tests, combined with other existing features
- add use cases for google benchmark
- add support to python bindings + flow test


src/VecSim/vec_sim.cpp
```diff
@@ -190,6 +190,23 @@ extern "C" VecSimQueryResult_List VecSimIndex_TopKQuery(VecSimIndex *index, cons
...
+    assert((order == BY_ID || order == BY_SCORE) &&
+           "Possible order values are only 'BY_ID' or 'BY_SCORE'");
+    assert(radius >= 0 && "radius must be non-negative");
+    VecSimQueryResult_List results = index->rangeQuery(queryBlob, radius, queryParams);
+
+    if (order == BY_SCORE) {
+        sort_results_by_score(results);
+    } else {
+        sort_results_by_id(results);
+    }
```
GuyAv46, line 206: This could wait for the HNSW PR, but I think that HNSW algorithm will return results ordered by score anyway, so maybe do the sorting in the specific algorithm function.

src/VecSim/vec_sim.h
```diff
@@ -118,6 +118,22 @@ VecSimQueryResult_List VecSimIndex_TopKQuery(VecSimIndex *index, const void *que
...
+ * @param radius the radius around the query vector to search vectors within it.
+ * @param queryParams run time params for the search, which are algorithm-specific.
+ * @param order the criterion to sort the results list by it. Options are by score, or by id.
+ * @return An opaque object the represents a list of results. User can access the id and score
+ * (which is the distance according to the index metric) of every result through
+ * VecSimQueryResult_Iterator.
+ */
+VecSimQueryResult_List VecSimIndex_RangeQuery(VecSimIndex *index, const void *queryBlob,
+                                              float radius, VecSimQueryParams *queryParams,
+                                              VecSimQueryResult_Order);
```
GuyAv46, line 136: I guess `float radius` is fine for now, but we may consider a more generic parameter

GuyAv46 (commented):
Well done!
two comments

src/VecSim/vec_sim.cpp
```diff
@@ -190,6 +190,23 @@ extern "C" VecSimQueryResult_List VecSimIndex_TopKQuery(VecSimIndex *index, cons
     return results;
 }
 
+extern "C" VecSimQueryResult_List VecSimIndex_RangeQuery(VecSimIndex *index, const void *queryBlob,
+                                                         float radius,
+                                                         VecSimQueryParams *queryParams,
+                                                         VecSimQueryResult_Order order) {
+    assert((order == BY_ID || order == BY_SCORE) &&
+           "Possible order values are only 'BY_ID' or 'BY_SCORE'");
+    assert(radius >= 0 && "radius must be non-negative");
```
DvirDukhan, line 199: throw

src/VecSim/vec_sim.cpp
alonre24: Actually I'm not sure, because in the range query implementation for HNSW, we would probably *not* use a heap sorted by the scores as the `top_candidates` set, since we will not need to remove vectors from it (if you're in the range - you're in), unlike in KNN.

DvirDukhan (approved):

PR #170, opened 2022-07-04
HNSW delete improvements

alonre24:
Introducing some performance improvements for HNSW delete method. This PR includes:
   
1. Changing the incoming unidirectional edges data structure per node in the graph from `std::set` to `std::vector`.
2. Replacing helper sets that saves node's neighbors in delete operation from `std::set` to bitmaps (`std::vector<bool>`)
3. Add encoding version v1 (and use it as default)
4. Change bindings and benchmark accordingly, upload indices serialized with v1 to s3 (instead of the existing ones).
5. Use DBPedia in google benchmark instead of random data. Also, refactoring the benchmark such that in every iteration we run with a different (random) vector - and take the average times and recalls.


src/VecSim/algorithms/hnsw/serialization.cpp
```diff
@@ -235,22 +262,30 @@ HNSWIndexSerializer::HNSWIndexSerializer(std::shared_ptr<HierarchicalNSW<float>>
     hnsw_index = std::move(hnsw_index_); // Hold a reference of the index.
 }
 
-void HNSWIndexSerializer::saveIndex(const std::string &location) {
+void HNSWIndexSerializer::saveIndex_v1(const std::string &location) {
```
GuyAv46, line 265: Do we have `saveIndex_v0` or not anymore?

src/VecSim/utils/vecsim_stl.h
```diff
@@ -8,13 +8,21 @@
...
+class vector : public VecsimBaseObject, public std::vector<T, VecsimSTLAllocator<T>> {
+public:
+    explicit vector(const std::shared_ptr<VecSimAllocator> &alloc)
+        : VecsimBaseObject(alloc), std::vector<T, VecsimSTLAllocator<T>>(alloc) {}
+    explicit vector(size_t cap, const std::shared_ptr<VecSimAllocator> &alloc)
+        : VecsimBaseObject(alloc), std::vector<T, VecsimSTLAllocator<T>>(cap, alloc) {}
+    explicit vector(size_t cap, T val, const std::shared_ptr<VecSimAllocator> &alloc)
+        : VecsimBaseObject(alloc), std::vector<T, VecsimSTLAllocator<T>>(cap, val, alloc) {}
+};
+
```
GuyAv46, line 25: why were the explicit constructors needed? I recall that we used them before

tests/flow/test_hnswlib.py
```diff
@@ -1,4 +1,8 @@
 import os
+import time
+
+import pytest
+
```
GuyAv46, line 5: needed?

GuyAv46 (commented):
Looks great!
what data is used with serialization v0 that we don't want to replace? seems like a lot of work for something only we use.

src/VecSim/algorithms/hnsw/serialization.cpp
alonre24: We don't need, since from now on we only want to save in V1 (while we do want to enable *loading* of older version...)

src/VecSim/utils/vecsim_stl.h
alonre24: We need then now, since our vector class is now a different class than `std::vector` (it inherits from it...)

tests/unit/test_allocator.cpp
```diff
@@ -331,7 +330,12 @@ TEST_F(AllocatorTest, testIncomingEdgesSet) {
             hnswIndex->getHNSWIndex()->size_links_per_element_ * vec_max_level + 1 +
             AllocatorTest::vecsimAllocationOverhead;
     }
-    expected_allocation_delta += AllocatorTest::setNodeSize;
+
+    // Expect that the first element is pushed to the incoming edges vector of element 1 in level 0.
+    // Then, we account for the capacity of the buffer that is allocated for the vector data.
+    expected_allocation_delta += hnswIndex->getHNSWIndex()->getIncomingEdgesPtr(1, 0)->capacity() *
```
DvirDukhan, line 336: Why don't you create with capacity 0?

tests/benchmark/data/hnsw_indices.txt
```diff
@@ -1,4 +1,4 @@
-https://s3.amazonaws.com/redismodules/VectorSimilarity/random-1M-100-l2.hnsw
+http://dev.cto.redis.s3.amazonaws.com/VectorSimilarity/DBpedia-n1M-cosine-d768-M64-EFC512.hnsw_v1
```
DvirDukhan, line 1: Why don't you push also the v0 file?

tests/unit/test_hnswlib.cpp
```diff
@@ -1189,7 +1188,7 @@ TEST_F(HNSWLibTest, hnsw_serialization) {
     VecSimIndex_DeleteVector(new_index, (size_t)(distrib(rng) * n));
 
     // Persist index, delete it from memory and restore.
-    serializer.saveIndex(file_name);
+    serializer.saveIndex_v1(file_name);
```
DvirDukhan, line 1191: remove the comment in line 1188, it is not valid anymore, 
is there a scenario that checks block removal + serialization?

tests/benchmark/bm_batch_iterator.cpp
```diff
@@ -16,55 +16,51 @@ class BM_BatchIterator : public benchmark::Fixture {
...
+                                                        .M = M,
+                                                        .efConstruction = ef_c}};
         hnsw_index = VecSimIndex_New(&params);
 
-        // Load pre-generated HNSW index containing the same vectors as the Flat index.
-        char *location = getcwd(NULL, 0);
-        auto file_name = std::string(location) + "/tests/benchmark/data/random-1M-100-l2.hnsw";
+        // Load pre-generated HNSW index.
+        char *location = getcwd(nullptr, 0);
+        auto file_name = std::string(location) +
```
DvirDukhan, line 42: use env ROOT?

tests/benchmark/bm_batch_iterator.cpp
```diff
@@ -109,57 +112,60 @@ BENCHMARK_DEFINE_F(BM_BatchIterator, BatchedSearch_HNSW)(benchmark::State &st) {
...
-        batchIterator = VecSimBatchIterator_New(hnsw_index, query.data(), nullptr);
-        res_num = 0;
+        st.PauseTiming();
+        // Generate random query vector before test.
+        std::uniform_real_distribution<double> distrib(-1.0, 1.0);
+        for (size_t i = 0; i < dim; ++i) {
+            query[i] = (float)distrib(rng);
+        }
+
+        st.ResumeTiming();
```
DvirDukhan, line 126: Why the addition of the query generation in the function?

tests/benchmark/bm_batch_iterator.cpp
```diff
@@ -179,32 +185,42 @@ BENCHMARK_DEFINE_F(BM_BatchIterator, TopK_BF)(benchmark::State &st) {
...
-                break;
+    size_t iter = 0;
+    for (auto _ : st) {
+        st.PauseTiming();
+        // Generate random query vector before test.
+        std::uniform_real_distribution<double> distrib(-1.0, 1.0);
+        for (size_t i = 0; i < dim; ++i) {
+            query[i] = (float)distrib(rng);
+        }
+        st.ResumeTiming();
```
DvirDukhan, line 197: Why the addition of the query generation in the function?

tests/benchmark/bm_basics.cpp
```diff
@@ -9,14 +9,19 @@
 static void GetHNSWIndex(VecSimIndex *hnsw_index) {
 
     // Load the index file, if it exists in the expected path.
-    char *location = getcwd(NULL, 0);
-    auto file_name = std::string(location) + "/tests/benchmark/data/random-1M-100-l2.hnsw";
+    char *location = getcwd(nullptr, 0);
```
DvirDukhan, line 12: use env ROOT?

tests/benchmark/bm_basics.cpp
```diff
@@ -90,81 +89,113 @@ class BM_VecSimBasics : public benchmark::Fixture {
 };
 
 BENCHMARK_DEFINE_F(BM_VecSimBasics, AddVectorHNSW)(benchmark::State &st) {
-    // Add a new single vector to the index, with the minimal unused id.
-    size_t vec_id = n_vectors;
+    // Save the first 500 vector, remove then from the index, and add a different vector in every
```
DvirDukhan, line 92:
```suggestion
    // Save the first 500 vectors, remove them from the index, and add a different vector in every
```

GuyAv46:
Can we perhaps load a serialized dataset of version 0 and then save it again as v1? Without rebuilding the entire index and end up with a single serialization version?

Makefile
```diff
@@ -236,7 +236,8 @@ _CTEST_ARGS += \
 endif
 
 unit_test:
-	$(SHOW)cd $(BINDIR)/unit_tests && GTEST_COLOR=1 ctest $(_CTEST_ARGS)
+	export ROOT
+	$(SHOW)cd $(BINDIR)/unit_tests && GTEST_COLOR=1 ROOT=$(ROOT) ctest $(_CTEST_ARGS)
```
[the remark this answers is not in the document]
alonre24: So we can access the ROOT var from the code (to locate the path of the serialized index)

.circleci/config.yml
```diff
@@ -124,6 +124,14 @@ commands:
           command: |
             python3 -m pip install -r sbin/requirements.txt
             python3 -m tox -e flowenv
+      # todo: remove before merge
```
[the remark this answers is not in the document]
alonre24: yep it is here to see that CI works after the changes in the benchmark (as it only run at nightly...)

src/VecSim/algorithms/hnsw/serialization.h
```diff
@@ -16,27 +16,31 @@ typedef struct HNSWIndexMetaData {
...
+    void saveIndex_v1(const std::string &location);
 
     // Check the validity of the reproduced index.
-    HNSWIndexMetaData checkIntegrity();
+    HNSWIndexMetaData checkIntegrity(EncodingVersion = EncodingVersion_V1);
 
     // Restore the index from the file in the specified location.
-    void loadIndex(const std::string &location, SpaceInterface<float> *s);
+    void loadIndex(const std::string &location, SpaceInterface<float> *s,
+                   EncodingVersion version = EncodingVersion_V1);
```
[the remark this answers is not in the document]
alonre24: where?

tests/unit/test_allocator.cpp
alonre24: That's just how STL works. When a new (empty) vector is created, the data buffer is not allocated at all

DvirDukhan (approved):

DvirDukhan (approved):

PR #174, opened 2022-07-24
Range query support for HNSW

alonre24:
Implement based on option B in the design doc - https://redislabs.atlassian.net/wiki/spaces/DX/pages/2644312347/VecSim+-+Range+query

This includes adding test cases to google benchmark and adding the new `epsilon` parameter to the python bindings (+ using it in flow tests). Also, `epsilon` parameter is now exposed as part of the HNSW index info.

### DIFF
--- a/src/VecSim/algorithms/hnsw/hnsw_wrapper.cpp
+++ b/src/VecSim/algorithms/hnsw/hnsw_wrapper.cpp
@@ -168,6 +168,37 @@ VecSimQueryResult_List HNSWIndex::topKQuery(const void *query_data, size_t k,
     return rl;
 }
 
+VecSimQueryResult_List HNSWIndex::rangeQuery(const void *queryBlob, float radius,
+                                                   VecSimQueryParams *queryParams) {
+	auto rl = (VecSimQueryResult_List){0};
+	void *timeoutCtx = nullptr;
+	this->last_mode = RANGE_QUERY;
+
+	float normalized_blob[this->dim]; // This will be use only if metric == VecSimMetric_Cosine
+	if (this->metric == VecSimMetric_Cosine) {
+		// TODO: need more generic when other types will be supported.
+		memcpy(normalized_blob, queryBlob, this->dim * sizeof(float));
+		float_vector_normalize(normalized_blob, this->dim);
+		queryBlob = normalized_blob;
+	}
+
+	float originalEpsilon = this->hnsw->getEpsilon();
+	if (queryParams) {
+		timeoutCtx = queryParams->timeoutCtx;
+		if (queryParams->hnswRuntimeParams.epsilon != 0) {
+			hnsw->setEpsilon(queryParams->hnswRuntimeParams.epsilon);
+		}
+	}
+	// call searchRange in hnswlib
+
+	// Restore epsilon
+	hnsw->setEpsilon(originalEpsilon);
+	assert(hnsw->getEpsilon() == originalEpsilon);
+
+	rl.code = VecSim_QueryResult_OK;
+	return rl;
+}
+
 VecSimIndexInfo HNSWIndex::info() const {
 
     VecSimIndexInfo info;
--- a/src/VecSim/algorithms/hnsw/hnsw_wrapper.cpp
+++ b/src/VecSim/algorithms/hnsw/hnsw_wrapper.cpp
@@ -28,7 +28,7 @@ HNSWIndex::HNSWIndex(const HNSWParams *params, std::shared_ptr<VecSimAllocator> 
           params->efConstruction ? params->efConstruction : HNSW_DEFAULT_EF_C)),
       last_mode(EMPTY_MODE) {
     hnsw->setEf(params->efRuntime ? params->efRuntime : HNSW_DEFAULT_EF_RT);
-    hnsw->setEpsilon((params->epsilon > 0.0f) ? params->epsilon : HNSW_DEFAULT_EPSILON);
+    hnsw->setEpsilon((params->epsilon > 0.0) ? params->epsilon : HNSW_DEFAULT_EPSILON);
 }
 
 /******************** Implementation **************/
@@ -183,10 +183,10 @@ VecSimQueryResult_List HNSWIndex::rangeQuery(const void *queryBlob, float radius
         queryBlob = normalized_blob;
     }
 
-    float originalEpsilon = this->hnsw->getEpsilon();
+    double originalEpsilon = this->hnsw->getEpsilon();
     if (queryParams) {
         timeoutCtx = queryParams->timeoutCtx;
-        if (queryParams->hnswRuntimeParams.epsilon != 0) {
+        if (queryParams->hnswRuntimeParams.epsilon != 0.0) {
             hnsw->setEpsilon(queryParams->hnswRuntimeParams.epsilon);
         }
     }
@@ -216,6 +216,7 @@ VecSimIndexInfo HNSWIndex::info() const {
     info.hnswInfo.M = this->hnsw->getM();
     info.hnswInfo.efConstruction = this->hnsw->getEfConstruction();
     info.hnswInfo.efRuntime = this->hnsw->getEf();
+    info.hnswInfo.epsilon = this->hnsw->getEpsilon();
     info.hnswInfo.indexSize = this->hnsw->getIndexSize();
     info.hnswInfo.max_level = this->hnsw->getMaxLevel();
     info.hnswInfo.entrypoint = this->hnsw->getEntryPointLabel();
@@ -286,6 +287,10 @@ VecSimInfoIterator *HNSWIndex::infoIterator() {
         VecSim_InfoField{.fieldName = VecSimCommonStrings::SEARCH_MODE_STRING,
                          .fieldType = INFOFIELD_STRING,
                          .stringValue = VecSimSearchMode_ToString(info.hnswInfo.last_mode)});
+    infoIterator->addInfoField(
+        VecSim_InfoField{.fieldName = VecSimCommonStrings::HNSW_EPSILON_STRING,
+                         .fieldType = INFOFIELD_FLOAT64,
+                         .floatingPointValue = info.hnswInfo.epsilon});
 
     return infoIterator;
 }
--- a/src/VecSim/algorithms/hnsw/hnsw_wrapper.cpp
+++ b/src/VecSim/algorithms/hnsw/hnsw_wrapper.cpp
@@ -28,6 +28,7 @@ HNSWIndex::HNSWIndex(const HNSWParams *params, std::shared_ptr<VecSimAllocator> 
           params->efConstruction ? params->efConstruction : HNSW_DEFAULT_EF_C)),
       last_mode(EMPTY_MODE) {
     hnsw->setEf(params->efRuntime ? params->efRuntime : HNSW_DEFAULT_EF_RT);
+    hnsw->setEpsilon((params->epsilon > 0.0f) ? params->epsilon : HNSW_DEFAULT_EPSILON);
 }
 
 /******************** Implementation **************/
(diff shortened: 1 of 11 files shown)


src/VecSim/algorithms/hnsw/hnswlib.h
```diff
@@ -405,6 +429,54 @@ dist_t HierarchicalNSW<dist_t>::processCandidate(tableint curNodeId, const void
...
+        if (this->visited_nodes_handler->getNodeTag(candidate_id) == visited_tag)
+            continue;
+        this->visited_nodes_handler->tagNode(candidate_id, visited_tag);
+        char *candidate_data = (getDataByInternalId(candidate_id));
+
+        dist_t candidate_dist = fstdistfunc_(query_data, candidate_data, dist_func_param_);
+        if (candidate_dist < dyn_range * (1.0f + epsilon)) {
+            candidate_set.emplace(-candidate_dist, candidate_id);
+
+            __builtin_prefetch(getDataByInternalId(candidate_set.top().second));
```
GuyAv46, line 467: We should understand why we need this. Maybe add an issue to the backlog

src/VecSim/algorithms/hnsw/hnswlib.h
```diff
@@ -1269,4 +1341,85 @@ HierarchicalNSW<dist_t>::searchKnn(const void *query_data, size_t k, void *timeo
...
+    }
+
+    candidate_set.emplace(-ep_dist, ep_id);
+    this->visited_nodes_handler->tagNode(ep_id, visited_tag);
+
+    while (!candidate_set.empty()) {
+        std::pair<dist_t, tableint> curr_el_pair = candidate_set.top();
+        // If the best candidate is outside the dynamic range in more than epsilon (relatively) - we
+        // finish the search.
+        if ((-curr_el_pair.first) > dynamic_range * (1.0f + epsilon)) {
```
GuyAv46, line 1377: This feels inefficient. Consider calculating this once when it is updated.

src/VecSim/algorithms/hnsw/hnswlib.h
```diff
@@ -1269,4 +1341,85 @@ HierarchicalNSW<dist_t>::searchKnn(const void *query_data, size_t k, void *timeo
...
+        if (__builtin_expect(VecSimIndex::timeoutCallback(timeoutCtx), 0)) {
+            *rc = VecSim_QueryResult_TimedOut;
+            return results;
+        }
+        candidate_set.pop();
+
+        // Decrease the effective range, but keep dyn_range >= radius.
+        if (-curr_el_pair.first < dynamic_range && -curr_el_pair.first > radius) {
+            dynamic_range = -curr_el_pair.first;
+        }
```
GuyAv46, line 1389: consider having `dynamic_range` and `search_range` or other name for the range and the calculated one with the epsilon.

src/VecSim/algorithms/hnsw/hnswlib.h
```diff
@@ -122,12 +123,21 @@ class HierarchicalNSW : public VecsimBaseObject {
                                    candidatesMaxHeap<dist_t> &top_candidates,
                                    candidatesMaxHeap<dist_t> &candidates_set,
                                    dist_t lowerBound) const;
+    inline void processCandidate_RangeSearch(tableint curNodeId, const void *data_point,
+                                             size_t layer, float epsilon, tag_t visited_tag,
+                                             VecSimQueryResult **top_candidates,
+                                             candidatesMaxHeap<dist_t> &candidate_set,
+                                             dist_t lowerBound, float radius) const;
```
GuyAv46, line 130: shouldn't the radius be of type `dist_t`?

src/VecSim/vec_sim_common.h
```diff
@@ -65,6 +66,7 @@ typedef struct {
     size_t M;
     size_t efConstruction;
     size_t efRuntime;
+    float epsilon;
```
GuyAv46, line 69: Consider having the epsilon as `double` for later support

tests/benchmark/bm_basics.cpp
```diff
@@ -203,8 +203,6 @@ BENCHMARK_DEFINE_F(BM_VecSimBasics, TopK_HNSW)(benchmark::State &st) {
         VecSimQueryResult_Free(bf_results);
         VecSimQueryResult_Free(hnsw_results);
         iter++;
-        if (iter == queries->size())
-            break;
```
GuyAv46, line 207: leftovers?

tests/benchmark/bm_basics.cpp
```diff
@@ -224,6 +222,38 @@ BENCHMARK_DEFINE_F(BM_VecSimBasics, Range_BF)(benchmark::State &st) {
...
+    float epsilon = (1.0f / 1000.0f) * (float)st.range(1);
+    size_t iter = 0;
+    size_t total_res = 0;
+    size_t total_res_bf = 0;
+
+    for (auto _ : st) {
+        st.PauseTiming();
+
+        auto query_params =
+            VecSimQueryParams{.hnswRuntimeParams = HNSWRuntimeParams{.epsilon = epsilon}};
```
GuyAv46, line 236: Why is this in the loop?

tests/flow/test_hnswlib.py
```diff
@@ -323,3 +323,61 @@ def test_serialization():
...
+
+    query_data = np.float32(np.random.random((1, dim)))
+
+    radius = 13.0
+
+    for epsilon_rt in [0.001, 0.01, 0.1]:
+        query_params = VecSimQueryParams()
+        if epsilon_rt != 0.01:
+            # Set epsilon for the current run, except for the default value which is 0.01.
+            query_params.hnswRuntimeParams.epsilon = epsilon_rt
```
GuyAv46, line 364: seems redundant.

GuyAv46 (commented):
Nicely done!
few notes

src/VecSim/algorithms/hnsw/hnsw_wrapper.cpp
```diff
@@ -168,6 +169,42 @@ VecSimQueryResult_List HNSWIndex::topKQuery(const void *query_data, size_t k,
     return rl;
 }
 
+VecSimQueryResult_List HNSWIndex::rangeQuery(const void *queryBlob, float radius,
+                                             VecSimQueryParams *queryParams) {
```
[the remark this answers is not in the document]
alonre24: Actually, I don't think we need it in `topKQuery` as well, as no one catches this `VecSim_QueryResult_Err`... If we get such an error - it's a bug. We should replace the exceptions in `hnswlib.h` with asserts and remove the try..catch completely in my opinion (in a different PR). @DvirDukhan WDYT?

src/VecSim/algorithms/hnsw/hnswlib.h
alonre24: I agree... perhaps I'll open an issue for Yuri

src/VecSim/algorithms/hnsw/hnsw_wrapper.cpp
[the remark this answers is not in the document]
GuyAv46: OK. If we decide to remove this we should delete `VecSim_QueryResult_Err`, as that is the only reason I added it

tests/benchmark/bm_basics.cpp
alonre24: yep

src/VecSim/algorithms/hnsw/hnswlib.h
alonre24: It is sent as `float` from the library API:
(`VecSimQueryResult_List VecSimIndex_RangeQuery(
VecSimIndex *index,
 const void *queryBlob,
float radius,
VecSimQueryParams *queryParams,
VecSimQueryResult_Order order)`

since flat supports only float currently. I guess that it we can change `radius` to be generic as part of the supporting `float64` vectors PR.

tests/benchmark/bm_basics.cpp
alonre24: Shouldn't be, thanks!

tests/flow/test_hnswlib.py
alonre24: removed

src/VecSim/algorithms/hnsw/hnswlib.h
```diff
@@ -449,6 +449,8 @@ void HierarchicalNSW<dist_t>::processCandidate_RangeSearch(tableint curNodeId,
     __builtin_prefetch(getDataByInternalId(*node_links));
     __builtin_prefetch(getDataByInternalId(*(node_links + 1)));
 
+    dist_t dynamic_range_search_boundaries = dyn_range * (1.0 + epsilon);
```
GuyAv46, line 452: Redundant. you can simply pass the calculated `dynamic_range_search_boundaries` here. you don't use the dynamic range in this function.

src/VecSim/algorithms/hnsw/hnswlib.h
```diff
@@ -1367,14 +1369,15 @@ VecSimQueryResult *HierarchicalNSW<dist_t>::searchRangeBottomLayer_WithTimeout(
...
     candidate_set.emplace(-ep_dist, ep_id);
     this->visited_nodes_handler->tagNode(ep_id, visited_tag);
 
     while (!candidate_set.empty()) {
         std::pair<dist_t, tableint> curr_el_pair = candidate_set.top();
         // If the best candidate is outside the dynamic range in more than epsilon (relatively) - we
         // finish the search.
-        if ((-curr_el_pair.first) > dynamic_range * (1.0f + epsilon)) {
+        if ((-curr_el_pair.first) > dynamic_range_search_boundaries) {
             break;
```
GuyAv46, line 1381: coverage says you don't get here. Isn't it strange that no test (even the flow) does not end here?

src/VecSim/algorithms/hnsw/hnswlib.h
GuyAv46: in this file, every parameter related to the distance value should be of type `dist_t`

src/VecSim/algorithms/hnsw/hnswlib.h
alonre24: We do not account for line of code that are covered only in flow tests (or in benchmark tests)... I believe that it is covered under flow tests (otherwise we would get perfect recall). Anyway, I adjusted the unit test to cover this case as well.

GuyAv46 (dismissed):
Well done!

tests/unit/test_hnswlib.cpp
```diff
@@ -1574,4 +1612,109 @@ TEST_F(HNSWLibTest, testTimeoutReturn_batch_iterator) {
...
+    uint expected_num_results = 11;
+    // To get 11 results in the range [pivot_id - 5, pivot_id + 5], set the radius as the L2 score
+    // in the boundaries.
+    float radius = dim * powf(expected_num_results / 2, 2);
+    runRangeQueryTest(index, query, radius, verify_res_by_score, expected_num_results, BY_SCORE);
+
+    // Rerun with a given query params. This high epsilon value will cause the range search main
+    // loop to break since we insert a candidate whose distance is within the dynamic range
+    // boundaries at the beginning of the search, but when this candidate is popped out from the
+    // queue, it's no longer within the dynamic range boundaries.
```
DvirDukhan, line 1651: Except for the coverage report, do we have other indications for this?

tests/flow/test_hnswlib.py
```diff
@@ -323,3 +323,64 @@ def test_serialization():
...
+    for epsilon_rt in [0.001, 0.01, 0.1]:
+        query_params = VecSimQueryParams()
+        query_params.hnswRuntimeParams.epsilon = epsilon_rt
+        start = time.time()
+        hnsw_labels, hnsw_distances = index.range_query(query_data, radius=radius, query_param=query_params)
+        end = time.time()
+        res_num = len(hnsw_labels[0])
+
+        dists = sorted([(key, spatial.distance.euclidean(query_data, vec)) for key, vec in vectors])
+        actual_results = [(key, dist) for key, dist in dists if dist <= radius**0.5]
```
DvirDukhan, line 370: please use `spatial.distance.sqeuclidean` https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.distance.sqeuclidean.html#scipy.spatial.distance.sqeuclidean

DvirDukhan (changes requested):

src/VecSim/algorithms/hnsw/hnswlib.h
```diff
@@ -1269,4 +1339,88 @@ HierarchicalNSW<dist_t>::searchKnn(const void *query_data, size_t k, void *timeo
...
+template <typename dist_t>
+VecSimQueryResult *HierarchicalNSW<dist_t>::searchRangeBottomLayer_WithTimeout(
+    tableint ep_id, const void *data_point, double epsilon, dist_t radius, void *timeoutCtx,
+    VecSimQueryResult_Code *rc) const {
+    auto *results = array_new<VecSimQueryResult>(10); // arbitrary initial cap.
+
+#ifdef ENABLE_PARALLELIZATION
+    this->visited_nodes_handler =
+        this->visited_nodes_handler_pool->getAvailableVisitedNodesHandler();
+#endif
```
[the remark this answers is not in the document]
GuyAv46: No. Do you want me to open a PR for the updates of the enabling?

src/VecSim/vec_sim_common.h
```diff
@@ -124,12 +127,13 @@ typedef struct {
...
-            size_t blockSize;        // Sets the amount to grow when resizing
-            size_t M;                // Number of allowed edges per node in graph.
-            size_t efConstruction;   // EF parameter for HNSW graph accuracy/latency for indexing.
-            size_t efRuntime;        // EF parameter for HNSW graph accuracy/latency for search.
-            size_t max_level;        // Number of graph levels.
+            size_t indexSize;      // Current count of vectors.
+            size_t blockSize;      // Sets the amount to grow when resizing
+            size_t M;              // Number of allowed edges per node in graph.
+            size_t efConstruction; // EF parameter for HNSW graph accuracy/latency for indexing.
+            size_t efRuntime;      // EF parameter for HNSW graph accuracy/latency for search.
```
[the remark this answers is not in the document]
alonre24: Yes I understand that. I believe that this is the right way to do it, no? Also, Since range query is planned for RediSearch 2.6 where we are adding our optional hybrid queries parameters, I guess we can also add the `EPSILON` optional parameter as well.

tests/unit/test_hnswlib.cpp
alonre24: Unfortunately I could think of a good way to do that...

src/VecSim/algorithms/hnsw/hnswlib.h
```diff
@@ -405,6 +429,52 @@ dist_t HierarchicalNSW<dist_t>::processCandidate(tableint curNodeId, const void
...
+        if (this->visited_nodes_handler->getNodeTag(candidate_id) == visited_tag)
+            continue;
+        this->visited_nodes_handler->tagNode(candidate_id, visited_tag);
+        char *candidate_data = getDataByInternalId(candidate_id);
+
+        dist_t candidate_dist = fstdistfunc_(query_data, candidate_data, dist_func_param_);
+        if (candidate_dist < dyn_range) {
+            candidate_set.emplace(-candidate_dist, candidate_id);
+
+            __builtin_prefetch(get_linklist_at_level(candidate_set.top().second, layer));
```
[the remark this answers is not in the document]
alonre24: I moved it to the end of the for loop (so we pre-fetch only once). Doesn't seem to have a major effect on the benchmark

DvirDukhan (approved):

PR #176, opened 2022-07-26
Google benchmark - refactor

alonre24:
This refactor includes the following:
1. Switch to using query vector from DBPedia test set for the benchmark test cases (instead of random vectors). We load these upon constructing the benchmark class from a file that has been uploaded to s3.
2. Use the same query vectors in every test case, but different vectors in every iteration (rather than using a new random vector in every iteration of every test). 
3. Load HNSW and FLAT indices in the constructor, rather than in the setup method. This saves significant amount of time. Note that google benchmark is creating an instance for every test case class *before* it runs the tests. To reduce the memory usage, we create the base class as a single-tone, and reuse the indices among different test cases (that are guaranteed to run serially).
4. Upgrade google-test and google-benchmark versions.

### DIFF
--- a/tests/benchmark/bm_basics.cpp
+++ b/tests/benchmark/bm_basics.cpp
@@ -139,25 +139,25 @@ BENCHMARK_DEFINE_F(BM_VecSimBasics, AddVectorHNSW)(benchmark::State &st) {
 
 BENCHMARK_DEFINE_F(BM_VecSimBasics, DeleteVectorHNSW)(benchmark::State &st) {
     // Remove a different vector in every execution.
-    float *blobs[1000];
+    std::vector<std::vector<float>> blobs;
     size_t id_to_remove = 0;
 
     for (auto _ : st) {
         st.PauseTiming();
-        blobs[id_to_remove] = new float[dim];
-        memcpy(blobs[id_to_remove],
+        auto removed_vec = std::vector<float>(dim);
+        memcpy(removed_vec.data(),
                reinterpret_cast<HNSWIndex *>(hnsw_index)
                    ->getHNSWIndex()
                    ->getDataByInternalId(id_to_remove),
                dim * sizeof(float));
+        blobs.push_back(removed_vec);
         st.ResumeTiming();
         VecSimIndex_DeleteVector(hnsw_index, id_to_remove++);
     }
 
     // Restore index state.
-    for (size_t i = 0; i < id_to_remove; i++) {
-        VecSimIndex_AddVector(hnsw_index, blobs[i], i);
-        delete[] blobs[i];
+    for (size_t i = 0; i < blobs.size(); i++) {
+        VecSimIndex_AddVector(hnsw_index, blobs[i].data(), i);
     }
 }
 
@@ -166,6 +166,8 @@ BENCHMARK_DEFINE_F(BM_VecSimBasics, TopK_BF)(benchmark::State &st) {
     size_t iter = 0;
     for (auto _ : st) {
         VecSimIndex_TopKQuery(bf_index, (*queries)[iter++].data(), k, nullptr, BY_SCORE);
+        if (iter == queries->size())
+            break;
     }
 }
 
@@ -203,6 +205,8 @@ BENCHMARK_DEFINE_F(BM_VecSimBasics, TopK_HNSW)(benchmark::State &st) {
         VecSimQueryResult_Free(bf_results);
         VecSimQueryResult_Free(hnsw_results);
         iter++;
+        if (iter == queries->size())
+            break;
         st.ResumeTiming();
     }
     st.counters["Recall"] = (float)correct / (k * iter);
@@ -217,6 +221,8 @@ BENCHMARK_DEFINE_F(BM_VecSimBasics, Range_BF)(benchmark::State &st) {
         auto res =
             VecSimIndex_RangeQuery(bf_index, (*queries)[iter++].data(), radius, nullptr, BY_ID);
         total_res += VecSimQueryResult_Len(res);
+        if (iter == queries->size())
+            break;
     }
     st.counters["Avg. results number"] = (double)total_res / iter;
 }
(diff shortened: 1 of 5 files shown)


tests/benchmark/bm_basics.cpp
```diff
@@ -27,122 +27,163 @@ static void GetHNSWIndex(VecSimIndex *hnsw_index) {
...
+        auto query_params =
+            VecSimQueryParams{.hnswRuntimeParams = HNSWRuntimeParams{.efRuntime = ef}};
+        auto hnsw_results =
+            VecSimIndex_TopKQuery(hnsw_index, (*queries)[iter].data(), k, &query_params, BY_SCORE);
         st.PauseTiming();
 
         // Measure recall:
-        auto bf_results = VecSimIndex_TopKQuery(bf_index, query.data(), k, nullptr, BY_SCORE);
+        auto bf_results =
+            VecSimIndex_TopKQuery(bf_index, (*queries)[iter].data(), k, nullptr, BY_SCORE);
```
GuyAv46, line 186: same goes here regarding generality, maybe add a module for the iteration to not overshoot the queries array

tests/benchmark/bm_basics.cpp
```diff
@@ -171,29 +212,21 @@ BENCHMARK_DEFINE_F(BM_VecSimBasics, Range_BF)(benchmark::State &st) {
...
-        // Generate random query vector before test.
-        std::uniform_real_distribution<double> distrib(-1.0, 1.0);
-        for (size_t i = 0; i < dim; ++i) {
-            query[i] = (float)distrib(rng);
-        }
-        st.ResumeTiming();
-        auto res = VecSimIndex_RangeQuery(bf_index, query.data(), radius, nullptr, BY_ID);
-        st.PauseTiming();
+        auto res =
+            VecSimIndex_RangeQuery(bf_index, (*queries)[iter++].data(), radius, nullptr, BY_ID);
```
GuyAv46, line 218: same goes here regarding generality, maybe add a module for the iteration to not overshoot the queries array

tests/benchmark/bm_basics.cpp
```diff
@@ -207,9 +240,17 @@ BENCHMARK_REGISTER_F(BM_VecSimBasics, TopK_BF)
...
+    ->Args({10, 10})
+    ->Iterations(100)
+    ->Args({200, 10})
+    ->Iterations(100)
+    ->Args({100, 100})
+    ->Iterations(100)
+    ->Args({200, 100})
+    ->Iterations(100)
+    ->Args({500, 500})
+    ->Iterations(100)
```
GuyAv46, line 253: no need to specify the iteration for every args. it will take the last `Iterations` value for all args

tests/benchmark/bm_batch_iterator.cpp
```diff
@@ -8,83 +8,98 @@
...
-        st.PauseTiming();
-        // Generate random query vector before test.
-        std::uniform_real_distribution<> distrib;
-        for (size_t i = 0; i < dim; ++i) {
-            query[i] = (float)distrib(rng);
-        }
-        st.ResumeTiming();
         VecSimBatchIterator *batchIterator =
-            VecSimBatchIterator_New(bf_index, query.data(), nullptr);
+            VecSimBatchIterator_New(bf_index, (*queries)[iter++].data(), nullptr);
```
GuyAv46, line 102: same goes here regarding generality, maybe add a module for the iteration to not overshoot the queries array.
this comment is regarding all queries array access in this file.

GuyAv46 (commented):
Looks awesome! great improvement.
Can you tell how quick the benchmarks are now?

tests/benchmark/bm_basics.cpp
```diff
@@ -27,122 +27,163 @@ static void GetHNSWIndex(VecSimIndex *hnsw_index) {
...
     for (auto _ : st) {
-        st.PauseTiming();
-        // Generate random query vector before test.
-        std::uniform_real_distribution<double> distrib(-1.0, 1.0);
-        for (size_t i = 0; i < dim; ++i) {
-            query[i] = (float)distrib(rng);
-        }
-        st.ResumeTiming();
-        VecSimIndex_TopKQuery(bf_index, query.data(), k, nullptr, BY_SCORE);
+        VecSimIndex_TopKQuery(bf_index, (*queries)[iter++].data(), k, nullptr, BY_SCORE);
```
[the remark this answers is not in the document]
alonre24: Not risky, we usually have only 3 iteration... but will add safety check

tests/benchmark/bm_basics.cpp
alonre24: thanks!

GuyAv46 (dismissed):

GuyAv46 (approved):

tests/benchmark/bm_basics.cpp
```diff
@@ -171,29 +217,22 @@ BENCHMARK_DEFINE_F(BM_VecSimBasics, Range_BF)(benchmark::State &st) {
...
     st.counters["Avg. results number"] = (double)total_res / iter;
 }
 
 // Register the function as a benchmark
 BENCHMARK_REGISTER_F(BM_VecSimBasics, Range_BF)
     // The actual radius will the given arg divided by 100, since arg must me and integer.
-    ->Arg(92)
-    ->Arg(95)
-    ->Arg(100)
+    ->Arg(20)
```
DvirDukhan, line 233: typo above `since arg must me and integer.` => `since arg must be an integer.`

tests/benchmark/bm_basics.cpp
```diff
@@ -27,122 +27,166 @@ static void GetHNSWIndex(VecSimIndex *hnsw_index) {
...
+
 class BM_VecSimBasics : public benchmark::Fixture {
 protected:
-    std::mt19937 rng;
     VecSimIndex *bf_index;
     VecSimIndex *hnsw_index;
     size_t dim;
     size_t n_vectors;
-    std::vector<float> query;
+    std::vector<std::vector<float>> *queries;
```
DvirDukhan, line 56: Must be a pointer?

tests/benchmark/bm_batch_iterator.cpp
```diff
@@ -8,83 +8,99 @@
 
 class BM_BatchIterator : public benchmark::Fixture {
 protected:
-    std::mt19937 rng;
     VecSimIndex *bf_index;
     VecSimIndex *hnsw_index;
     size_t dim;
     size_t n_vectors;
-    std::vector<float> query;
+    std::vector<std::vector<float>> *queries;
```
DvirDukhan, line 15: Must be a pointer?

tests/benchmark/bm_basics.cpp
alonre24: To avoid copying it when we create several instances

DvirDukhan (approved):

GuyAv46 (approved):

PR #181, opened 2022-07-31
Improving pre-fetch logic - HNSW

alonre24:
Remove redundant unclear pre-fetching, and update pre-fetching of top candidate in the loop that scans node's neighbors. Also, add a pre-fetch call in batch iterator that we didn't had so far. 
In google benchmarks, we cannot see a significant difference, except for in batch iterator.
Meanwhile, an issue was opened in HNSWLIB repo, to have better understanding of the pre-fetch logic - https://github.com/nmslib/hnswlib/issues/397


src/VecSim/algorithms/hnsw/hnsw_batch_iterator.cpp
```diff
@@ -126,6 +124,7 @@ candidatesMaxHeap HNSW_BatchIterator::scanGraph(candidatesMinHeap &candidates,
             float candidate_dist = dist_func(this->getQueryBlob(), (const void *)candidate_data,
                                              this->space->get_data_dim());
             candidates.emplace(candidate_dist, candidate_id);
+            __builtin_prefetch(hnsw_index->get_linklist_at_level(candidates.top().second, 0));
```
GuyAv46, line 127: Here we prefetch each time we update the candidates set. Consider benchmarking an alternative where we prefetch once before calling `processCandidates`

src/VecSim/algorithms/hnsw/hnswlib.h
```diff
@@ -110,8 +110,8 @@ class HierarchicalNSW : public VecsimBaseObject {
     size_t getRandomLevel(double reverse_size);
     vecsim_stl::vector<tableint> *getIncomingEdgesPtr(tableint internal_id, size_t level) const;
     void setIncomingEdgesPtr(tableint internal_id, size_t level, void *set_ptr);
-    linklistsizeint *get_linklist0(tableint internal_id) const;
-    linklistsizeint *get_linklist(tableint internal_id, size_t level) const;
+    inline linklistsizeint *get_linklist0(tableint internal_id) const;
+    inline linklistsizeint *get_linklist(tableint internal_id, size_t level) const;
```
GuyAv46, line 114: there are more functions to inline in the class. consider doing them all or doing them all in another PR. There is an issue already regarding this.

src/VecSim/algorithms/hnsw/hnsw_batch_iterator.cpp
GuyAv46: also, if you only prefetch layer 0, you can call directly to `get_linklist0`

alonre24: In `addPpoint` we might prefetch other layers as well...

DvirDukhan (approved):

PR #190, opened 2022-08-03
added sanity test for updating vector in HNSW algo

GuyAv46:
As a follow up for @meiravgri PR that found a bug in brute force algorithm (#178), I add a test for checking that HNSW move blobs correctly on updates

### DIFF
--- a/tests/unit/test_hnswlib.cpp
+++ b/tests/unit/test_hnswlib.cpp
@@ -78,7 +78,7 @@ TEST_F(HNSWLibTest, hnswlib_blob_sanity_test) {
     ASSERT_HNSW_BLOB_EQ(1, b);
     ASSERT_EQ(GET_LABEL(1), 46);
 
-    // After inserting c with label 1, we first delete id 1 from the index.
+    // After inserting c with label 46, we first delete id 1 from the index.
     // we expect id 0 to not change
     VecSimIndex_AddVector(index, (const void *)c, 46);
     ASSERT_EQ(VecSimIndex_IndexSize(index), 2);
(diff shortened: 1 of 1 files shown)


tests/unit/test_hnswlib.cpp
```diff
@@ -39,6 +39,49 @@ TEST_F(HNSWLibTest, hnswlib_vector_add_test) {
...
+    ASSERT_EQ(VecSimIndex_IndexSize(index), 1);
+    v = reinterpret_cast<HNSWIndex *>(index)->getHNSWIndex()->getDataByInternalId(0);
+    ASSERT_FALSE(memcmp(v, a, sizeof(a)));
+
+    VecSimIndex_AddVector(index, (const void *)b, 1);
+    ASSERT_EQ(VecSimIndex_IndexSize(index), 2);
+    v = reinterpret_cast<HNSWIndex *>(index)->getHNSWIndex()->getDataByInternalId(1);
+    ASSERT_FALSE(memcmp(v, b, sizeof(b)));
+
+    // After inserting c with label 0, we first delete id 0 and move the last id (1) to be 0.
```
meiravgri, line 73: consider checking also updating the last vector

meiravgri:
cool :)

tests/unit/test_hnswlib.cpp
```diff
@@ -39,6 +39,65 @@ TEST_F(HNSWLibTest, hnswlib_vector_add_test) {
...
     VecSimIndex_Free(index);
 }
+
+TEST_F(HNSWLibTest, hnswlib_blob_sanity_test) {
+    size_t dim = 4;
+    size_t bs = 1;
+#define ASSERT_HNSW_BLOB_EQ(id, blob)                                                              \
+    do {                                                                                           \
+        void *v = reinterpret_cast<HNSWIndex *>(index)->getHNSWIndex()->getDataByInternalId(id);   \
+        ASSERT_FALSE(memcmp(v, blob, sizeof(blob)));                                               \
```
DvirDukhan, line 49: Is sizeof(blob) return the actual length or the size of the pointer?

GuyAv46: Size of the entire blob in bytes

DvirDukhan (approved):

PR #202, opened 2022-08-14
Fp64_add_datasize_to_vectorBlock

meiravgri:
**brute force vector block:**
- Changed brute force vector block to hold vector bytes count instead of vector dimension
- Changed the type of data pointer to char*


GuyAv46 (commented):
Looks good. My only comment is that now this data structure is not bounded to vectors specifically, so consider changing its name to DataBlock.

GuyAv46 (dismissed):

.gitignore
```diff
@@ -59,6 +59,3 @@ __pycache__/
 
 .vscode
 **/.ipynb_checkpoints
-
-# spelling artifact, locally
-wordlist.dic
```
DvirDukhan, line 64: Why the deletion?

DvirDukhan (dismissed):

DvirDukhan (approved):

PR #203, opened 2022-08-14
Adding spellcheck to CI, and fixing docs along the way

chayim:

A clear and concise description of what the PR is solving.

1. #...
2. MOD...

1. ...
2. ...

**Mark if applicable**


.github/workflows/spellcheck.yml
```diff
@@ -0,0 +1,16 @@
+name: Spellcheck
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
```
DvirDukhan, line 6: No need for version branches?

.gitignore
```diff
@@ -59,3 +59,5 @@ __pycache__/
 
 .vscode
 **/.ipynb_checkpoints
+
+wordlist.dic
```
DvirDukhan, line 63: is this an artifact?

DvirDukhan (commented):
thanks @chayim 
1. can you please update PR description
2. should we add "make spellcheck" ?

.github/workflows/spellcheck.yml
chayim: Added branches.

I don't think we should have a make target - CI will tell us, fairly quickly. If we see it's necessary, let's do it then, but IMHO we shouldn't.

DvirDukhan (approved):

PR #208, opened 2022-08-17
Changed distance function last argument type

meiravgri:
In the current implemntation, distance functions last argument was of type const void *, which was converted to size_t in the function body
Since it is only used to pass the vector dimensions, we changed the type to size_t.

This change saves casting + derefrencing the dim parameter each time we call a function to calculate distance between two vectors.


src/VecSim/spaces/spaces.h
```diff
@@ -8,7 +8,7 @@ static const Arch_Optimization arch_opt = getArchitectureOptimization();
 
 // TODO change last arg to size_t
```
GuyAv46, line 9: Done :)

src/VecSim/spaces/IP/IP_AVX.cpp
```diff
@@ -1,16 +1,14 @@
 
+#include <cstdlib>
+
 #include "IP_AVX.h"
 #include "VecSim/spaces/space_includes.h"
 #include "VecSim/spaces/IP/IP.h"
```
GuyAv46, line 6: Are all of these needed?

GuyAv46 (approved):
Looking good!
Consider going over some of the includes and reorganize them.
also please mark this PR for benchmark and see if there is an effect on the distance functions benchmarks

src/VecSim/spaces/IP/IP_AVX.cpp
meiravgri: yes (except stdlib)

DvirDukhan (approved):

PR #225, opened 2022-08-29
update outbound DS

meiravgri:

**query results** struct now holds doubles.
It means that in case the distance function return type is float, each score of the top candidates will be casted to double.

**Serializer class** is templated according to distType as it holds an hnswlib<DistType> instance

radius in **VecSimIndex_RangeQuery** is now double.

**python bindings**
added vecType and metric to PyHNSWLibIndex class 
used to deduce which templetaed serializer to call

1. VecSimQueryResult struct holds float, which are anyway casted in RediSearch
2. hard coded float in hnswlib type data member of serilizer
3. range query radius was double

1. VecSimQueryResult struct
2. HNSWIndexSerializer

**Mark if applicable**

**VecSimQueryResult_GetScore**  returns double instead of float. However, when called in RediSearch the return value is stored as double anyway.
**VecSimIndex_RangeQuery** receives radius as double instead of float, but no calls from RediSearch to this function yet.


tests/unit/test_bruteforce.cpp
```diff
@@ -1751,15 +1751,15 @@ TEST_F(BruteForceTest, rangeQueryCosine) {
...
-        float first_coordinate = float(n - index) / n;
+        double first_coordinate = double(n - index) / n;
         // By cosine definition: 1 - ((A \dot B) / (norm(A)*norm(B))), where A is the query vector
         // and B is the current result vector.
-        float expected_score =
+        double expected_score =
             1.0f -
-            ((first_coordinate + (float)dim - 1.0f) /
-             (sqrtf((float)dim) * sqrtf((float)(dim - 1) + first_coordinate * first_coordinate)));
+            ((first_coordinate + (double)dim - 1.0f) /
```
alonre24, line 1761:
```suggestion
            ((first_coordinate + (double)dim - 1.0) /
```

also in other places, remove the `f` (which stands for float) where it isn't necessary..

alonre24 (commented):
All good, few small comments

src/VecSim/algorithms/hnsw/serialization.h
```diff
@@ -17,9 +22,10 @@ typedef struct HNSWIndexMetaData {
 
 typedef enum EncodingVersion { EncodingVersion_V1 = 1 } EncodingVersion;
 
+template <typename DistType>
```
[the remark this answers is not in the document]
meiravgri: no,  wrapper needs both types, but batch iterator holds hnswlib
if and when we unify wrapper and lib, then batch iterartor will have both types

src/python_bindings/bindings.cpp
```diff
@@ -125,15 +123,30 @@ class PyHNSWLibIndex : public PyVecSimIndex {
...
+                auto serializer = hnswlib::HNSWIndexSerializer<float>(
+                    reinterpret_cast<HNSWIndex *>(index)->getHNSWIndex());
+                serializer.loadIndex(location);
+            }
+        }
     }
+
+private:
+    VecSimType vecType;
+    VecSimMetric metric;
```
alonre24, line 149: I don't think that these fields are necessary... you can use `VecSimIndex_Info(this->index)` instead to get the index type and create the appropriate serializer

tests/unit/test_bruteforce.cpp
```diff
@@ -1751,26 +1751,26 @@
...
     // to get the radius.
     float edge_first_coordinate = (float)(n - expected_num_results + 1) / n;
-    float radius =
-        1.0f - ((edge_first_coordinate + (float)dim - 1.0f) /
-                (sqrtf((float)dim) *
-                 sqrtf((float)(dim - 1) + edge_first_coordinate * edge_first_coordinate)));
+    double radius =
+        1.0 - ((edge_first_coordinate + (double)dim - 1.0) /
+               (sqrt((double)dim) *
+                sqrt((double)(dim - 1) + edge_first_coordinate * edge_first_coordinate)));
```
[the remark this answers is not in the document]
alonre24: I don't understand this... do you? @meiravgri

src/python_bindings/bindings.cpp
```diff
@@ -125,15 +123,30 @@ class PyHNSWLibIndex : public PyVecSimIndex {
         hnsw->setEf(ef);
     }
     void saveIndex(const std::string &location) {
-        auto serializer = hnswlib::HNSWIndexSerializer<float>(
-            reinterpret_cast<HNSWIndex *>(index)->getHNSWIndex());
-        serializer.saveIndex(location);
+        if (vecType == VecSimType_FLOAT32) {
```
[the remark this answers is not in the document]
meiravgri: these are not validations, when we have double ill add "else if type == double...."

src/python_bindings/bindings.cpp
meiravgri: isn't it a bit redundant to call this function, and retrieve type and metric from the vecsiminfo struct, at the beginning of every function in this class (except the first)

tests/unit/test_bruteforce.cpp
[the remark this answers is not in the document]
meiravgri: i fixed this
if you look in the source code you'll see that I forgot to change edge_first_coordinate type to double (one line above)

src/python_bindings/bindings.cpp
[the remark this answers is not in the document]
alonre24: I meant the metric validations

src/python_bindings/bindings.cpp
alonre24: As I see it, the `vecType` field is already stored in `this->index`, so storing it again in another field of the class is a bit redundant. Though it's not critical...

meiravgri: K

src/python_bindings/bindings.cpp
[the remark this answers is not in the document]
meiravgri: both type and metric determine the type of the distance function, I'm not sure which is more important
but you are right that as long as these are the only metrics, checking the type is enough

alonre24 (approved):

src/VecSim/algorithms/brute_force/brute_force.h
```diff
@@ -33,7 +33,7 @@ class BruteForceIndex : public VecSimIndexAbstract<DistType> {
                                                     VecSimQueryResult_Code *rc) const;
     virtual VecSimQueryResult_List topKQuery(const void *queryBlob, size_t k,
                                              VecSimQueryParams *queryParams) override;
-    VecSimQueryResult_List rangeQuery(const void *queryBlob, DistType radius,
+    VecSimQueryResult_List rangeQuery(const void *queryBlob, double radius,
```
DvirDukhan, line 36: Please remind me why we moved to `double` from `DistType`?

src/VecSim/algorithms/hnsw/hnsw.h
```diff
@@ -98,8 +98,9 @@ class HNSWIndex : public VecSimIndexAbstract<DistType> {
 #endif
 
 #ifdef BUILD_TESTS
-    friend class HNSWIndexSerializer;
     // Allow the following test to access the index size private member.
+    template <typename, typename>
```
DvirDukhan, line 102:
```suggestion
    template <DataType, DistType>
```

meiravgri: This wont compile with 
error: "declaration of template parameter ‘DataType’ shadows template parameter"
error: "declaration of template parameter ‘DistType’ shadows template parameter"
since it is the same as the class's template type names

src/VecSim/algorithms/brute_force/brute_force.h
meiravgri: VecSimIndexInterface is not templates, so its inheritors must implement exactly VecSimQueryResult_List rangeQuery(... double radius)
hence for floats it gives an error

src/python_bindings/bindings.cpp
```diff
@@ -125,10 +123,13 @@ class PyHNSWLibIndex : public PyVecSimIndex {
         hnsw->setEf(ef);
     }
     void saveIndex(const std::string &location) {
+        VecSimIndexInfo info = this->index->info();
```
DvirDukhan, line 126: remove?

src/python_bindings/bindings.cpp
```diff
@@ -125,10 +123,13 @@ class PyHNSWLibIndex : public PyVecSimIndex {
...
         auto serializer = HNSWIndexSerializer(reinterpret_cast<HNSWIndex<float, float> *>(index));
         serializer.saveIndex(location);
     }
+
     void loadIndex(const std::string &location) {
+        VecSimIndexInfo info = this->index->info();
```
DvirDukhan, line 132: remove?

DvirDukhan (changes requested):

DvirDukhan (approved):

PR #228, opened 2022-08-31
forcing setuptools into the conda env

chayim:


alonre24:
I was able to fix it here  - #229 
Should we close this one? @chayim

PR #232, opened 2022-09-06
[BUG] fix for hnsw break condition

GuyAv46:
When the entry point for the base layer is close to the query vector, the lower bound might be set to a too low value, and therefore we might not get `ef` or even `k` results in the `top_candidates` heap.


GuyAv46:
After further investigation, this is not a bug on single value HNSW so there is no need to add this change to `main` and older versions.

PR #257, opened 2022-10-13
Epsilon runtime parameter MOD-4165

DvirDukhan:

Adds epsilon query parameter for HNSW runtime query parameters

1. MOD-4165

1. `VecSimIndex_ResolveParams` signature now hold the query type instead of boolean flag for hybrid query

**Mark if applicable**


src/VecSim/vec_sim.cpp
```diff
@@ -48,9 +49,31 @@ static VecSimResolveCode _ResolveParams_BatchSize(VecSimRawParam rparam, VecSimQ
     return VecSimParamResolver_OK;
 }
 
+static VecSimResolveCode _ResolveParams_Epsilon(VecSimAlgo index_type, VecSimRawParam rparam,
+                                                VecSimQueryParams *qparams,
+                                                VecsimQueryType query_type) {
+    double num_val;
+    // EF_RUNTIME is a valid parameter only in HNSW algorithm.
```
alonre24, line 56:
```suggestion
    // EPSILON is a valid parameter only in HNSW algorithm.
```

src/VecSim/vec_sim.cpp
```diff
@@ -48,9 +49,31 @@ static VecSimResolveCode _ResolveParams_BatchSize(VecSimRawParam rparam, VecSimQ
...
+    if (index_type != VecSimAlgo_HNSWLIB) {
+        return VecSimParamResolverErr_UnknownParam;
+    }
+    if (query_type != QUERY_TYPE_RANGE) {
+        return VecSimParamResolverErr_InvalidPolicy_NRange;
```
alonre24, line 61: Can we cover this also?

tests/unit/test_utils.h
```diff
@@ -55,6 +55,8 @@ inline VecSimIndex *CreateNewIndex(IndexParams &index_params, VecSimType type,
     return VecSimIndex_New(&params);
 }
 
+extern VecsimQueryType query_types[4];
```
alonre24, line 58: Why using `extern` instead of defining the `query_types` here?

alonre24 (commented):
few small comments. Extension of the param resolve tests looks good!

tests/unit/test_utils.h
DvirDukhan: Because it is a global array on `h` file. If included multiple times, there will be multiple definitions of the same array

alonre24 (approved):

PR #272, opened 2022-11-02
upgraded mac vesrion

meiravgri:


meiravgri:
wont be relevant after #283

PR #278, opened 2022-11-06
Introduce tiered HNSW index MOD-4300 MOD-4299

alonre24:
- Introduce abstract `VecSimTieredIndex` that inherits from `VecSimIndexInterface`, and its `TieredHNSWIndex` implementation. 
- Add an API in HNSW factory that creates `TieredHNSWIndex` based on combined params (for HNSW and for the async tiered index properties) 
- `VecSimTieredIndex` contains a pointer to BF index, and creates it as part of the object constructor, based on the main index params (this had required adding two getters to `VecSimIndexAbstract`).
- Add the required data structures and callbacks to the common defines (that is, the "job" definition, submit to queue callback etc...)
- Add a new unit test suit, that uses "mock" job queue and callbacks to allow testing of the proper creation of the  `TieredHNSWIndex` (to be used later on in tests...)

**Notes**
- This PR include upgrading our build to c++20 standard (to be suitable with `<shared_mutex>` object)
- All the pure virtual `VecSimIndexInterface` functions were to be implemented in order to create an instance. These functions (and some other signature functions) are (temporary) not covered by tests.


src/VecSim/vec_sim_tiered_index.h
```diff
@@ -0,0 +1,41 @@
...
+class VecSimTieredIndex : public VecSimIndexInterface {
+protected:
+    BruteForceIndex<DataType, DistType> *tempFlat;
+    VecSimIndexAbstract<DistType> *index;
+
+    void *jobQueue;
+    SubmitCB SubmitJobsToQueue;
+
+    void *memoryCtx;
+    UpdateMemoryCB UpdateIndexMemory;
```
GuyAv46, line 18: Consider documenting what are the expectations from these callbacks and pointers (here or with the type definitions)

src/VecSim/algorithms/hnsw/hnsw_factory.cpp
```diff
@@ -119,4 +120,22 @@ size_t EstimateElementSize(const HNSWParams *params) {
...
+    if (params->hnswParams.type == VecSimType_FLOAT32) {
+        auto *hnsw_index =
+            NewIndex_ChooseMultiOrSingle<float, float>(&params->hnswParams, allocator);
+        return new (allocator) TieredHNSWIndex<float, float>(
+            hnsw_index, params->jobQueue, params->submitCb, params->memoryCtx, params->UpdateMemCb);
+    } else if (params->hnswParams.type == VecSimType_FLOAT64) {
+        auto *hnsw_index =
+            NewIndex_ChooseMultiOrSingle<double, double>(&params->hnswParams, allocator);
+        return new (allocator) TieredHNSWIndex<double, double>(
+            hnsw_index, params->jobQueue, params->submitCb, params->memoryCtx, params->UpdateMemCb);
```
GuyAv46, line 135: Why not just pass `&params->hnswParams` to the hnsw new function? this way we'll have to update both this and the HNSW new function when support for other types will be added

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -0,0 +1,65 @@
...
+                    void *memory_ctx, UpdateMemoryCB UpdateMemCb)
+        : VecSimTieredIndex<DataType, DistType>(hnsw_index, job_queue_, submitCb, memory_ctx,
+                                                UpdateMemCb) {}
+    virtual ~TieredHNSWIndex() = default;
+
+    // TODO: Implement the actual methods instead of these temporary ones.
+    int addVector(const void *blob, labelType id) override {
+        return this->index->addVector(blob, id);
+    }
+    int deleteVector(labelType id) override { return this->index->deleteVector(id); }
```
GuyAv46, line 39: So long for async...

GuyAv46 (commented):
Looks very good!
Consider reorganizing the hnsw directory into subdirectories (tiered, batch...)

src/VecSim/vec_sim_common.h
```diff
@@ -83,11 +90,30 @@ typedef struct {
...
+    void *memoryCtx;
+    UpdateMemoryCB UpdateMemCb;
+} TieredHNSWParams;
+
+typedef enum { HNSW_INSERT_JOB, HNSW_DELETE_JOB } HNSWJobType;
+
+typedef struct HNSWJob {
+    HNSWJobType jobType;
+    labelType label;
+    void *blob;
```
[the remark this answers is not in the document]
alonre24: I guess... we can adjust this struct later on for multi if we see that we'll need to...

src/VecSim/vec_sim_tiered_index.h
alonre24: I added docus in `vec_sim_common.h`

GuyAv46 (approved):
LGTM

src/VecSim/vec_sim_common.h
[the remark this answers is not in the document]
alonre24: After revising, the insert job struct should contain the internal id as well, so we can access each blob in a MULTI index for moving it from the flat into the HNSW index as part of every separate job.

src/VecSim/vec_sim_common.h
```diff
@@ -83,11 +90,34 @@ typedef struct {
...
+    UpdateMemoryCB
+        UpdateMemCb; // A callback that updates the memoryCtx with a given memory (number).
+} TieredHNSWParams;
+
+typedef enum { HNSW_INSERT_JOB, HNSW_DELETE_JOB } HNSWJobType;
+
+/**
+ * Definition of HNSW job structure for asynchronous tiered index.
+ */
+typedef struct HNSWJob {
```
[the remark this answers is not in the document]
alonre24: Changed to generic

src/VecSim/vec_sim_tiered_index.h
```diff
@@ -0,0 +1,41 @@
...
+
+public:
+    VecSimTieredIndex(VecSimIndexAbstract<DistType> *index_, void *job_queue_, SubmitCB submitCb,
+                      void *memory_ctx, UpdateMemoryCB UpdateMemCb)
+        : VecSimIndexInterface(index_->getAllocator()), index(index_), jobQueue(job_queue_),
+          SubmitJobsToQueue(submitCb), memoryCtx(memory_ctx), UpdateIndexMemory(UpdateMemCb) {
+        BFParams bf_params = {.type = index_->getType(),
+                              .dim = index_->getDim(),
+                              .metric = index_->getMetric(),
+                              .multi = index_->isMultiValue()};
```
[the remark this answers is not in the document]
alonre24: So you propose a change, or just wanted to raise awareness?

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -0,0 +1,65 @@
...
+typedef struct UpdateMessage {
+    HNSWJobType type;
+    // ...
+} UpdateMessage;
+
+template <typename DataType, typename DistType>
+class TieredHNSWIndex : public VecSimTieredIndex<DataType, DistType> {
+private:
+    // Todo: implement these methods later on
+    void executeJob(HNSWJob *job) {}
```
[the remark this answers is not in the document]
alonre24: It has changed in the meantime, but the intention is to implement every specific job's callback directly in the derived async index that uses that job...

src/VecSim/algorithms/brute_force/brute_force.h
```diff
@@ -16,6 +16,7 @@
 #include <queue>
 #include <cassert>
 #include <limits>
+#include <stdexcept>
```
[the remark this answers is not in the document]
alonre24: redundant...

src/VecSim/vec_sim_common.h
```diff
@@ -92,14 +99,80 @@ typedef struct {
     size_t blockSize;
 } BFParams;
 
+// A struct that contains the common tiered index params.
+typedef struct {
+    void *jobQueue;    // External queue that holds the jobs.
+    SubmitCB submitCb; // A callback that submits an array of jobs into a given jobQueue.
+    void *memoryCtx;   // External context that stores the index memory consumption.
+    UpdateMemoryCB
```
DvirDukhan, line 107: is this the result of `make format`?

src/VecSim/vec_sim_tiered_index.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include "vec_sim_index.h"
+#include "algorithms/brute_force/brute_force.h"
+
+#include <shared_mutex>
+
+template <typename DataType, typename DistType>
+class VecSimTieredIndex : public VecSimIndexInterface {
+protected:
+    BruteForceIndex<DataType, DistType> *tempFlat;
```
DvirDukhan, line 11: maybe consider rename to `flatBuffer`

src/VecSim/algorithms/hnsw/hnsw_factory.h
```diff
@@ -16,6 +16,8 @@
 namespace HNSWFactory {
 
 VecSimIndex *NewIndex(const HNSWParams *params, std::shared_ptr<VecSimAllocator> allocator);
+VecSimIndex *NewTieredIndex(const TieredHNSWParams *params,
```
DvirDukhan, line 19: who calls this function right now other than the tests?

src/VecSim/vec_sim_tiered_index.h
```diff
@@ -0,0 +1,41 @@
...
+    void *memoryCtx;
+    UpdateMemoryCB UpdateIndexMemory;
+
+    // Consider putting these in the derived class instead. Also - see if we should use
+    // std::shared_mutex
+    std::shared_timed_mutex flatIndexGuard;
+    std::shared_timed_mutex mainIndexGuard;
+
+public:
+    VecSimTieredIndex(VecSimIndexAbstract<DistType> *index_, void *job_queue_, SubmitCB submitCb,
```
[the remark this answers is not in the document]
DvirDukhan: this comment still relevant

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -0,0 +1,79 @@
...
+        VecSimIndex_AddVector(tiered_index->tempFlat, vector, vector_label);
+
+        // Create a mock job that inserts some vector into the HNSW index.
+        auto insert_to_index = [](void *job) {
+            auto *my_insert_job = reinterpret_cast<HNSWInsertJob *>(job);
+            auto my_index =
+                reinterpret_cast<TieredHNSWIndex<TEST_DATA_T, TEST_DIST_T> *>(my_insert_job->index);
+
+            // Move the vector from the temp flat index into the HNSW index.
+            my_index->addVector(my_index->tempFlat->getDataByInternalId(my_insert_job->id),
```
DvirDukhan, line 44: add comment explaining we are addign by internal id due to multi value index

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -0,0 +1,79 @@
...
+                                my_insert_job->label);
+            VecSimIndex_DeleteVector(my_index->tempFlat, my_insert_job->label);
```
DvirDukhan, line 46: delete by internal id - new PR

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -0,0 +1,79 @@
...
+            auto it = my_index->labelToInsertJobs[my_insert_job->label].begin();
```
DvirDukhan, line 47: explain this flow happens in a fixture that expects this task already be in the queue
in other words
this fixture tests the `labelToInsertJobs` correctness

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -0,0 +1,79 @@
...
+            my_index->labelToInsertJobs[my_insert_job->label].erase(it);
+            my_index->UpdateIndexMemory(my_index->memoryCtx,
+                                        my_index->getAllocator()->getAllocationSize());
+        };
+
+        HNSWInsertJob job = {
+            .base = AsyncJob{.jobType = HNSW_INSERT_VECTOR_JOB, .Execute = insert_to_index},
+            .index = tiered_index,
+            .label = vector_label};
+        tiered_index->labelToInsertJobs[vector_label].push_back(&job);
```
DvirDukhan, line 58: don't we need a specific API call for this, not breaking abstraction? 
maybe add TODO to push this logic into the `add_vector` of the tiered ineex

src/VecSim/vec_sim_common.h
alonre24: I guess... I'll reorganize it

src/VecSim/vec_sim_common.h
```diff
@@ -92,14 +99,80 @@ typedef struct {
...
         HNSWParams hnswParams;
         BFParams bfParams;
+        TieredHNSWParams tieredHNSWParams;
     };
 } VecSimParams;
 
+/**
+ * The specific job types in use (to be extended in the future by demand)
+ */
+typedef enum {
```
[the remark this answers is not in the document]
alonre24: The idea was to put all the definitions in this PR, and split the implementation into multiple tasks...
I guess that we can use `UNDEFINED` or `INVALID_JOB` for the readability of future functions that will check what is the job type.

src/VecSim/algorithms/hnsw/hnsw_factory.h
alonre24: Currently no one. In the future, it should be called from `VecSimIndex_New`. Should it be part of this PR?

src/VecSim/vec_sim_common.h
```diff
@@ -129,7 +129,8 @@ typedef enum {
     HNSW_INSERT_VECTOR_JOB,
     HNSW_REPAIR_NODE_CONNECTIONS_JOB,
     HNSW_SEARCH_JOB,
-    HNSW_SWAP_JOB
+    HNSW_SWAP_JOB,
+    INVALID_JOB // to indicate that finding a JobType >= INVALID_JOB is an error
```
DvirDukhan, line 133: I think this should be the first one

tests/unit/CMakeLists.txt
```diff
@@ -13,8 +13,8 @@ include(CTest)
 
 set(CMAKE_CXX_STANDARD 20)
 
-set(CMAKE_CXX_FLAGS_DEBUG "-g")
-set(CMAKE_CXX_FLAGS_RELEASE "-O3")
+set(CMAKE_CXX_FLAGS_DEBUG "-g -pthread")
+set(CMAKE_CXX_FLAGS_RELEASE "-O3 -pthread")
```
DvirDukhan, line 17: are we using pthreads here?

src/VecSim/vec_sim_common.h
alonre24: why?

tests/unit/CMakeLists.txt
alonre24: I thought it was necessary for using the `Thread` object in the test but looks like it doesn't. removing...

DvirDukhan (approved):

PR #283, opened 2022-11-16
GH actions CI

DvirDukhan:


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,107 @@
...
+          aws-region: ${{ secrets.AWS_REGION }}
+      - name: Start EC2 runner
+        id: start-ec2-runner
+        uses: machulav/ec2-github-runner@v2
+        with:
+          mode: start
+          github-token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
+          # Ubuntu 22.04 128GB Storage AMI
+          ec2-image-id: ami-0ba430d4b7b64de57
+          ec2-instance-type: r6i.large
```
chayim, line 27: Maybe this should be outside of the workflow in a secret, purely so that you can change sizes without a commit?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,107 @@
...
+    name: Run the benchmarks on the runner
+    needs: start-runner # required to start the main job when the runner is ready
+    runs-on: ${{ needs.start-runner.outputs.label }} # run the job on the newly created runner
+    steps:
+      - name: checkout
+        uses: actions/checkout@v3
+
+      - name: Print runner info
+        run: |
+          printf "Runner lscpu:\n$(lscpu)\n"
```
chayim, line 41: Why printf and not just calling the commands directly with echo >> $file?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,107 @@
...
+          printf "Runner lsmem:\n$(lsmem)\n"
+          printf "Runner nproc:\n$(nproc)\n"
+          printf "Runner uname:\n$(uname -a)\n"
+
+      - name: Install benchmark dependencies
+        run: |
+             sudo .install/install_script.sh
+             sudo apt install python3-pip -y
+             pip3 install --upgrade pip PyYAML setuptools
```
chayim, line 50: Any reason not to combine the pip lines... or even have your install script do this?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,107 @@
...
+          uptime
+
+      # TODO: remove "--no-check-certificate" when possible
+      - name: Download pre-generated indices
+        timeout-minutes: 20
+        run: wget --no-check-certificate -q -i tests/benchmark/data/hnsw_indices.txt -P tests/benchmark/data
+
+      - name: Benchmark
+        timeout-minutes: 120
+        run: bash -l -c "make benchmark"
```
chayim, line 72: Why have bash retrigger a login? I think make benchmark should work fine here, no? This already runs in a new shell.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,107 @@
...
+          --redistimeseries_port      ${{ secrets.PERFORMANCE_RTS_PORT }}           \
+          --redistimeseries_pass      ${{ secrets.PERFORMANCE_RTS_AUTH }}           \
+          --github_repo               ${{ github.event.repository.name }}           \
+          --github_org                ${{ github.repository_owner }}                \
+          --github_branch             ${{ github.head_ref || github.ref_name }}     \
+          --github_actor              ${{ github.triggering_actor }}                \
+          --results-format            google.benchmark                              \
+          --benchmark-result-file     {}_results.json
+
+  stop-runner:
```
chayim, line 87: Given that this is used so often.. I wonder if you can pull it into a reusable workflow (see redis stack)

.github/workflows/default.yml
```diff
@@ -0,0 +1,49 @@
...
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}
+      cancel-in-progress: true
+    steps:
+      - name: checkout
+        uses: actions/checkout@v3
+      - name: setup python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.8'
```
chayim, line 23: pull python-version into an environment variable and use that.
Also, any reason to not just jump (or at least try) 3.11/latest?

.github/workflows/jammy.yml
```diff
@@ -0,0 +1,23 @@
+name: jammy flow
+
+on: workflow_call # TODO: add pull_request-labeled with "jammy" label when specifying a label is supported
+
+jobs:
```
chayim, line 5: IMHO this shows you can do the reusable workflow thing. Less to maintain.

.install/macos.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+brew install make cmake
```
chayim, line 2: You're not getting a guaranteed version of cmake here. Why not use the same mechansim as Linux?

.install/ubuntu_20.04.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -e
+export DEBIAN_FRONTEND=noninteractive
```
chayim, line 3: This line should be in all ubuntu scripts.

.github/workflows/default.yml
DvirDukhan: why not check on the most primitive one?

.github/workflows/benchmark.yml
DvirDukhan: prefer to have it visible for better debug and visibility

.github/workflows/benchmark.yml
DvirDukhan: to see in logs I guess. maybe @GuyAv46 can elaborate here

GuyAv46: Filipe suggested we will put this data in a time series. For now we just print it but this will be changed

.github/workflows/default.yml
chayim: Only because they're tools - for tooling I tend to try and use latest, but it's a preference, not a requirement.

chayim (approved):

chayim (approved):
Approving my parts

PR #298, opened 2022-12-08
fixed get AVX512F functions MOD-4593

meiravgri:
Before this PR, for fp64, we swapped between the AVX512F distance calculation function of 2 cases:
1. the dimension is a multiplier of 2, 
or
2. The dimension is bigger than 8 and the residual is smaller than 2
So, for example for dim = 12 ( % 2 == 0 )
We were calculating the first 8 numbers as a batch, and the rest 4 numbers one by one (this will yield the correct result but less optimal than calculating the residual in batches of 2.
for dim = 13 we were taking the first 8 and then calculating the rest in batches of 2, which means that the last batch exceeds the valid dimension.

1. Returning the wrong optimization function when using AVX512 with fp64

1. FP64_GetDistFunc in L2 and IP spaces

**Mark if applicable**

### DIFF
--- a/tests/unit/test_spaces.cpp
+++ b/tests/unit/test_spaces.cpp
@@ -99,6 +99,7 @@ TEST_F(SpacesTest, double_ip_no_optimization_func_test) {
 #ifdef CPU_FEATURES_ARCH_X86_64
 
 using spaces::dist_func_t;
+
 namespace spaces_test {
 // Each array contains all the possible architecture optimization related to one dimension
 // optimization. For example: L2_dist_funcs_16Ext[ARCH_OPT_NONE = 0] = FP32_L2Sqr,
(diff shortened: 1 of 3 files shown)


DvirDukhan (commented):
Is this testable?
Also please fix the PR description

tests/unit/test_spaces.cpp
```diff
@@ -163,6 +164,150 @@ static dist_func_t<double> IP_dist_funcs_2ExtResiduals[] = {
     FP64_InnerProductSIMD2ExtResiduals_AVX512};
 } // namespace spaces_test
 
+TEST_F(SpacesTest, fp32_L2_GetDistFunc) {
+    using namespace spaces_test;
+    using namespace spaces;
+    // fp32_L2_dist_funcs_array[i][j] is a pointer to a function that match CalculationGuideline[i]
+    // and architecture optimization Arch_Optimization[j]
+    static dist_func_t<float> *fp32_L2_dist_funcs_array[] = {
+        nullptr, L2_dist_funcs_16Ext, L2_dist_funcs_4Ext, L2_dist_funcs_16ExtResiduals,
```
DvirDukhan, line 173: why the first entry is null?

tests/unit/test_spaces.cpp
```diff
@@ -163,6 +164,150 @@ static dist_func_t<double> IP_dist_funcs_2ExtResiduals[] = {
...
+    /* enum Arch_Optimization {
+    ARCH_OPT_NONE,
+    ARCH_OPT_SSE,
+    ARCH_OPT_AVX,
+    ARCH_OPT_AVX512_F,
+    ARCH_OPT_AVX512_DQ // Relevant only to fp64
+}; */
+    // For each arch
+    for (int arch = ARCH_OPT_NONE; arch <= ARCH_OPT_AVX512_F; ++arch) {
+        ASSERT_EQ(L2_FP32_GetDistFunc(dimensions[NO_OPTIMIZATION], (Arch_Optimization)arch),
```
DvirDukhan, line 191: can you add this logic to the loop?

tests/unit/test_spaces.cpp
meiravgri: i find it less confusing then
`fp32_L2_dist_funcs_array[dim_opt + 1][arch]`

tests/unit/test_spaces.cpp
meiravgri: let's say we remove the nullptr and the inner loop begins with dim_opt = NO_OPTIMIZATION
then the first entire loop is ok,
but in the second iteration (arch = 1) we get 
`ASSERT_EQ(L2_FP32_GetDistFunc(dimensions[0], (Arch_Optimization)1), //should be fp32_L2Sqr
                      fp32_L2_dist_funcs_array[0][1] // SSE`

DvirDukhan (approved):

PR #299, opened 2022-12-08
HNSW refactor - store data in blocks - MOD-4550

GuyAv46:

This PR attempts to solve the issue [MOD-4550](https://redislabs.atlassian.net/browse/MOD-4550).

Before introducing this PR, we allocated the entire index in a single allocation. When the index acquired close to half of the available memory and had to resize, the `realloc` call would fail.

This PR intrudes DataBlocks to HNSW. They are the same data structures used in the FLAT algorithm. Every resize call is now allocating a new DataBlock for the vectors and a new DataBlock for the graph data (links and incoming links for every level).
This PR also replaces the offsets method that is currently used with structures to simplify the code and unify different meta-data access.

**NOTICE**
When this PR is applied to RediSearch, we should update HNSW build validation (like FLAT, check that the index + 1 block can fit in the memory).
We should also expose `BLOCK_SIZE` in HNSW parameters.

1. Moving `VectorBlock` from BF to utils and renaming to `DataBlock`
2. Adding structures for element meta-data.
3. Using DataBlocks in HNSW.
4. Deprecating V1 and V2 serialization files (should be re-serialized).

**Mark if applicable**

- [x] This PR introduces serialization changes


tests/unit/test_hnsw.cpp
```diff
@@ -876,8 +876,6 @@ TYPED_TEST(HNSWTest, hnsw_bad_params) {
         1,          // Will fail because 1/log(M).
         100000000,  // Will fail on M * 2 overflow.
         UINT16_MAX, // Will fail on M * 2 overflow.
-        UINT16_MAX /
-            2, // Will fail on this->allocator->callocate(max_elements_ * size_data_per_element_)
```
GuyAv46, line 880: No longer needed - we don't allocate the data on creation

alonre24 (commented):
WIP

PR #319, opened 2022-12-27
Refactor add/delete API as an enabler for tiered index

alonre24:
- Remove "unmark deleted" API, as it is currently redundant for our async index
 - Let add and delete vector APIs in the index interface to return the number of vectors added/deleted (and take into consideration overwrites).
 - Add an `override_allowed` flag to `addVector` interface, so we can enforce that the internal HNSW index within the tiered index will not be able to override existing vectors (rather than creating a delete job in advance).
 - Add `indexCapacity()` and `resize()` to index interface and adjust the existing indexes accordingly. Call resize *before* we add new vectors externally in `vec_sim.cpp`

### DIFF
--- a/src/VecSim/algorithms/hnsw/hnsw_single.h
+++ b/src/VecSim/algorithms/hnsw/hnsw_single.h
@@ -154,8 +154,8 @@ std::vector<idType> HNSWIndex_Single<DataType, DistType>::markDelete(labelType l
     if (search == label_lookup_.end()) {
         return idsToDelete;
     }
-    label_lookup_.erase(search);
     this->markDeletedInternal(search->second);
     idsToDelete.push_back(search->second);
+    label_lookup_.erase(search);
     return idsToDelete;
 }
--- a/src/VecSim/algorithms/hnsw/hnsw_single.h
+++ b/src/VecSim/algorithms/hnsw/hnsw_single.h
@@ -52,10 +52,9 @@ public:
                                           VecSimQueryParams *queryParams) const override;
 
     int deleteVector(labelType label) override;
-    int addVector(const void *vector_data, labelType label) override;
+    int addVector(const void *vector_data, labelType label, bool override_allowed = true) override;
     double getDistanceFrom(labelType label, const void *vector_data) const override;
-    inline int markDelete(labelType label) override;
-    inline int unmarkDelete(labelType label) override;
+    inline std::vector<idType> markDelete(labelType label) override;
 };
 
 /**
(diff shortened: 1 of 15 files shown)


src/VecSim/vec_sim_interface.h
```diff
@@ -69,6 +72,18 @@ struct VecSimIndexInterface : public VecsimBaseObject {
...
+     * @brief Return the index capacity, so we know if resize is required for adding new vectors.
+     *
+     * @return index capacity.
+     */
+    virtual size_t indexCapacity() const = 0;
+
+    /**
+     * @brief Change the index capacity (without chaning its data), by adding another block.
+     */
+    virtual void resize() = 0;
```
DvirDukhan, line 85: `resize` is ambiguous 
consider a name like `resizeAddBlock`

src/VecSim/algorithms/hnsw/hnsw_single.h
```diff
@@ -98,25 +97,35 @@ void HNSWIndex_Single<DataType, DistType>::resizeLabelLookup(size_t new_max_elem
...
+        } else {
+            // If override is not allowed, we don't do anything.
+            return -1;
+        }
     }
 
-    return this->appendVector(vector_data, label);
+    this->appendVector(vector_data, label);
+    // Return the delta in the index size due to the insertion.
+    return this->indexSize() - index_size_before;
```
DvirDukhan, line 128: why not 1?

src/VecSim/vec_sim_interface.h
```diff
@@ -37,18 +37,21 @@ struct VecSimIndexInterface : public VecsimBaseObject {
...
-     * @param id the id of the added vector
-     * @return always returns true
+     * @param label the label of the added vector.
+     * @param overrideAllowed if true and id already exists in the index, override it. Otherwise,
+     * ignore the new vector.
+     * @return the number of new vectors inserted (1 for new insertion, 0 for override), or -1
+     * in case that override is not allowed and label already exists.
      */
-    virtual int addVector(const void *blob, labelType id) = 0;
+    virtual int addVector(const void *blob, labelType label, bool overrideAllowed = true) = 0;
```
DvirDukhan, line 46:
```suggestion
    virtual int addVector(const void *blob, labelType label, bool overwriteAllowed = true) = 0;
```

tests/unit/test_hnsw.cpp
```diff
@@ -2093,13 +2097,14 @@ TYPED_TEST(HNSWTest, markDelete) {
...
+    // Re-add the previously marked vectors (under new internal ids).
+    for (labelType label = 0; label < n; label++) {
+        if (label % 2 == ep_reminder) {
+            GenerateAndAddVector<TEST_DATA_T>(index, dim, label, label);
+        }
+    }
 
-    ASSERT_EQ(this->CastToHNSW(index)->getNumMarkedDeleted(), 0);
-    ASSERT_EQ(VecSimIndex_IndexSize(index), n + 1);
+    ASSERT_EQ(VecSimIndex_IndexSize(index), n + n / 2 + 1);
```
DvirDukhan, line 2107: also add validation that the number of the vectors that are marked as deleted is still n/2

tests/unit/test_hnsw_multi.cpp
```diff
@@ -1725,21 +1729,5 @@ TYPED_TEST(HNSWMultiTest, markDelete) {
...
-    ASSERT_EQ(VecSimIndex_IndexSize(index), n + 1);
-
-    // Search for k results around the middle again. expect to find the same results we found in the
-    // first search.
-    runTopKSearchTest(index, query, k, verify_res);
-    runRangeQueryTest(index, query, dim * all_element * all_element, verify_res, k, BY_SCORE);
-    batchIterator = VecSimBatchIterator_New(index, query, nullptr);
-    runBatchIteratorSearchTest(batchIterator, k, verify_res);
-    VecSimBatchIterator_Free(batchIterator);
-
```
DvirDukhan, line 1743: no need for this?

src/VecSim/algorithms/hnsw/hnsw_single.h
alonre24: Because in case of override, we should return 0

src/VecSim/vec_sim_interface.h
alonre24: How about `increaseCapacity`?

tests/unit/test_hnsw_multi.cpp
alonre24: I thought it was redundant after `unmarkDelete` was removed. But on second thought, I'll replace this with reinserting the vectors as I did in HNSW_single.

DvirDukhan (commented):
please change function comments from `override` to `overwrite`

DvirDukhan (dismissed):

DvirDukhan (approved):

PR #329, opened 2023-02-09
Tiered index insert job MOD-4321

alonre24:

Implement the asynchronous insert job - move a vector from the flat index into HNSW index. This doesn't support overwrites (but the code contains some preparations for this case, such as checking if a label is INVALID, which means it was deleted in the meantime).
Also, to allow the deletion of a single vector that is stored under a label in MULTI index, we add an API to BF that enables deleting a vector by its internal id.
In addition, this included changing the `overwrite_allowed` flag in the `addVector` API into `new_vec_id` which allows specifying the new vector id, in cases where we use a tiered index and we set the new vector's internal id in advanced (otherwise, we use a fresh id and send -1 as default).


meiravgri (commented):
Looks good 🤩 
Few comments
I would add some more documentation about each lock and its purpose
Also, the mechanism of invalidating a job is unclear to me, especially the fact that it can happen during and at any stage of the execution
Consider adding more tests to the edge cases such as removing the last vector, job invalidation etc

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -119,6 +121,90 @@ void TieredHNSWIndex<DataType, DistType>::executeInsertJobWrapper(void *job) {
...
+
+    if (job->label == HNSW_INVALID_LABEL) {
+        // Job has been invalidated in the meantime.
+        this->flatIndexGuard.unlock_shared();
+        this->mainIndexGuard.unlock_shared();
+        goto finish;
+    }
+
+    // Take the vector from the flat buffer and insert it to HNSW (overwrite should not occur).
+    hnsw_index->addVector(this->flatBuffer->getDataByInternalId(job->id), job->label, false);
```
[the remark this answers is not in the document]
alonre24: For searching the flat index

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -119,6 +121,90 @@ void TieredHNSWIndex<DataType, DistType>::executeInsertJobWrapper(void *job) {
...
+void TieredHNSWIndex<DataType, DistType>::executeInsertJob(HNSWInsertJob *job) {
+    HNSWIndex<DataType, DistType> *hnsw_index = this->getHNSWIndex();
+    this->flatIndexGuard.lock_shared();
+    if (job->label == HNSW_INVALID_LABEL) {
+        // Job has been invalidated in the meantime.
+        this->flatIndexGuard.unlock_shared();
+        goto finish;
+    }
+    // Acquire the index data lock, so we know what is the exact index size at this time.
+    this->mainIndexGuard.lock_shared();
```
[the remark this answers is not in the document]
alonre24: Because of the `addVector` call to HNSW that should be protected with a read lock. Also, to avoid deadlocks, we always acquire the main index lock *before* we take internal HNSW locks.

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -119,6 +121,90 @@ void TieredHNSWIndex<DataType, DistType>::executeInsertJobWrapper(void *job) {
...
+        this->flatIndexGuard.unlock_shared();
+        this->mainIndexGuard.unlock_shared();
+        this->mainIndexGuard.lock();
+        hnsw_index->increaseCapacity();
+        this->mainIndexGuard.unlock();
+        // reacquire the read locks
+        this->flatIndexGuard.lock_shared();
+        this->mainIndexGuard.lock_shared();
+    }
+    hnsw_index->index_data_guard_.unlock();
```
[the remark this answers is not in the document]
alonre24: Added more comments

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -119,6 +121,90 @@ void TieredHNSWIndex<DataType, DistType>::executeInsertJobWrapper(void *job) {
...
+HNSWIndex<DataType, DistType> *TieredHNSWIndex<DataType, DistType>::getHNSWIndex() {
+    return reinterpret_cast<HNSWIndex<DataType, DistType> *>(this->index);
+}
+
+/******************** Job's callbacks **********************************/
+template <typename DataType, typename DistType>
+void TieredHNSWIndex<DataType, DistType>::executeInsertJob(HNSWInsertJob *job) {
+    HNSWIndex<DataType, DistType> *hnsw_index = this->getHNSWIndex();
+    this->flatIndexGuard.lock_shared();
+    if (job->label == HNSW_INVALID_LABEL) {
```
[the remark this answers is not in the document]
alonre24: This flag is set when we overwrite a vector that is in the flat index and has a pending insert job (let's discuss it).

src/VecSim/algorithms/hnsw/hnsw_single.h
```diff
@@ -124,19 +124,14 @@ int HNSWIndex_Single<DataType, DistType>::addVector(const void *vector_data, con
                                                     bool overwrite_allowed) {
 
     // Checking if an element with the given label already exists.
-    std::unique_lock<std::mutex> index_data_lock(this->index_data_guard_);
     bool label_exists = false;
-    if (label_lookup_.find(label) != label_lookup_.end()) {
-        label_exists = true;
-        if (overwrite_allowed) {
+    if (overwrite_allowed) {
```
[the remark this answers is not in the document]
alonre24: It the caller responsibility to ensure that this label doesn't exist in the index before calling this (in our case the tiered index), and the purpose is that we won't access the `label_lookup_` for nothing (also, had we access this map, we had to lock)

src/VecSim/algorithms/hnsw/hnsw.h
```diff
@@ -1381,7 +1402,11 @@ void HNSWIndex<DataType, DistType>::appendVector(const void *vector_data, const
 
     // Access and update the index global data structures with the new vector.
     std::unique_lock<std::mutex> index_data_lock(index_data_guard_);
```
alonre24, line 1404: In the tiered index scenario, we lock the `index_data_lock` here for the second time (the first time is when we decide whether we should resize the index or not). Consider moving all the logic here to the tiered index and avoid the two times locking...

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -137,48 +137,44 @@ void TieredHNSWIndex<DataType, DistType>::executeInsertJob(HNSWInsertJob *job) {
...
         this->flatIndexGuard.unlock_shared();
-        this->mainIndexGuard.unlock_shared();
 
         this->mainIndexGuard.lock();
+        hnsw_index->lockIndexDataGuard();
         // Check if resizing is still required (another thread might have done it in the meantime
         // while we release the shared lock).
         if (hnsw_index->indexCapacity() == hnsw_index->indexSize()) {
             hnsw_index->increaseCapacity();
         }
```
meiravgri, line 157: you can move `this->flatIndexGuard.unlock_shared()` here

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -137,48 +137,44 @@ void TieredHNSWIndex<DataType, DistType>::executeInsertJob(HNSWInsertJob *job) {
         return;
     }
     // Acquire the index data lock, so we know what is the exact index size at this time.
-    // To avoid deadlocks, we always acquire the main index lock *before* we take internal HNSW
-    // locks.
-    this->mainIndexGuard.lock_shared();
     hnsw_index->lockIndexDataGuard();
-    bool unlock_hnsw_data_guard_required = true;
+    idType new_vec_id = hnsw_index->indexSize();
```
meiravgri, line 141: We can take the new id once, after resize checking and performing if needed

src/VecSim/algorithms/hnsw/hnsw_single.h
```diff
@@ -62,7 +62,7 @@ class HNSWIndex_Single : public HNSWIndex<DataType, DistType> {
                                           VecSimQueryParams *queryParams) const override;
 
     int deleteVector(labelType label) override;
-    int addVector(const void *vector_data, labelType label, bool overwrite_allowed = true) override;
+    int addVector(const void *vector_data, labelType label, idType new_vec_id = -1) override;
```
meiravgri, line 65:
```suggestion
    int addVector(const void *vector_data, labelType label, idType new_vec_id = HNSW_INVALID_ID) override;
```

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -70,7 +70,7 @@ class TieredHNSWIndex : public VecSimTieredIndex<DataType, DistType> {
     TieredHNSWIndex(HNSWIndex<DataType, DistType> *hnsw_index, TieredIndexParams tieredParams);
     virtual ~TieredHNSWIndex();
 
-    int addVector(const void *blob, labelType label, bool overwrite_allowed) override;
+    int addVector(const void *blob, labelType label, idType new_vec_id = -1) override;
```
meiravgri, line 73:
```suggestion
    int addVector(const void *blob, labelType label, idType new_vec_id = HNSW_INVALID_ID) override;
```

src/VecSim/vec_sim_interface.h
```diff
@@ -38,12 +38,15 @@ struct VecSimIndexInterface : public VecsimBaseObject {
...
-    virtual int addVector(const void *blob, labelType label, bool overwriteAllowed) = 0;
+     * @param new_vec_id if this is not the main index (but a layer in a tiered index for example)
+     * we use a predefined internal id for this vector. Otherwise, if new_vec_id is -1 ,use a new
+     * fresh id.
+     * In addition, if id is not given, and this label already exists overwrite it. Otherwise,
+     * it's the caller main index responsibility to validate that the new label and id are
+     * appropriate.
+     * @return the number of new vectors inserted (1 for new insertion, 0 for override).
+     */
+    virtual int addVector(const void *blob, labelType label, idType new_vec_id = -1) = 0;
```
meiravgri, line 49:
```suggestion
    virtual int addVector(const void *blob, labelType label, idType new_vec_id = HNSW_INVALID_ID) = 0;
```

alonre24: This is the generic interface, so I prefer not using here HNSW constant...

meiravgri (approved):

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -214,48 +208,228 @@ TYPED_TEST(HNSWTieredIndexTest, manageIndexOwnership) {
...
+    // Launch the BG threads loop that takes jobs from the queue and executes them.
+    bool run_thread = true;
+    for (size_t i = 0; i < THREAD_POOL_SIZE; i++) {
+        thread_pool.emplace_back(thread_main_loop, std::ref(jobQ), std::ref(run_thread));
+    }
+
+    // Insert vectors
+    for (size_t i = 0; i < n / per_label; i++) {
+        for (size_t j = 0; j < per_label; j++) {
+            GenerateAndAddVector<TEST_DATA_T>(tiered_index, dim, i, i * per_label + j);
```
DvirDukhan, line 400: I think you can cache those if you want to compare them after

src/VecSim/algorithms/brute_force/brute_force.h
```diff
@@ -68,6 +69,7 @@ class BruteForceIndex : public VecSimIndexAbstract<DistType> {
      */
     virtual void getDataByLabel(labelType label,
                                 std::vector<std::vector<DataType>> &vectors_output) const = 0;
+    virtual int deleteVectorById(labelType label, idType id) = 0;
```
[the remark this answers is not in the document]
alonre24: Yes, tiered index is using it for deleting the specific id of the vectors whose insert job has just executed (it's required for MULTI where the label is not enough to determine which vector to delete) - so it shouldn't be under `BUILD_TEST`

src/VecSim/algorithms/brute_force/brute_force.h
```diff
@@ -56,6 +56,7 @@ class BruteForceIndex : public VecSimIndexAbstract<DistType> {
     inline labelType getVectorLabel(idType id) const { return idToLabelMapping.at(id); }
 
     inline vecsim_stl::vector<VectorBlock *> getVectorBlocks() const { return vectorBlocks; }
+    inline const vecsim_stl::vector<labelType> &getIdToLabelMap() const { return idToLabelMapping; }
```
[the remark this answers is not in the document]
alonre24: Right, better having it like this

src/VecSim/algorithms/brute_force/brute_force_multi.h
```diff
@@ -108,6 +109,31 @@ int BruteForceIndex_Multi<DataType, DistType>::deleteVector(labelType label) {
     return ret;
 }
 
+template <typename DataType, typename DistType>
+int BruteForceIndex_Multi<DataType, DistType>::deleteVectorById(labelType label, idType id) {
```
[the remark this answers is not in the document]
alonre24: yes

tests/unit/test_hnsw_tiered.cpp
alonre24: I guess... will do

DvirDukhan (approved):

PR #331, opened 2023-02-15
Set an external index context for tiered index - for ownership management MOD-4875

alonre24:

Let the tiered index receive an external opaque index context, and send this context in turn to the submit callback upon submitting a job to be executed asynchronously. The purpose is that the caller would have the option to create a weak reference to the underline index (or the index spec, in RediSearch case) and associate it with the job. When the job is executed, the weak reference is promoted as long as the index has not been deleted. Hence, if the index is deleted while the job is being executed, it is guarded. 

This includes small refactor - moving the pointer to the index to the base `AsyncJob`

### DIFF
--- a/src/VecSim/vec_sim_common.h
+++ b/src/VecSim/vec_sim_common.h
@@ -63,12 +63,14 @@ typedef enum {
     VecSimParamResolverErr_InvalidPolicy_AdHoc_With_EfRuntime
 } VecSimResolveCode;
 
+typedef struct AsyncJob AsyncJob; // forward declaration.
+
 /**
  * Callback signatures for asynchronous tiered index.
  */
-typedef int (*SubmitCB)(void *job_queue, void **jobs, size_t jobs_len);
+typedef int (*SubmitCB)(void *job_queue, AsyncJob **jobs, size_t jobs_len, void *index_ctx);
 typedef int (*UpdateMemoryCB)(void *memory_ctx, size_t memory);
-typedef void (*JobCallback)(void *);
+typedef void (*JobCallback)(AsyncJob *);
 
 /**
  * @brief Index initialization parameters.
@@ -99,6 +101,7 @@ typedef struct {
 // A struct that contains the common tiered index params.
 typedef struct {
     void *jobQueue;             // External queue that holds the jobs.
+    void *indexCtx;             // External context to be sent to the submit callback.
     SubmitCB submitCb;          // A callback that submits an array of jobs into a given jobQueue.
     void *memoryCtx;            // External context that stores the index memory consumption.
     UpdateMemoryCB UpdateMemCb; // A callback that updates the memoryCtx
--- a/src/VecSim/vec_sim_common.h
+++ b/src/VecSim/vec_sim_common.h
@@ -101,7 +101,7 @@ typedef struct {
 // A struct that contains the common tiered index params.
 typedef struct {
     void *jobQueue;             // External queue that holds the jobs.
-    void *indexCtx;             // External context to be sent to the submit callback.
+    void *jobQueueCtx;          // External context to be sent to the submit callback.
     SubmitCB submitCb;          // A callback that submits an array of jobs into a given jobQueue.
     void *memoryCtx;            // External context that stores the index memory consumption.
     UpdateMemoryCB UpdateMemCb; // A callback that updates the memoryCtx
--- a/src/VecSim/vec_sim_tiered_index.h
+++ b/src/VecSim/vec_sim_tiered_index.h
@@ -11,9 +11,11 @@
 struct AsyncJob : public VecsimBaseObject {
     JobType jobType;
     JobCallback Execute; // A callback that receives a job as its input and executes the job.
+    VecSimIndex *index;
 
-    AsyncJob(std::shared_ptr<VecSimAllocator> allocator, JobType type, JobCallback callback)
-        : VecsimBaseObject(allocator), jobType(type), Execute(callback) {}
+    AsyncJob(std::shared_ptr<VecSimAllocator> allocator, JobType type, JobCallback callback,
+             VecSimIndex *index_ref)
+        : VecsimBaseObject(allocator), jobType(type), Execute(callback), index(index_ref) {}
 };
 
 template <typename DataType, typename DistType>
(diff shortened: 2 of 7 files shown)


tests/unit/test_hnsw_tiered.cpp
```diff
@@ -165,6 +174,81 @@ TYPED_TEST(HNSWTieredIndexTest, addVector) {
...
+        tiered_index->submitSingleJob((AsyncJob *)&job);
+        tiered_index->submitSingleJob((AsyncJob *)&job);
+        ASSERT_EQ(jobQ->size(), 2);
+
+        // Execute the job from the queue asynchronously, delete the index in the meantime.
+        auto run_fn = [&jobQ]() {
+            if (auto temp_ref = jobQ->front().relatedIndex.lock()) {
+                EXPECT_EQ(jobQ->front().relatedIndex.use_count(), 2);
+                reinterpret_cast<AsyncJob *>(jobQ->front().job)->Execute(jobQ->front().job);
+            }
```
GuyAv46, line 232: consider having this definition as past of the mock

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -165,6 +174,81 @@ TYPED_TEST(HNSWTieredIndexTest, addVector) {
...
+        };
+        std::thread t1(run_fn);
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+        delete index_ctx;
+        EXPECT_EQ(jobQ->front().relatedIndex.use_count(), 1);
+        t1.join();
+        // Expect that the first job will succeed.
+        ASSERT_GE(memory_ctx, initial_mem);
+        size_t cur_mem = memory_ctx;
+        jobQ->pop();
```
GuyAv46, line 242: why do you have to pop the last job from the queue? shouldn't it get popped by the thread (and not being executed)?

tests/unit/test_hnsw_tiered.cpp
alonre24: I could, but line 30 is specific to this test. Also, this is not precisely the general main loop for the background thread, it is lighter to suit this test (the more general is this one - https://github.com/RedisAI/VectorSimilarity/blob/tiered_index_insert_job/tests/unit/test_hnsw_tiered.cpp#L321)

tests/unit/test_hnsw_tiered.cpp
alonre24: Basically yes, but again the thread function here is lighter than what is should be in the general case. But I can move the pop for better practice

tests/unit/test_hnsw_tiered.cpp
GuyAv46: OK. Maybe we should have both options as part of the mock for easier test writing in the future, or at least the more general one and specifics can be written for tests that need it?

alonre24: Sure... I'll have it in the other PR where I use the general thread routine

GuyAv46 (approved):

src/VecSim/vec_sim_common.h
```diff
@@ -63,12 +63,14 @@ typedef enum {
     VecSimParamResolverErr_InvalidPolicy_AdHoc_With_EfRuntime
 } VecSimResolveCode;
 
+typedef struct AsyncJob AsyncJob; // forward declaration.
+
 /**
  * Callback signatures for asynchronous tiered index.
  */
-typedef int (*SubmitCB)(void *job_queue, void **jobs, size_t jobs_len);
+typedef int (*SubmitCB)(void *job_queue, AsyncJob **jobs, size_t jobs_len, void *index_ctx);
```
DvirDukhan, line 71: I don't think `AsyncJob` is a well known type outside of the library.

src/VecSim/vec_sim_tiered_index.h
```diff
@@ -23,28 +25,28 @@ class VecSimTieredIndex : public VecSimIndexInterface {
     VecSimIndexAbstract<DistType> *index;
 
     void *jobQueue;
+    void *indexCtx; // External context to be sent to the submit callback.
```
DvirDukhan, line 28: jobQueueCtx

tests/unit/test_utils.h
```diff
@@ -129,7 +129,18 @@ inline double GetInfVal(VecSimType type) {
...
+    AsyncJob *job;
+    std::weak_ptr<VecSimIndex> relatedIndex;
+} OwnedJob;
+
+using JobQueue = std::queue<OwnedJob>;
+int submit_callback(void *job_queue, AsyncJob **jobs, size_t len, void *index_ctx);
 int update_mem_callback(void *mem_ctx, size_t mem);
+
+typedef struct IndexExtCtx {
+    std::shared_ptr<VecSimIndex> index_ref;
```
DvirDukhan, line 143:
```suggestion
    std::shared_ptr<VecSimIndex> index_strong_ref;
```

tests/unit/test_utils.h
```diff
@@ -129,7 +129,18 @@ inline double GetInfVal(VecSimType type) {
     }
 
 namespace tiered_index_mock {
-using JobQueue = std::queue<void *>;
-int submit_callback(void *job_queue, void **jobs, size_t len);
+
+typedef struct OwnedJob {
+    AsyncJob *job;
+    std::weak_ptr<VecSimIndex> relatedIndex;
```
DvirDukhan, line 135:
```suggestion
    std::weak_ptr<VecSimIndex> index_weak_ref;
```

tests/unit/test_utils.h
```diff
@@ -129,7 +129,18 @@ inline double GetInfVal(VecSimType type) {
     }
 
 namespace tiered_index_mock {
-using JobQueue = std::queue<void *>;
-int submit_callback(void *job_queue, void **jobs, size_t len);
+
+typedef struct OwnedJob {
```
DvirDukhan, line 133:
```suggestion
typedef struct RefManagedJob {
```

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -165,6 +174,81 @@ TYPED_TEST(HNSWTieredIndexTest, addVector) {
...
+        // Wrap this job with an array and submit the jobs to the queue.
+        tiered_index->submitSingleJob((AsyncJob *)&job);
+        tiered_index->submitSingleJob((AsyncJob *)&job);
+        ASSERT_EQ(jobQ->size(), 2);
+
+        // Execute the job from the queue asynchronously, delete the index in the meantime.
+        auto run_fn = [&jobQ]() {
+            if (auto temp_ref = jobQ->front().relatedIndex.lock()) {
+                EXPECT_EQ(jobQ->front().relatedIndex.use_count(), 2);
+                reinterpret_cast<AsyncJob *>(jobQ->front().job)->Execute(jobQ->front().job);
```
DvirDukhan, line 231: rewrite a bit clearer?

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -165,6 +174,81 @@ TYPED_TEST(HNSWTieredIndexTest, addVector) {
...
+
+        // Wrap this job with an array and submit the jobs to the queue.
+        tiered_index->submitSingleJob((AsyncJob *)&job);
+        tiered_index->submitSingleJob((AsyncJob *)&job);
+        ASSERT_EQ(jobQ->size(), 2);
+
+        // Execute the job from the queue asynchronously, delete the index in the meantime.
+        auto run_fn = [&jobQ]() {
+            if (auto temp_ref = jobQ->front().relatedIndex.lock()) {
+                EXPECT_EQ(jobQ->front().relatedIndex.use_count(), 2);
```
DvirDukhan, line 230: add comment about this check

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -165,6 +174,81 @@ TYPED_TEST(HNSWTieredIndexTest, addVector) {
...
+                reinterpret_cast<AsyncJob *>(jobQ->front().job)->Execute(jobQ->front().job);
+            }
+            jobQ->pop();
+        };
+        std::thread t1(run_fn);
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+        delete index_ctx;
+        EXPECT_EQ(jobQ->front().relatedIndex.use_count(), 1);
```
DvirDukhan, line 238: move those lines before lunching `t2` and explain the flow

src/VecSim/vec_sim_common.h
alonre24: I moved it to `vec_sim_common.h` (as opaque object) so it will be known outside as well

tests/unit/test_hnsw_tiered.cpp
alonre24: It's here on purpose, I want to delete the index strong ref *before* the first job is done, so we can validate that the index is not deleted while the job is running...

tests/unit/test_hnsw_tiered.cpp
alonre24: Not much here to rewrite... added more comments, see if it's better now

DvirDukhan (approved):

PR #334, opened 2023-02-28
Tiered HNSW - knn search - [MOD-4319]

GuyAv46:

This PR includes the implementation of the simple KNN search of tiered HNSW index (single and multi)
it also includes a minor modification to the updatable heap so the results will be sorted by priority (distance) and break even by value (label), so an optimal implementation for merging the top-k results from the HNSW index and FLAT index will be possible (as we need to avoid duplications)

1. implement KNN search logic for tiered HNSW, including edge cases for optimization.
2. minor fix to the updatable heap. no need to be included in previous versions.
3. tests

**Mark if applicable**


tests/unit/test_hnsw_tiered.cpp
```diff
@@ -259,3 +259,72 @@ TYPED_TEST(HNSWTieredIndexTest, manageIndexOwnership) {
...
+    // TEST 5: Search for 10 vectors so all the vectors will be from the flat index.
+    // TEST 6: Search for 10 vectors so all the vectors will be from the main index.
+    // TEST 7: Search for 10 vectors so some of the vectors will be from the main index and some from the flat index.
+
+    // MORE TESTS:
+    // 2. Search for vectors when the flat index is not empty but the main index is empty.
+    // 3. Search for more vectors than the index size.
+    // 4. Search for less vectors than the index size, but more than the flat and main index sizes.
+    // 5. Search for more vectors than the flat index size, but less than the main index size.
+    // 6. Search for more vectors than the main index size, but less than the flat index size.
```
alonre24, line 325: These are all good tests for basic and edge cases.
In addition, it's important to have parallel tests:
1. multiple parallel searches (only)
2. multiple parallel searches with multiple parallel indexing into HNSW (that is, while insert jobs are running by background threads) + ensure that we have a use case with resizing during insertions. No overwrites (currently not supported)
3. multiple parallel searches with multiple parallel background indexing into HNSW and single parallel `addVector` into flat buffer (again, without overwriting).
4. Tets some of the above variations with MULTI

alonre24 (commented):
Very clear and elegant!
Some minor comments and I suggest a small refactor for the parallel tests (as discussed)

src/VecSim/utils/vec_utils.cpp
```diff
@@ -168,3 +176,54 @@ size_t VecSimType_sizeof(VecSimType type) {
     }
     return 0;
 }
+
+static void concat_results(VecSimQueryResult *dst, VecSimQueryResult *src,
+                           const VecSimQueryResult *src_end, size_t limit) {
+    while (limit && src != src_end) {
+        array_append(dst, *src);
```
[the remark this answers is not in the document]
GuyAv46: I don't agree. It is safe because we know we allocated enough memory, and the compiler will optimize the `array_append` to not include a `mov` for each append call

tests/unit/test_utils.h
```diff
@@ -136,6 +140,20 @@ typedef struct RefManagedJob {
     std::weak_ptr<VecSimIndex> index_weak_ref;
 } RefManagedJob;
 
+struct SearchJobMock : public AsyncJob {
+    void *query;
+    size_t k;
+    size_t n;
```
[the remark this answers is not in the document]
GuyAv46: index size WRT the test, for internal calculations (if you want to relay on the number of results in the expected res CB)

tests/unit/test_utils.h
```diff
@@ -136,6 +140,20 @@ typedef struct RefManagedJob {
...
+    size_t dim;
+    std::atomic_int &successful_searches;
```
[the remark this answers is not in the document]
GuyAv46: but that's the only way to use it in the callback. that's why this struct is considered a mock

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -436,3 +436,371 @@ TYPED_TEST(HNSWTieredIndexTest, insertJobAsyncMulti) {
...
+        GenerateAndAddVector<TEST_DATA_T>(flat_index, dim, i, i);
+    }
+    ASSERT_EQ(tiered_index->indexSize(), n);
+    ASSERT_EQ(tiered_index->indexSize(), hnsw_index->indexSize() + flat_index->indexSize());
+
+    // Search for k vectors so all the vectors will be from the flat index.
+    runTopKSearchTest(tiered_index, query_0, k, ver_res_0);
+    // Search for k vectors so all the vectors will be from the main index.
+    runTopKSearchTest(tiered_index, query_n, k, ver_res_n);
+    // Search for k so some of the results will be from the main and some from the flat index.
```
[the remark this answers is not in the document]
GuyAv46: right... damn copilot :)

tests/unit/test_utils.h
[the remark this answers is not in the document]
alonre24: OK, just document it then

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -436,3 +436,458 @@ TYPED_TEST(HNSWTieredIndexTest, insertJobAsyncMulti) {
...
+    static const size_t checks_in_flat = flat_index->indexSize();
+    VecSim_SetTimeoutCallbackFunction([](void *ctx) {
+        static size_t count = checks_in_flat;
+        if (count == 0) {
+            return 1;
+        }
+        count--;
+        return 0;
+    });
+    res = VecSimIndex_TopKQuery(tiered_index, query_0, k, nullptr, BY_SCORE);
```
alonre24, line 621: No actual way of validating that the timeout was caught in HNSW other than checking the coverage report right?

src/VecSim/utils/merge_results.h
```diff
@@ -0,0 +1,79 @@
...
+    VecSimQueryResult *results = array_new<VecSimQueryResult>(limit);
+    VecSimQueryResult_List mergedResults{.results = results, .code = VecSim_QueryResult_OK};
+    VecSimQueryResult *cur_first = first.results;
+    VecSimQueryResult *cur_second = second.results;
+    const auto first_end = cur_first + VecSimQueryResult_Len(first);
+    const auto second_end = cur_second + VecSimQueryResult_Len(second);
+
+    // Will hold the ids of the results we've already added to the merged results.
+    // Will be used only if withSet is true.
+    std::unordered_set<size_t> ids;
```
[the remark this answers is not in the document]
GuyAv46: it might... but the set/number of lit bits will be `k` at most so I can't tell what will be better. we might need to revisit this.
Maybe we should look for a sparse bitmap implementation?

tests/unit/test_hnsw_tiered.cpp
GuyAv46: right. I can't think of a simple way to change that. This test will work as long as we don't change the way we check for a timeout in the `BF` index.
I'll add an assertion that querying the flat index only does not result in a timeout, so we will fail if the assumption I made will be changed

src/VecSim/utils/merge_results.h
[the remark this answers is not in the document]
alonre24: All right, let's keep it like this

alonre24 (approved):
GGGRREEAAT

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -436,3 +436,465 @@ TYPED_TEST(HNSWTieredIndexTest, insertJobAsyncMulti) {
...
+    TEST_DATA_T query_n[dim];
+    GenerateVector<TEST_DATA_T>(query_n, dim, n - 1);
+
+    // Search for vectors when the index is empty.
+    runTopKSearchTest(tiered_index, query_0, k, nullptr);
+
+    // Define the verification functions.
+    auto ver_res_0 = [&](size_t id, double score, size_t index) {
+        ASSERT_EQ(id, index);
+        ASSERT_DOUBLE_EQ(score, dim * id * id);
```
DvirDukhan, line 488: pow here too?

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -436,3 +436,465 @@ TYPED_TEST(HNSWTieredIndexTest, insertJobAsyncMulti) {
...
+
+        std::atomic_int successful_searches(0);
+        auto parallel_knn_search = [](AsyncJob *job) {
+            auto *search_job = reinterpret_cast<SearchJobMock *>(job);
+            size_t k = search_job->k;
+            size_t dim = search_job->dim;
+            auto query = search_job->query;
+
+            auto verify_res = [&](size_t id, double score, size_t res_index) {
+                TEST_DATA_T el = *(TEST_DATA_T *)query;
```
DvirDukhan, line 675: can you rename `el` to not `el`?

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -436,3 +436,465 @@ TYPED_TEST(HNSWTieredIndexTest, insertJobAsyncMulti) {
...
+            // Add a search job. Make sure the query element is between k and n - k.
+            auto query = (TEST_DATA_T *)allocator->allocate(dim * sizeof(TEST_DATA_T));
+            GenerateVector<TEST_DATA_T>(query, dim, (i % (n_labels - (2 * k))) + k);
+            auto search_job =
+                new (allocator) SearchJobMock(allocator, parallel_knn_search, tiered_index, query,
+                                              k, n, dim, successful_searches);
+            tiered_index->submitSingleJob(search_job);
+        }
+
+        EXPECT_EQ(tiered_index->indexSize(), n) << (isMulti ? "multi" : "single");
```
DvirDukhan, line 703: think about better debug print?

tests/unit/test_utils.cpp
```diff
@@ -44,11 +44,26 @@ VecSimQueryParams CreateQueryParams(const HNSWRuntimeParams &RuntimeParams) {
...
+           dynamic_cast<VecSimTieredIndex<double, double> *>(index) != nullptr;
+}
+
+void runTopKSearchTest(VecSimIndex *index, const void *query, size_t k, size_t expected_num_res,
                        std::function<void(size_t, double, size_t)> ResCB, VecSimQueryParams *params,
                        VecSimQueryResult_Order order) {
     VecSimQueryResult_List res = VecSimIndex_TopKQuery(index, query, k, params, order);
-    ASSERT_EQ(VecSimQueryResult_Len(res), k);
+    if (is_async_index(index)) {
+        // Async index may return more or less than the expected number of results,
```
DvirDukhan, line 57: It should never return more results, no?

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -436,3 +436,465 @@ TYPED_TEST(HNSWTieredIndexTest, insertJobAsyncMulti) {
...
+        for (size_t i = 0; i < n; i++) {
+            GenerateAndAddVector<TEST_DATA_T>(tiered_index, dim, i % n_labels, i);
+            auto query = (TEST_DATA_T *)allocator->allocate(dim * sizeof(TEST_DATA_T));
+            GenerateVector<TEST_DATA_T>(query, dim, (TEST_DATA_T)n / 4 + (i % 1000) * M_PI);
+            auto search_job =
+                new (allocator) SearchJobMock(allocator, parallel_knn_search, tiered_index, query,
+                                              k, n, dim, successful_searches);
+            tiered_index->submitSingleJob(search_job);
+        }
+        ASSERT_GE(tiered_index->labelToInsertJobs.size(), 0) << (isMulti ? "multi" : "single");
```
DvirDukhan, line 815: this test means nothing

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -436,3 +436,465 @@ TYPED_TEST(HNSWTieredIndexTest, insertJobAsyncMulti) {
...
+            thread_pool.emplace_back(thread_main_loop, std::ref(jobQ), std::ref(run_thread));
+        }
+        std::atomic_int successful_searches(0);
+
+        auto parallel_knn_search = [](AsyncJob *job) {
+            auto *search_job = reinterpret_cast<SearchJobMock *>(job);
+            size_t k = search_job->k;
+            auto query = search_job->query;
+            // In this test we don't care about the results, just that the search doesn't crash
+            // and returns the correct number of valid results.
```
DvirDukhan, line 797: why?

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -436,3 +436,465 @@ TYPED_TEST(HNSWTieredIndexTest, insertJobAsyncMulti) {
...
+            thread_pool[i].join();
+        }
+        EXPECT_EQ(successful_searches, n) << (isMulti ? "multi" : "single");
+        EXPECT_EQ(tiered_index->index->indexSize(), n) << (isMulti ? "multi" : "single");
+        EXPECT_EQ(tiered_index->index->indexLabelCount(), n_labels)
+            << (isMulti ? "multi" : "single");
+        EXPECT_EQ(tiered_index->flatBuffer->indexSize(), 0) << (isMulti ? "multi" : "single");
+        EXPECT_EQ(tiered_index->labelToInsertJobs.size(), 0) << (isMulti ? "multi" : "single");
+        EXPECT_EQ(jobQ.size(), 0);
+        // Verify that the vectors were inserted to HNSW as expected
```
DvirDukhan, line 837: isn't it a part of the insert test?

src/VecSim/algorithms/hnsw/hnsw.h
```diff
@@ -1829,6 +1829,12 @@ VecSimQueryResult_List HNSWIndex<DataType, DistType>::topKQuery(const void *quer
     idType bottom_layer_ep = searchBottomLayerEP(query_data, timeoutCtx, &rl.code);
     if (VecSim_OK != rl.code) {
         return rl;
+    } else if (bottom_layer_ep == INVALID_ID) {
+        // Although we checked that the index is not empty (cur_element_count == 0), it might be
```
DvirDukhan, line 1833: do we have a test for this scenario?

tests/unit/test_hnsw_tiered.cpp
GuyAv46: I like the simple multiplication, I think it is more readable

tests/unit/test_hnsw_tiered.cpp
GuyAv46: I made one in the next PR, I can move it to this one

tests/unit/test_utils.cpp
GuyAv46: if the expectation is based on the index size (which is the default expectation), we might have an out-of-date expectation (index size was 9, we looked for 10, but before/while we searched another vector was added so we found 10 results)

tests/unit/test_hnsw_tiered.cpp
GuyAv46: because it's too hard to predict the results as we don't know the order of execution and what vectors are going to be in the index at the execution time. We have other tests for validating the score or the expected ids, so I think it is better to accept the chaos here and only test the simple stuff (number of results, uniqueness) and not limit the case for better prediction

src/VecSim/algorithms/hnsw/hnsw.h
GuyAv46: the test of inserting and searching at the same time catches this case

src/VecSim/utils/merge_results.h
```diff
@@ -0,0 +1,82 @@
...
+static int cmpVecSimQueryResultByScoreThenId(const VecSimQueryResult *res1,
+                                             const VecSimQueryResult *res2) {
+    return (VecSimQueryResult_GetScore(res1) != VecSimQueryResult_GetScore(res2))
+               ? cmpVecSimQueryResultByScore(res1, res2)
+               : cmpVecSimQueryResultById(res1, res2);
+}
+
+// Assumes that the arrays are sorted by score firstly and by id secondarily.
+template <bool withSet>
+VecSimQueryResult_List merge_results(VecSimQueryResult_List first, VecSimQueryResult_List second,
```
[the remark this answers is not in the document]
GuyAv46: Agrees

DvirDukhan (approved):

PR #335, opened 2023-03-02
Tiered index delete vector internal MOD-4320

alonre24:

Implement the internal routine for triggering a deletion vector(s) that correspond to a label from HNSW. That includes collecting every incoming edge for the deleted element(s) in every level and creating appropriate repair and swap jobs (swap jobs are not implemented yet).

This also includes the following change - entry point node in HNSW cannot be marked as deleted. We validate that we don't choose a deleted element as an entry point upon replacing the entry point, and we replace the entry point upon marking it as deleted.


src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -124,6 +139,61 @@ HNSWIndex<DataType, DistType> *TieredHNSWIndex<DataType, DistType>::getHNSWIndex
...
+                    new (this->allocator) HNSWRepairJob(this->allocator, node.first, node.second,
+                                                        executeRepairJobWrapper, this, swap_job);
+                repair_jobs.emplace_back(repair_job);
+                // Insert the repair job into the repair jobs lookup (for fast update in case that
+                // the node id is changed due to swap job).
+                if (idToRepairJobs.find(node.first) == idToRepairJobs.end()) {
+                    idToRepairJobs.insert(
+                        {node.first, vecsim_stl::vector<HNSWRepairJob *>(this->allocator)});
+                }
+                idToRepairJobs.at(node.first).push_back(repair_job);
```
GuyAv46, line 180: can't we use this map to decide if we need a new job? an id doesn't really need more than one repair at a time

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -227,6 +297,16 @@ TieredHNSWIndex<DataType, DistType>::~TieredHNSWIndex() {
...
+    // Delete all the pending repair jobs.
+    for (auto &jobs : this->idToRepairJobs) {
+        for (auto *job : jobs.second) {
+            delete job;
+        }
+    }
+    // Delete all the pending swap jobs.
+    for (auto *job : this->swapJobs) {
+        delete job;
+    }
```
GuyAv46, line 309: is this thread safe? Is that our responsibility or should a job delete itself if it realizes that the relevant index is deleted? This is what we implemented in the GC in RediSearch for reference

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -436,3 +436,184 @@ TYPED_TEST(HNSWTieredIndexTest, insertJobAsyncMulti) {
...
+
+        size_t num_repair_jobs = jobQ.size();
+        // There should be at least two nodes to repair, the neighbors of next_id in levels 0 and 1
+        ASSERT_GE(num_repair_jobs, 2);
+        while (jobQ.size() > 1) {
+            // First we should have jobs for repairing nodes in level 0.
+            ASSERT_EQ(((HNSWRepairJob *)(jobQ.front().job))->level, 0);
+            jobQ.pop();
+        }
+        // The last job should be repairing the single neighbor in level 1.
```
GuyAv46, line 547: 1 and higher no?

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -436,3 +436,184 @@ TYPED_TEST(HNSWTieredIndexTest, insertJobAsyncMulti) {
     // Cleanup.
     thread_pool.clear();
 }
+
+TYPED_TEST(HNSWTieredIndexTest, deleteFromHNSWBasic) {
+    // Create TieredHNSW index instance with a mock queue.
+    std::shared_ptr<VecSimAllocator> allocator = VecSimAllocator::newVecsimAllocator();
+    size_t dim = 4;
+
+    for (auto is_multi : {false, true}) {
```
GuyAv46, line 445: I made a macro for logging the mode, depending on this flag to be called `isMulti`. Consider at least renaming for easier fix in the future

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -436,3 +436,184 @@ TYPED_TEST(HNSWTieredIndexTest, insertJobAsyncMulti) {
...
+    }
+}
+
+TYPED_TEST(HNSWTieredIndexTest, deleteFromHNSWWithRepairJobExec) {
+    // Create TieredHNSW index instance with a mock queue.
+    std::shared_ptr<VecSimAllocator> allocator = VecSimAllocator::newVecsimAllocator();
+    size_t n = 1000;
+    size_t dim = 4;
+
+    for (auto is_multi : {false, true}) {
```
GuyAv46, line 561: same here

GuyAv46 (commented):
Looks good, few notes
Shouldn't we have parallel tests for deletion?

src/VecSim/algorithms/hnsw/hnsw_single.h
```diff
@@ -23,6 +23,14 @@ class HNSWIndex_Single : public HNSWIndex<DataType, DistType> {
     inline void replaceIdOfLabel(labelType label, idType new_id, idType old_id) override;
     inline void setVectorId(labelType label, idType id) override { label_lookup_[label] = id; }
     inline void resizeLabelLookup(size_t new_max_elements) override;
+    inline vecsim_stl::vector<idType> getIdsOfLabel(labelType label) const override {
+        vecsim_stl::vector<idType> ids(this->allocator);
+        auto it = label_lookup_.find(label);
+        if (it != label_lookup_.end()) {
+            ids.push_back(it->second); // put the id that was found in the result
```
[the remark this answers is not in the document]
alonre24: This function is redundant anyway, removed (we're getting the ids from `markDelete`)

src/VecSim/algorithms/hnsw/hnsw_tiered.h
alonre24: We can, but there is another reason why it's better (IMO) to have another repair job anyway - and that is for tracking the repair jobs that correspond a swap job (let's discuss it)

src/VecSim/algorithms/hnsw/hnsw.h
```diff
@@ -1154,7 +1172,9 @@ void HNSWIndex<DataType, DistType>::greedySearchLevel(const void *vector_data, s
         for (int i = 0; i < links_count; i++) {
             idType candidate = node_links[i];
             assert(candidate < this->cur_element_count && "candidate error: out of index range");
-            if (isInProcess(candidate)) {
+            // Don't allow choosing a deleted node as an entry point upon searching for neighbors
+            // candidates (that is, we're NOT running a query, but inserting a new vector).
+            if (isInProcess(candidate) || (!running_query && isMarkedDeleted(candidate))) {
```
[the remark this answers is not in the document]
alonre24: We can always miss the best element, even when we don't have any deleted elements no? this is why its a greedy search... (let's discuss it)

src/VecSim/algorithms/hnsw/hnsw_tiered.h
alonre24: It should be RediSearch responsibility to decrease the index ref count upon receiving `ft.drop` or something BEFORE this destructor is called. Then, when the jobs are executed asynchronously, they should first test if the index is alive, and only if so upgrade and execute the managed job callback that accesses this internal job pointer. EIther way, it should be our responsibility to release the pending jobs.

tests/unit/test_hnsw_tiered.cpp
alonre24: In this scenario specifically we expect that the maximal level of both nodes whose max level is higher than 0 is *exactly* 1

alonre24:
> Looks good, few notes Shouldn't we have parallel tests for deletion?

parallel tests with delete are to come in the upcoming PR where we implement the full delete flow

GuyAv46 (approved):

src/VecSim/algorithms/hnsw/hnsw_tiered.h
alonre24: No. I'll add one

src/VecSim/algorithms/hnsw/hnsw_base_tests_friends.h
```diff
@@ -15,3 +15,5 @@ INDEX_TEST_FRIEND_CLASS(IndexAllocatorTest_test_hnsw_reclaim_memory_Test)
 INDEX_TEST_FRIEND_CLASS(HNSWTest_markDelete_Test)
 INDEX_TEST_FRIEND_CLASS(HNSWTest_allMarkedDeletedLevel_Test)
 INDEX_TEST_FRIEND_CLASS(HNSWTestParallel)
+INDEX_TEST_FRIEND_CLASS(HNSWTieredIndexTest_deleteFromHNSWBasic_Test)
```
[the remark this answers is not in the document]
alonre24: Yes, this test is accessing HNSW private fields (not just the tiered index fields)

src/VecSim/algorithms/hnsw/hnsw.h
```diff
@@ -1160,11 +1181,63 @@ void HNSWIndex<DataType, DistType>::greedySearchLevel(const void *vector_data, s
             DistType d = this->dist_func(vector_data, getDataByInternalId(candidate), this->dim);
             if (d < curDist) {
                 curDist = d;
-                curObj = candidate;
+                bestCand = candidate;
                 changed = true;
+                // Run this code only for non-query code - update the best non deleted cand as well.
+                if (!running_query && !isMarkedDeleted(candidate)) {
+                    bestNonDeletedCand = bestCand;
```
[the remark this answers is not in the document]
alonre24: yes

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -436,3 +436,191 @@ TYPED_TEST(HNSWTieredIndexTest, insertJobAsyncMulti) {
...
+        ASSERT_EQ(tiered_index->deleteLabelFromHNSW(--vec_id), 1);
+        ASSERT_EQ(tiered_index->getHNSWIndex()->element_levels_[vec_id], 1);
+        auto *level_one_neighbors =
+            tiered_index->getHNSWIndex()->getNodeNeighborsAtLevel(vec_id, 1);
+        ASSERT_EQ(tiered_index->getHNSWIndex()->getNodeNeighborsCount(level_one_neighbors), 1);
+
+        size_t num_repair_jobs = jobQ.size();
+        // There should be at least two nodes to repair, the neighbors of next_id in levels 0 and 1
+        ASSERT_GE(num_repair_jobs, 2);
+        while (jobQ.size() > 1) {
```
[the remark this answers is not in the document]
alonre24: I do because it's GREATER EQUAL than 2 (I noticed now)...

DvirDukhan (approved):

PR #345, opened 2023-04-04
Fix deletion of vector from index multi BF [bug]

alonre24:

Fix bug that may occur upon deleting a label with multiple vectors in BF index.
Until now we were using a mechanism that changes the internal ids to delete that are affected due to a swap that happens after some deletion. The bug occurred when some id appeared twice in the internal ids to delete list, and we were updating the *first* occurrence of the affected id due to the swap. The solution is changing the *last* occurrence - the one that we are about to go over - and not the first one which may be invalid after we deleted it.


tests/unit/test_bruteforce_multi.cpp
```diff
@@ -710,6 +710,50 @@ TYPED_TEST(BruteForceMultiTest, search_empty_index) {
...
+    GenerateAndAddVector<TEST_DATA_T>(index, dim, second_label);
+    GenerateAndAddVector<TEST_DATA_T>(index, dim, second_label);
+    ASSERT_EQ(VecSimIndex_IndexSize(index), n);
+
+    // Artificially reorder the internal ids to test that we make the right changes
+    // when we have an id that appears twice in the array upon deleting the ids one by one.
+    ASSERT_EQ(index->labelToIdsLookup.at(second_label).size(), n / 2);
+    index->labelToIdsLookup.at(second_label)[0] = 4;
+    index->labelToIdsLookup.at(second_label)[1] = 2;
+    index->labelToIdsLookup.at(second_label)[2] = 5;
```
GuyAv46, line 738: consider adding assertions to "prove" that the ids of the second label are 2, 4, and 5 before setting them to these values

GuyAv46 (dismissed):
small suggestion
We should also consider in the future sorting the ids vector and then deleting them from big to small, so we will minimize unnecessary vector moving and somewhat preserve the relative order of all the swapped vectors

alonre24:
replaced by #346

PR #361, opened 2023-04-24
Tiered index enable sync mode MOD-5106

alonre24:

Add a global `WriteMode` flag that indicates whether the tiered index should write (insert/delete) using async jobs (default behavior) or directly to the backend index. Also, add an API that allows setting this flag by 3rd party (to be used before and after "burst threads" are in action). 
Also, add a flat buffer limit for the tiered index to optionally bound the number of pending vectors in the flat buffer. Once the flat buffer reaches this limit, every new vector should be inserted directly into the HNSW backend index. 

**Mark if applicable**

- [ X ] This PR introduces API changes


meiravgri (commented):
Good!
I think we should re-consider not locking the flat index when checking if we reached the buffer limit

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -529,6 +547,17 @@ size_t TieredHNSWIndex<DataType, DistType>::indexLabelCount() const {
 template <typename DataType, typename DistType>
 int TieredHNSWIndex<DataType, DistType>::addVector(const void *blob, labelType label,
                                                    void *auxiliaryCtx) {
+
+    if (this->getWriteMode() == VecSim_WriteInPlace ||
+        this->frontendIndex->indexSize() >= this->flatBufferLimit) {
+        auto hnsw_index = this->getHNSWIndex();
```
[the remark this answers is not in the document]
alonre24: As discussed - if we reached to a point where the buffer is full, it's OK to write in place (even if a few vectors will be removed from the buffer in the meantime...) I believe we should avoid locking here...

src/VecSim/vec_sim_interface.h
```diff
@@ -185,4 +185,15 @@ struct VecSimIndexInterface : public VecsimBaseObject {
...
+    /**
+     * @brief Allow 3rd party to set the write mode for tiered index - async insert/delete using
+     * background jobs, or insert/delete inplace.
+     *
+     * @param mode VecSimWriteMode to mode to write in.
+     */
+    static VecSimWriteMode asyncWriteMode;
+    inline static void setWriteMode(VecSimWriteMode mode) {
+        VecSimIndexInterface::asyncWriteMode = mode;
+    }
```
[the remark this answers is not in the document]
alonre24: No... That is the current decision. We said that if we see a request for having it index specific we might allow it, but currently, it doesn't seems realistic that a user will have both ways in the same cluster...

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -598,18 +598,30 @@ size_t TieredHNSWIndex<DataType, DistType>::indexLabelCount() const {
...
                                                    void *auxiliaryCtx) {
-
-    if (this->getWriteMode() == VecSim_WriteInPlace ||
-        this->frontendIndex->indexSize() >= this->flatBufferLimit) {
-        auto hnsw_index = this->getHNSWIndex();
-        // Insert vector directly to HNSW (since flat buffer guard was not held, no need to release
-        // it internally).
-        this->insertVectorToHNSW<false>(hnsw_index, label, blob);
+    int ret = 1;
+    auto hnsw_index = this->getHNSWIndex();
```
meiravgri, line 602: please add comments explaining the different scenarios

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -598,18 +598,30 @@ size_t TieredHNSWIndex<DataType, DistType>::indexLabelCount() const {
...
         this->UpdateIndexMemory(this->memoryCtx, this->getAllocationSize());
-        return 1;
+        return ret;
+    }
+    if (this->frontendIndex->indexSize() >= this->flatBufferLimit) {
+        // Handle overwrite situation.
+        if (!this->backendIndex->isMultiValue()) {
+            ret -= this->deleteVector(label);
+        }
+        if (this->frontendIndex->indexSize() >= this->flatBufferLimit) {
```
meiravgri, line 616: this second check is redundant

alonre24: It doesn't, since `this->deleteVector(label)` may decrease the flat buffer size...

meiravgri (approved):

src/VecSim/algorithms/hnsw/hnsw_tiered.h
alonre24: Right...

src/VecSim/vec_sim_tiered_index.h
```diff
@@ -60,6 +63,9 @@ class VecSimTieredIndex : public VecSimIndexInterface {
     VecSimQueryResult_List topKQuery(const void *queryBlob, size_t k,
                                      VecSimQueryParams *queryParams) override;
 
+    // Return the current state of the global write mode (async/in-place).
+    static VecSimWriteMode getWriteMode() { return VecSimIndexInterface::asyncWriteMode; }
```
DvirDukhan, line 67: why do we need this function in this class?

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -293,6 +300,7 @@ void TieredHNSWIndex<DataType, DistType>::executeReadySwapJobs() {
 template <typename DataType, typename DistType>
 int TieredHNSWIndex<DataType, DistType>::deleteLabelFromHNSW(labelType label) {
     auto *hnsw_index = getHNSWIndex();
+    this->mainIndexGuard.lock_shared();
```
DvirDukhan, line 303: why the addition?

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -579,6 +599,34 @@ template <typename DataType, typename DistType>
...
+    if (this->getWriteMode() == VecSim_WriteInPlace) {
+        this->mainIndexGuard.lock();
+        // Internally, we may overwrite (delete the previous vector stored under this label), and
+        // may need to increase the capacity when we append the new vector afterwards.
+        ret = hnsw_index->addVector(blob, label);
+        this->mainIndexGuard.unlock();
+        this->UpdateIndexMemory(this->memoryCtx, this->getAllocationSize());
+        return ret;
+    }
+    if (this->frontendIndex->indexSize() >= this->flatBufferLimit) {
```
DvirDukhan, line 612: Don't you need to lock here?

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -619,10 +667,9 @@ int TieredHNSWIndex<DataType, DistType>::addVector(const void *blob, labelType l
     // (in case of override in non-MULTI index) - so if it's there, we remove it (and create the
     // required repair jobs), *before* we submit the insert job.
     if (!this->backendIndex->isMultiValue()) {
-        this->mainIndexGuard.lock_shared();
-        this->deleteLabelFromHNSW(label);
-        this->mainIndexGuard.unlock_shared();
-        ret = 0;
+        // If we removed the previous vector from both HNSW and flat in the overwrite process,
+        // we still return 0 (not -1).
+        ret = MAX(ret - this->deleteLabelFromHNSW(label), 0);
```
DvirDukhan, line 672: do we have a test for this scenario?

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -2799,3 +2803,261 @@ TYPED_TEST(HNSWTieredIndexTestBasic, overwriteVectorAsync) {
...
+
+    // Insert another n more vectors INPLACE, while the previous vectors are still being indexed.
+    VecSim_SetWriteMode(VecSim_WriteInPlace);
+    EXPECT_LT(tiered_index->backendIndex->indexSize(), n);
+    for (size_t i = 0; i < n; i++) {
+        TEST_DATA_T vector[dim];
+        for (size_t j = 0; j < dim; j++) {
+            vector[j] = std::rand() / (TEST_DATA_T)RAND_MAX;
+        }
+        VecSimIndex_AddVector(tiered_index, vector, i % n_labels + n_labels);
```
DvirDukhan, line 2893: consider caching the result in line 2887 and now `tiered_index->backendIndex->indexSize()` should be this result + n, before calling join

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -2799,3 +2803,261 @@ TYPED_TEST(HNSWTieredIndexTestBasic, overwriteVectorAsync) {
...
+        VecSimIndex_DeleteVector(tiered_index, n_labels + i);
+    }
+    // At this point, repair jobs should be executed in the background.
+    EXPECT_EQ(tiered_index->getHNSWIndex()->getNumMarkedDeleted(), n);
+
+    // Insert INPLACE another n vector (instead of the ones that were deleted).
+    VecSim_SetWriteMode(VecSim_WriteInPlace);
+    // Run twice, at first run we insert non-existing labels, in the second run we overwrite them
+    // (for single-value index only).
+    for (auto overwrite : {0, 1}) {
```
DvirDukhan, line 2914: why this scenario is related to burst threads?

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -2799,3 +2803,261 @@ TYPED_TEST(HNSWTieredIndexTestBasic, overwriteVectorAsync) {
...
+                                                     DEFAULT_PENDING_SWAP_JOBS_THRESHOLD, 0);
+    auto allocator = tiered_index->getAllocator();
+
+    GenerateAndAddVector<TEST_DATA_T>(tiered_index, dim, 0);
+    ASSERT_EQ(tiered_index->backendIndex->indexSize(), 1);
+    ASSERT_EQ(tiered_index->frontendIndex->indexSize(), 0);
+    ASSERT_EQ(tiered_index->labelToInsertJobs.size(), 0);
+
+    // Set the flat limit to 1 and insert another vector - expect it to go to the flat buffer.
+    tiered_index->flatBufferLimit = 1;
```
DvirDukhan, line 2967: don't we need it in the parameters?

src/VecSim/algorithms/hnsw/hnsw.h
```diff
@@ -1834,6 +1834,9 @@ void HNSWIndex<DataType, DistType>::appendVector(const void *vector_data, const
     AddVectorCtx state{};
     if (auxiliaryCtx == nullptr) {
         this->lockIndexDataGuard();
+        if (indexSize() == indexCapacity()) {
+            increaseCapacity();
```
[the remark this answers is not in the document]
alonre24: If we do overwrite and we are in "write in place" mode, we might remove the previous vector which is the last in the last block -> decrease capacity -> insert the new vector. So we need to verify that we increase the index capacity if necessary in this scenario as well

src/VecSim/vec_sim_tiered_index.h
alonre24: Since the write mode is relevant for tiered index(es) only. Would you expect something otherwise?

src/VecSim/algorithms/hnsw/hnsw_tiered.h
alonre24: I moved the lock/unlock into `deleteLabelFromHNSW` (rather than locking before and after in the caller function)

src/VecSim/algorithms/hnsw/hnsw_tiered.h
alonre24: I don't think so... see https://github.com/RedisAI/VectorSimilarity/pull/361#discussion_r1181255582

src/VecSim/algorithms/hnsw/hnsw_tiered.h
alonre24: Yes it's part of the `bufferLimitAsync` test

tests/unit/test_hnsw_tiered.cpp
alonre24: But that's not necessarily what will happen... the vectors that were being indexed in the background when we ran line 2887 are still being indexed while we insert the new vectors in place... We didn't "stop the world" here when we moved to in-place mode...

tests/unit/test_hnsw_tiered.cpp
alonre24: Actually, it doesn't relate directly to the burst threads scenario, I added the second phase (overwrite in place) to test the overwrite functionality. I guess I should rename the test to `swtichWriteModes`

tests/unit/test_hnsw_tiered.cpp
alonre24: We have it. Only in this test, I set the field directly.

DvirDukhan (approved):

PR #368, opened 2023-05-07
Tiered HNSW - `preferAdHocSearch` fix

GuyAv46:

Fixed a bug that might happen if one passes to `preferAdHocSearch` a subset size smaller than the tiered index size, but larger than any of the underlying index sizes.

Should we consider performing the fix inside the `preferAdHocSearch implementations in HNSW and BF instead of throwing a runtime error?

**Mark if applicable**


tests/unit/test_bruteforce.cpp
```diff
@@ -1238,13 +1238,8 @@ TYPED_TEST(BruteForceTest, preferAdHocOptimization) {
...
 
     // Corner cases - subset size is greater than index size.
-    try {
-        VecSimIndex_PreferAdHocSearch(index, 1, 50, true);
-        FAIL() << "Expected std::runtime error";
-    } catch (std::runtime_error const &err) {
-        EXPECT_EQ(err.what(),
-                  std::string("internal error: subset size cannot be larger than index size"));
-    }
+    ASSERT_NO_THROW(VecSimIndex_PreferAdHocSearch(index, 1, 50, true));
```
alonre24, line 1241: Assert that it is equals to the result with subsetSize = indexSzie rather than just assert no throw...

DvirDukhan (approved):

alonre24 (approved):

PR #379, opened 2023-05-21
disable fp64 index unit tests by default

DvirDukhan:

This PR disables FP64 unit tests for indexing algorithms, but keeps distance metric tests.
To trigger the full tests we can call `make unit_test FP_64=1`.
This is done due to long running valgrind tests


meiravgri (commented):
cool
I would also add a label to the GHA

meiravgri (approved):

PR #380, opened 2023-05-21
Add static info API

alonre24:

Split some index meta-data from the unified info struct, and add `VecSimIndex_StaticInfo` API to retrieve these fields alone. The motivation for that is to avoid heavy computation and locking that may occur when calling `VecSimIndex_Info`, since today, in the tiered index, this includes computing the number of labels in the tiered index that includes going over both sub-indices labels-set while holding a lock. So now we can replace the call for `VecSimIndex_Info` upon resolving query params and creating a vector iterator in RediSearch, and remove this bottleneck.

Also,  add additional 2 fields for tiered index - buffer limit and swap jobs threshold.

**Mark if applicable**

- [ X ] This PR introduces API changes

### DIFF
--- a/src/VecSim/algorithms/hnsw/hnsw.h
+++ b/src/VecSim/algorithms/hnsw/hnsw.h
@@ -239,6 +239,7 @@ public:
     inline VisitedNodesHandler *getVisitedList() const;
     inline void returnVisitedList(VisitedNodesHandler *visited_nodes_handler) const;
     VecSimIndexInfo info() const override;
+    VecSimIndexStaticInfo staticInfo() const override;
     VecSimInfoIterator *infoIterator() const override;
     bool preferAdHocSearch(size_t subsetSize, size_t k, bool initial_check) const override;
     char *getDataByInternalId(idType internal_id) const;
@@ -2153,7 +2154,7 @@ VecSimIndexInfo HNSWIndex<DataType, DistType>::info() const {
     VecSimIndexInfo info;
     info.commonInfo = this->getCommonInfo();
 
-    info.algo = VecSimAlgo_HNSWLIB;
+    info.commonInfo.basicInfo.algo = VecSimAlgo_HNSWLIB;
     info.hnswInfo.M = this->getM();
     info.hnswInfo.efConstruction = this->getEfConstruction();
     info.hnswInfo.efRuntime = this->getEf();
@@ -2166,23 +2167,31 @@ VecSimIndexInfo HNSWIndex<DataType, DistType>::info() const {
 }
 
 template <typename DataType, typename DistType>
+VecSimIndexStaticInfo HNSWIndex<DataType, DistType>::staticInfo() const {
+    VecSimIndexStaticInfo info = this->getStaticInfo();
+    info.algo = VecSimAlgo_HNSWLIB;
+    return info;
+}
+
+template <typename DataType, typename DistType>
 VecSimInfoIterator *HNSWIndex<DataType, DistType>::infoIterator() const {
     VecSimIndexInfo info = this->info();
     // For readability. Update this number when needed.
     size_t numberOfInfoFields = 17;
     VecSimInfoIterator *infoIterator = new VecSimInfoIterator(numberOfInfoFields);
 
-    infoIterator->addInfoField(VecSim_InfoField{
-        .fieldName = VecSimCommonStrings::ALGORITHM_STRING,
-        .fieldType = INFOFIELD_STRING,
-        .fieldValue = {FieldValue{.stringValue = VecSimAlgo_ToString(info.algo)}}});
+    infoIterator->addInfoField(
+        VecSim_InfoField{.fieldName = VecSimCommonStrings::ALGORITHM_STRING,
+                         .fieldType = INFOFIELD_STRING,
+                         .fieldValue = {FieldValue{
+                             .stringValue = VecSimAlgo_ToString(info.commonInfo.basicInfo.algo)}}});
 
     this->addCommonInfoToIterator(infoIterator, info.commonInfo);
 
-    infoIterator->addInfoField(
-        VecSim_InfoField{.fieldName = VecSimCommonStrings::BLOCK_SIZE_STRING,
-                         .fieldType = INFOFIELD_UINT64,
-                         .fieldValue = {FieldValue{.uintegerValue = info.commonInfo.blockSize}}});
+    infoIterator->addInfoField(VecSim_InfoField{
+        .fieldName = VecSimCommonStrings::BLOCK_SIZE_STRING,
+        .fieldType = INFOFIELD_UINT64,
+        .fieldValue = {FieldValue{.uintegerValue = info.commonInfo.basicInfo.blockSize}}});
 
     infoIterator->addInfoField(
         VecSim_InfoField{.fieldName = VecSimCommonStrings::HNSW_M_STRING,
--- a/src/VecSim/algorithms/hnsw/hnsw.h
+++ b/src/VecSim/algorithms/hnsw/hnsw.h
@@ -239,7 +239,7 @@ public:
     inline VisitedNodesHandler *getVisitedList() const;
     inline void returnVisitedList(VisitedNodesHandler *visited_nodes_handler) const;
     VecSimIndexInfo info() const override;
-    VecSimIndexStaticInfo staticInfo() const override;
+    VecSimIndexBasicInfo basicInfo() const override;
     VecSimInfoIterator *infoIterator() const override;
     bool preferAdHocSearch(size_t subsetSize, size_t k, bool initial_check) const override;
     char *getDataByInternalId(idType internal_id) const;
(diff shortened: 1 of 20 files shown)


src/VecSim/vec_sim_common.h
```diff
@@ -193,16 +193,26 @@ typedef struct {
...
-    size_t dim;              // Vector size (dimension).
-    VecSearchMode last_mode; // The mode in which the last query ran.
+    VecSimAlgo algo;     // Algorithm being used.
+    size_t blockSize;    // Brute force algorithm vector block (mini matrix) size
+    VecSimMetric metric; // Index distance metric
+    VecSimType type;     // Datatype the index holds.
+    bool isMulti;        // Determines if the index should multi-index or not.
+    size_t dim;          // Vector size (dimension).
+
+    VecSimAlgo tieredBackendAlgo; // The algorithm for the tiered index (if algo is tiered).
```
GuyAv46, line 207: I think it would be clearer and more meaningful to have `algo` as the primary algorithm, and to have `isTiered` as a bool

src/VecSim/vec_sim.cpp
```diff
@@ -235,6 +235,8 @@ extern "C" VecSimInfoIterator *VecSimIndex_InfoIterator(VecSimIndex *index) {
     return index->infoIterator();
 }
 
+VecSimIndexStaticInfo VecSimIndex_StaticInfo(VecSimIndex *index) { return index->staticInfo(); }
```
DvirDukhan, line 238: extern C?

src/VecSim/vec_sim_common.h
DvirDukhan: why the backend algorithm is relevant here? we have it at `tieredInfoStruct`

src/VecSim/vec_sim_common.h
```diff
@@ -193,16 +193,26 @@ typedef struct {
...
+} VecSimIndexStaticInfo;
```
DvirDukhan, line 208:
```suggestion
} VecSimIndexBasicInfo;
```

src/VecSim/vec_sim_interface.h
```diff
@@ -160,10 +160,19 @@ struct VecSimIndexInterface : public VecsimBaseObject {
...
+     * locking and going over the labels sets). So this should be used carefully.
      */
     virtual VecSimIndexInfo info() const = 0;
 
+    /**
+     * @brief Return index static information.
+     *
+     * @return Index general and specific meta-data (for quick and lock-less data retrieval)_
+     */
+    virtual VecSimIndexStaticInfo staticInfo() const = 0;
```
DvirDukhan, line 174:
```suggestion
    virtual VecSimIndexStaticInfo basicInfo() const = 0;
```

src/VecSim/vec_sim.h
```diff
@@ -148,6 +148,13 @@ VecSimQueryResult_List VecSimIndex_RangeQuery(VecSimIndex *index, const void *qu
  */
 VecSimIndexInfo VecSimIndex_Info(VecSimIndex *index);
 
+/**
+ * @brief Return basic immutable index information.
+ * @param index the index to return its info.
+ * @return Index basic meta-data.
+ */
+VecSimIndexStaticInfo VecSimIndex_StaticInfo(VecSimIndex *index);
```
DvirDukhan, line 156:
```suggestion
VecSimIndexStaticInfo VecSimIndex_BasicInfo(VecSimIndex *index);
```

src/VecSim/vec_sim_tiered_index.h
```diff
@@ -88,7 +88,6 @@ class VecSimTieredIndex : public VecSimIndexInterface {
     }
 
     virtual VecSimIndexInfo info() const override;
-    virtual VecSimInfoIterator *infoIterator() const override;
```
DvirDukhan, line 91: why?

src/VecSim/algorithms/hnsw/hnsw.h
```diff
@@ -2165,24 +2166,32 @@ VecSimIndexInfo HNSWIndex<DataType, DistType>::info() const {
     return info;
 }
 
+template <typename DataType, typename DistType>
+VecSimIndexStaticInfo HNSWIndex<DataType, DistType>::staticInfo() const {
+    VecSimIndexStaticInfo info = this->getStaticInfo();
```
DvirDukhan, line 2171: shouldn't be a part of the abstract class?

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -1001,3 +1003,73 @@ void TieredHNSWIndex<DataType, DistType>::TieredHNSW_BatchIterator::filter_irrel
     // Update number of results (pop the tail)
     array_pop_back_n(rl.results, end - cur_end);
 }
+
+template <typename DataType, typename DistType>
+VecSimIndexInfo TieredHNSWIndex<DataType, DistType>::info() const {
+    auto info = VecSimTieredIndex<DataType, DistType>::info();
```
DvirDukhan, line 1009: Can you explain why you removed the general purpose tiered info implementation?
I think you could have called it and just inject the two lines
```
    HnswTieredInfo hnswTieredInfo = {.pendingSwapJobsThreshold = this->pendingSwapJobsThreshold};
    info.tieredInfo.specificTieredBackendInfo.hnswTieredInfo = hnswTieredInfo;
```

src/VecSim/vec_sim_common.h
alonre24: I needed it in the concise `VecSimIndexStaticInfo` struct because we use it in `VecSimIndex_ResolveParams` (which is called in every query).

src/VecSim/vec_sim_tiered_index.h
alonre24: Because we have implementation-specific field ( swap jobs threshold)

src/VecSim/algorithms/hnsw/hnsw.h
alonre24: I pit it here since we need to set the backend index algorithm (which is known only here and not in the base class)

src/VecSim/algorithms/hnsw/hnsw_tiered.h
alonre24: I thought it was better and cleaner like this (that the derived class sets its own info), but not critical, I can move the backend data info back to the base class.

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -1001,3 +1003,73 @@ void TieredHNSWIndex<DataType, DistType>::TieredHNSW_BatchIterator::filter_irrel
...
+    info.tieredInfo.backendAlgo = VecSimAlgo_HNSWLIB;
+    info.tieredInfo.backendInfo.hnswInfo = this->backendIndex->info().hnswInfo;
+    HnswTieredInfo hnswTieredInfo = {.pendingSwapJobsThreshold = this->pendingSwapJobsThreshold};
+    info.tieredInfo.specificTieredBackendInfo.hnswTieredInfo = hnswTieredInfo;
+
+    return info;
+}
+
+template <typename DataType, typename DistType>
+VecSimInfoIterator *TieredHNSWIndex<DataType, DistType>::infoIterator() const {
```
GuyAv46, line 1019: Why did these two functions move? it is almost entirely common to any tiered algorithm, Consider overriding and calling the ancestor's function for the extra stuff.

src/VecSim/vec_sim_common.h
alonre24: @GuyAv46  I like this idea

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -1001,3 +1003,73 @@ void TieredHNSWIndex<DataType, DistType>::TieredHNSW_BatchIterator::filter_irrel
...
+        .fieldValue = {FieldValue{.uintegerValue = info.tieredInfo.specificTieredBackendInfo
+                                                       .hnswTieredInfo.pendingSwapJobsThreshold}}});
+
+    return infoIterator;
+}
+
+template <typename DataType, typename DistType>
+VecSimIndexStaticInfo TieredHNSWIndex<DataType, DistType>::staticInfo() const {
+    VecSimIndexStaticInfo info = this->backendIndex->getStaticInfo();
+    info.blockSize = INVALID_INFO;
```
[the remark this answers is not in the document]
alonre24: So you are saying that you agree with what you see here?

src/VecSim/algorithms/hnsw/hnsw_tiered.h
alonre24: For `info` that's basically what I did, I'll do it for `infoIterator` as well

src/VecSim/algorithms/hnsw/hnsw_tiered.h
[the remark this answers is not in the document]
GuyAv46: No. I meant that maybe we should report the block size forward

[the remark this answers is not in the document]
alonre24: It's fine by me, but why didn't we do it until now?

DvirDukhan (approved):

PR #387, opened 2023-05-29
Additional benchmarks - memory and search

alonre24:

Add micro-benchmarks for parallel KNN with tiered index, and a "dummy benchmark" for having the tiered index memory size. Also, this includes unifying the bindings mock thread pool with the global thread pool util, and some minor changes in flow tests.

**Mark if applicable**


tests/utils/mock_thread_pool.cpp
```diff
@@ -79,11 +79,11 @@ void tieredIndexMock::thread_pool_join() {
     thread_pool.clear();
 }
 
-void tieredIndexMock::thread_pool_wait() {
+void tieredIndexMock::thread_pool_wait(size_t waiting_duration) {
     while (true) {
         if (jobQ.empty() && executions_status.count() == 0) {
```
meiravgri, line 84: once the queue is empty we need to check only the execution status
also, no need to lock the jobq for this check?

tests/utils/mock_thread_pool.h
```diff
@@ -42,13 +42,29 @@ class tieredIndexMock {
...
-              dim(dim_), successful_searches(successful_searches_) {}
+              dim(dim_), successful_searches(successful_searches_), ef(-1), iter(-1),
+              all_results(nullptr) {}
+        SearchJobMock(std::shared_ptr<VecSimAllocator> allocator, JobCallback searchCB,
+                      VecSimIndex *index_, size_t k_, size_t ef_, size_t iter_,
+                      VecSimQueryResult_List *all_results_)
+            : AsyncJob(allocator, HNSW_SEARCH_JOB, searchCB, index_), query(nullptr), k(k_), n(-1),
+              dim(-1), successful_searches(nullptr), ef(ef_), iter(iter_),
+              all_results(all_results_) {}
+
```
meiravgri, line 67: a comment about where do we use each? 
also, setting dim and n to -1 seems kinda risky to me

tests/utils/mock_thread_pool.h
```diff
@@ -42,13 +42,29 @@ class tieredIndexMock {
         size_t k;    // The number of results to return.
         size_t n;    // The number of vectors in the index (might be useful for the mock)
         size_t dim;  // The dimension of the vectors in the index (might be useful for the mock)
```
meiravgri, line 44: can't these be conculded from the index?

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -729,7 +729,7 @@ TYPED_TEST(HNSWTieredIndexTest, parallelSearch) {
             ASSERT_EQ(score, dim * (id - element) * (id - element));
         };
         runTopKSearchTest(job->index, query, k, verify_res);
-        search_job->successful_searches++;
+        (*search_job->successful_searches)++;
```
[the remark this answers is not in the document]
alonre24: I don't think that the dereference should be atomic... The increment is atomic for sure

tests/benchmark/bm_common.h
```diff
@@ -113,3 +177,15 @@ void BM_VecSimCommon<index_type_t>::TopK_HNSW(benchmark::State &st, unsigned sho
...
+// {ef_runtime, k} (recall that always ef_runtime >= k)
+#define REGISTER_TopK_Tiered(BM_CLASS, BM_FUNC)                                                    \
+    BENCHMARK_REGISTER_F(BM_CLASS, BM_FUNC)                                                        \
+        ->Args({10, 10})                                                                           \
+        ->Args({200, 10})                                                                          \
+        ->Args({100, 100})                                                                         \
+        ->Args({200, 100})                                                                         \
+        ->Args({500, 500})                                                                         \
+        ->ArgNames({"ef_runtime", "k"})                                                            \
+        ->Iterations(50)                                                                           \
```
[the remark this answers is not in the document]
alonre24: Yes, because I want to have more than 10 runs so the parallelism will have more effect here (with 10 it's only 1-2 iterations per thread).

src/python_bindings/bindings.cpp
```diff
@@ -368,87 +367,71 @@ class PyTieredIndex : public PyVecSimIndex {
...
-    explicit PyTieredIndex(size_t BufferLimit = 3000000)
-        : submitCb(submit_callback), flatBufferLimit(BufferLimit), run_thread(true) {
-
-        for (size_t i = 0; i < THREAD_POOL_SIZE; i++) {
-            ThreadParams params(run_thread, executions_status, i, jobQueue);
-            thread_pool.emplace_back(thread_main_loop, params);
+    explicit PyTieredIndex() {
+        if (tieredIndexMock::ctx != nullptr || !tieredIndexMock::jobQ.empty() ||
+            tieredIndexMock::thread_pool.size() > 0) {
+            throw std::runtime_error("Previous tests did not ended successfully - aborting test");
```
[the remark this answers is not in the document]
alonre24: Yes, if the test doesn't pass the distractor will not be invoked (I checked). In that case, I don't think that we should proceed because we cannot always clean up properly and continue...

tests/utils/mock_thread_pool.cpp
alonre24: Right I guess we should

tests/utils/mock_thread_pool.h
alonre24: Added comments.
Why risky? it's for having an indication that this is an illegal usage and crash/fails ASAP (it's better than having garbage there in my opinion). Also, recall that this is a non-production code...

tests/utils/mock_thread_pool.h
alonre24: I guess... I believe that @GuyAv46 did it like this for convenience

src/python_bindings/bindings.cpp
```diff
@@ -409,13 +396,14 @@ class PyTiered_HNSWIndex : public PyTieredIndex {
...
         // Create TieredIndexParams. Assume that tieredIndexMock::ctx was allocated in the base
         // class constructor, and that the thread pool was created.
         TieredIndexParams tiered_params = {
-            .jobQueue = &tieredIndexMock::jobQ,
-            .jobQueueCtx = tieredIndexMock::ctx,
+            .jobQueue = &(this->mock_thread_pool.jobQ),
+            .jobQueueCtx = this->mock_thread_pool.ctx,
             .submitCb = tieredIndexMock::submit_callback,
             .flatBufferLimit = buffer_limit,
         };
```
meiravgri: I suggest moving this to the base class under `getTieredIndexParams() ` or `generateTieredIndexParams() `

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -24,13 +24,12 @@ class HNSWTieredIndexTest : public ::testing::Test {
         return tiered_index->getHNSWIndex();
     }
     TieredHNSWIndex<data_t, dist_t> *CreateTieredHNSWIndex(VecSimParams &hnsw_params,
-                                                           tieredIndexMock::IndexExtCtx *ctx,
+                                                           tieredIndexMock &mock_thread_pool,
```
meiravgri, line 27: I think that there is a way to define a ctor to the test. We can just put the threadpool initialization there

meiravgri (commented):
LGTM !
Some little comments

tests/unit/test_hnsw_tiered.cpp
alonre24: But it's not required for all tests...

meiravgri (approved):

PR #388, opened 2023-06-14
Expose swap jobs API

alonre24:

Expose a new API that allows running the GC in tiered index. In HNSW tiered, this means that we run up to `DEFAULT_PENDING_SWAP_JOBS_THRESHOLD` ready swap jobs at each call.

**Mark if applicable**

- [X ] This PR introduces API changes


src/VecSim/vec_sim_interface.h
```diff
@@ -220,6 +220,11 @@ struct VecSimIndexInterface : public VecsimBaseObject {
      */
     virtual inline void setLastSearchMode(VecSearchMode mode) = 0;
 
+    /**
+     * @brief Run async garbage collection for tiered async index.
+     */
+    virtual void runGC() = 0;
```
DvirDukhan, line 226: Consider implementing the no-op on the abstract class so we will need to have no-op on flat/hnsw etc..

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -275,13 +280,8 @@ HNSWIndex<DataType, DistType> *TieredHNSWIndex<DataType, DistType>::getHNSWIndex
 }
 
 template <typename DataType, typename DistType>
-void TieredHNSWIndex<DataType, DistType>::executeReadySwapJobs() {
+void TieredHNSWIndex<DataType, DistType>::executeReadySwapJobs(size_t maxJobsToRun) {
```
DvirDukhan, line 283: why do we need it here
Also
consider moving the "best effort" implementation to also run up to a certain threshold

DvirDukhan (dismissed):

DvirDukhan (approved):

PR #399, opened 2023-07-03
Optimize - implement align allocation for vector alignment - [MOD-5433]

GuyAv46:

Implement aligned allocation for the allocator. This might be useful for allocating vector blocks when the dimension is some multiplication of the highest hardware optimization available - FLOAT32 with AVX512 - 16 (64B); FLOAT64 with SSE - 2 (8B).

The added cost is the alignment needed * the number of blocks, in bytes.
For example, a FLOAT32 index of 1,048,576 (1024^2) vectors, with default 1024 vectors per block, HNSW algorithm, and AVX512 enabled, we will get 1024*64 = 65536B = 64KB.

In addition, we now copy the query vector (insertion and query) into an aligned buffer if alignment is specified, so every multiplication of vectors will always be of 2 aligned ones. This change has no overhead.

1. data blocks
2. distance function getters (returning alignment hint)
3. VecSimIndexAbstract (new `alignment` member)

**Mark if applicable**


src/VecSim/vec_sim_tiered_index.h
```diff
@@ -99,36 +100,36 @@ class VecSimTieredIndex : public VecSimIndexInterface {
...
-        return this->topKQuery(query_to_send, k, queryParams);
+        char PORTABLE_ALIGN processed_blob[this->backendIndex->getDataSize()];
+        this->backendIndex->processBlob(queryBlob, processed_blob);
+        return this->topKQuery(processed_blob, k, queryParams);
     }
 
     virtual VecSimQueryResult_List rangeQueryWrapper(const void *queryBlob, double radius,
                                                      VecSimQueryParams *queryParams,
                                                      VecSimQueryResult_Order order) const override {
         // Will be used only if a processing stage is needed
```
meiravgri, line 119: remove comment

tests/unit/test_allocator.cpp
```diff
@@ -413,7 +413,7 @@ TYPED_TEST(IndexAllocatorTest, test_hnsw_reclaim_memory) {
     expected_mem_delta +=
         (hnswIndex->labelLookup.bucket_count() - prev_bucket_count) * sizeof(size_t);
     // New blocks allocated
-    expected_mem_delta += 2 * (sizeof(DataBlock) + vecsimAllocationOverhead);
+    expected_mem_delta += 2 * (sizeof(DataBlock) + vecsimAllocationOverhead + hnswIndex->alignment);
```
meiravgri, line 416: what about the offset overhead for each vector?

meiravgri (commented):
Do we stil benefit from this alignment after handling the tail first in distance calculations?

src/VecSim/utils/alignment.h
```diff
@@ -0,0 +1,32 @@
...
+#define PORTABLE_ALIGN64 __attribute__((aligned(64)))
+#elif defined(_MSC_VER)
+#define PORTABLE_ALIGN16 __declspec(align(16))
+#define PORTABLE_ALIGN32 __declspec(align(32))
+#define PORTABLE_ALIGN64 __declspec(align(64))
+#endif
+
+#define PORTABLE_ALIGN PORTABLE_ALIGN64
+
+// TODO: relax the above alignment requirements according to the CPU architecture
```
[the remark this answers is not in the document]
GuyAv46: This is used for stack allocation only so it's not urgent. It's also a compile-time decision so it's not as important as choosing a distance function. can be left as is for now.

src/VecSim/spaces/IP_space.cpp
```diff
@@ -15,9 +15,15 @@
 #include "VecSim/spaces/implementation_chooser.h"
 
 namespace spaces {
-dist_func_t<float> IP_FP32_GetDistFunc(size_t dim, const Arch_Optimization arch_opt) {
+dist_func_t<float> IP_FP32_GetDistFunc(size_t dim, const Arch_Optimization arch_opt,
+                                       unsigned char *alignment) {
+    unsigned char dummy_alignment;
+    if (alignment == nullptr) {
```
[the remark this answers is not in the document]
GuyAv46: RN only in the unit-tests

src/VecSim/spaces/L2_space.cpp
```diff
@@ -31,16 +37,19 @@ dist_func_t<float> L2_FP32_GetDistFunc(size_t dim, const Arch_Optimization arch_
     case ARCH_OPT_AVX512_F:
 #ifdef __AVX512F__
         CHOOSE_IMPLEMENTATION(ret_dist_func, dim, 16, FP32_L2SqrSIMD16Ext_AVX512);
```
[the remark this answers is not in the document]
GuyAv46: The alignment has no effect when used with a non-512bit dimension. I can add this consideration to the decision

src/VecSim/vec_sim_index.h
```diff
@@ -167,19 +169,15 @@ struct VecSimIndexAbstract : public VecSimIndexInterface {
             .fieldType = INFOFIELD_STRING,
             .fieldValue = {FieldValue{.stringValue = VecSimSearchMode_ToString(info.lastMode)}}});
     }
-    const void *processBlob(const void *original_blob, void *processed_blob) const {
+    void processBlob(const void *original_blob, void *processed_blob) const {
+        // copy original blob to the output blob
+        memcpy(processed_blob, original_blob, this->dataSize);
```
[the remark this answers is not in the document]
GuyAv46: Not really. I'll add a check for that

src/VecSim/algorithms/hnsw/hnsw_serializer.h
```diff
@@ -155,7 +155,8 @@ void HNSWIndex<DataType, DistType>::restoreGraph(std::ifstream &input) {
 
     // Get data blocks
     for (size_t i = 0; i < num_blocks; i++) {
-        this->vectorBlocks.emplace_back(this->blockSize, this->dataSize, this->allocator);
+        this->vectorBlocks.emplace_back(this->blockSize, this->dataSize, this->alignment,
```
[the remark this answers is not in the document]
GuyAv46: Not sure I understand the question. The serialization files contain the data only, not the chosen optimization and not the data blocks or the way they are aligned. This is decided when we load the data and not according to the serialization file

tests/unit/test_allocator.cpp
GuyAv46: what overhead? We allocate the blocks in an aligned manner, and then each vector appears right after the previous one.

src/VecSim/vec_sim_index.h
```diff
@@ -85,7 +87,7 @@ struct VecSimIndexAbstract : public VecSimIndexInterface {
           blockSize(params.blockSize ? params.blockSize : DEFAULT_BLOCK_SIZE), lastMode(EMPTY_MODE),
           isMulti(params.multi), logCallbackCtx(params.logCtx) {
         assert(VecSimType_sizeof(vecType));
-        spaces::SetDistFunc(metric, dim, &distFunc);
+        spaces::SetDistFunc(metric, dim, &distFunc, &alignment);
```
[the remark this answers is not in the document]
GuyAv46: The alignment is not a distance attribute so it's a bit strange to call it this way. We can think of another name. I'll make the comment anyway

src/VecSim/vec_sim_index.h
```diff
@@ -167,19 +169,23 @@ struct VecSimIndexAbstract : public VecSimIndexInterface {
...
+        } else {
+            processed_blob = (void *)original_blob;
+        }
+
         // if the metric is cosine, we need to normalize
         if (this->metric == VecSimMetric_Cosine) {
-            // copy original blob to the output blob
-            memcpy(processed_blob, original_blob, this->dataSize);
             // normalize the copy in place
             normalize_func(processed_blob, this->dim);
```
meiravgri, line 185: can be moved inside the first if to avoid checking if cosine twice when it isnt

src/VecSim/vec_sim_tiered_index.h
```diff
@@ -98,37 +99,33 @@ class VecSimTieredIndex : public VecSimIndexInterface {
...
+        char PORTABLE_ALIGN aligned_mem[this->backendIndex->getDataSize()];
+        const void *processed_blob = this->backendIndex->processBlob(queryBlob, aligned_mem);
 
-        return this->rangeQuery(query_to_send, radius, queryParams, order);
+        return this->rangeQuery(processed_blob, radius, queryParams, order);
     }
 
     virtual VecSimBatchIterator *
     newBatchIteratorWrapper(const void *queryBlob, VecSimQueryParams *queryParams) const override {
-        // Will be used only if a processing stage is needed
```
meiravgri, line 127: return the comment? 😅

src/VecSim/spaces/L2_space.cpp
```diff
@@ -16,9 +16,15 @@
...
-dist_func_t<float> L2_FP32_GetDistFunc(size_t dim, const Arch_Optimization arch_opt) {
+dist_func_t<float> L2_FP32_GetDistFunc(size_t dim, const Arch_Optimization arch_opt,
+                                       unsigned char *alignment) {
+    unsigned char dummy_alignment;
+    if (!alignment) {
+        alignment = &dummy_alignment;
+    }
 
     dist_func_t<float> ret_dist_func = FP32_L2Sqr;
+    *alignment = 4; // Default alignment for float
```
meiravgri, line 27: Why do we need a default value if we decide not to align?

meiravgri (commented):
😎

src/VecSim/vec_sim_index.h
GuyAv46: I trust the compiler to do so. This is done once in a query/insertion anyway

src/VecSim/vec_sim_tiered_index.h
GuyAv46: I think that now it's a matter of luck, but the new name suggests better whether it will be used or not, don't you think?

src/VecSim/spaces/L2_space.cpp
GuyAv46: we currently must give some alignment (at least `1`) to use the `DataBlock` constructor without garbage in `alignment`.

src/VecSim/vec_sim_index.h
meiravgri: in gcc we trust

src/VecSim/vec_sim_tiered_index.h
meiravgri: Maybe the comment should be on top of processed_blob

DvirDukhan (approved):

PR #408, opened 2023-07-24
Logs extension

alonre24:

Add log level to our log messages and callback. Use "verbose" for the existing logs and add some more messages regarding capacity changes. + tiny fix in a test to reduce flakiness.

### DIFF
--- a/src/VecSim/vec_sim_tiered_index.h
+++ b/src/VecSim/vec_sim_tiered_index.h
@@ -53,6 +53,10 @@ protected:
         this->SubmitJobsToQueue(this->jobQueue, this->jobQueueCtx, jobs.data(), callbacks.data(),
                                 jobs.size());
     }
+    template <typename... Args>
+    void log(const char *level, const char *msg, Args... args) const {
+        this->backendIndex->log(level, msg, args...);
+    }
 
 public:
     VecSimTieredIndex(VecSimIndexAbstract<DistType> *backendIndex_,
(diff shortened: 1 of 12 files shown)


src/VecSim/vec_sim_tiered_index.h
```diff
@@ -53,6 +53,10 @@ class VecSimTieredIndex : public VecSimIndexInterface {
         this->SubmitJobsToQueue(this->jobQueue, this->jobQueueCtx, jobs.data(), callbacks.data(),
                                 jobs.size());
     }
+    template <typename... Args>
+    void log(const char *level, const char *msg, Args... args) const {
+        this->backendIndex->log(level, msg, args...);
+    }
```
meiravgri, line 59: So for each call to log, we will have an additional function symbol in the binary file 
for example: ("%s, %s", "meow", "charlie") and ("%s",meow) yields 2 symbols
I suggest that the tiered index will hold a copy to the backend index log function

tests/unit/test_common.cpp
```diff
@@ -464,12 +464,12 @@ TEST(CommonAPITest, testlog) {
         dynamic_cast<BruteForceIndex<float, float> *>(BruteForceFactory::NewIndex(&params));
```
meiravgri, line 464: consider to add a a test also for tierred index

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -2889,7 +2889,7 @@ TYPED_TEST(HNSWTieredIndexTest, switchWriteModes) {
                          .metric = VecSimMetric_L2,
                          .multi = TypeParam::isMulti(),
                          .M = 32,
-                         .efRuntime = 2 * n};
+                         .efRuntime = 3 * n};
```
meiravgri, line 2892: why the change?

meiravgri (commented):
See comment below :)

tests/unit/test_hnsw_tiered.cpp
alonre24: see https://redis.slack.com/archives/C03QDNA8K0B/p1690090714770389

meiravgri (approved):

DvirDukhan (approved):

PR #412, opened 2023-08-01
Refactor query result - [MOD-5530]

GuyAv46:

Remove the `arr_cpp.h` array implementation and moved to use vectors.
Also, make `VecSimQueryResult_List` an opaque type and change the API to return a pointer to it instead of by value, and rename it to `VecSimQueryReply`.

This PR enables compiling the library with GCC 12 and 13 as well.

1. #414 
2. MOD-5524

1. remove the `arr_cpp.h` implementation
3. changes in API (internally and external)

**Mark if applicable**


src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -870,47 +871,51 @@ TieredHNSWIndex<DataType, DistType>::TieredHNSW_BatchIterator::~TieredHNSW_Batch
...
         this->index->flatIndexGuard.lock_shared();
-        this->flat_results = this->flat_iterator->getNextResults(n_res, BY_SCORE_THEN_ID);
+        auto cur_flat_results = this->flat_iterator->getNextResults(n_res, BY_SCORE_THEN_ID);
         this->index->flatIndexGuard.unlock_shared();
         // This is also the only time `getNextResults` on the BF iterator can fail.
-        if (VecSim_OK != flat_results.code) {
-            return flat_results;
+        if (VecSim_OK != cur_flat_results->code) {
+            return cur_flat_results;
         }
```
meiravgri, line 895: should this->flat_results be updated?

meiravgri (commented):
will done, looks like a really tidious PR
I do think that some re-organization needs to be done in the query results files
i suggest renaming query_result.h to query_result_c_api and divided it into query_result, query_list, and query_iterator sections.
I didn't quite understand why is query_result_struct belongs to a seperated file, maybe a btter naming would clearify its purpose.

src/VecSim/algorithms/hnsw/hnsw_tiered.h
GuyAv46: No.
1. We return them all. `this->flat_results` holds results that are "cached" for the next batch
2. We got a timeout. Most likely the iterator is going to be freed soon

meiravgri (dismissed):

src/VecSim/algorithms/hnsw/hnsw.h
```diff
@@ -193,10 +192,10 @@ class HNSWIndex : public VecSimIndexAbstract<DistType>,
     searchBottomLayer_WithTimeout(idType ep_id, const void *data_point, size_t ef, size_t k,
                                   void *timeoutCtx, VecSimQueryResult_Code *rc) const;
     template <bool has_marked_deleted>
-    VecSimQueryResult *searchRangeBottomLayer_WithTimeout(idType ep_id, const void *data_point,
-                                                          double epsilon, DistType radius,
-                                                          void *timeoutCtx,
-                                                          VecSimQueryResult_Code *rc) const;
+    vecsim_stl::vector<VecSimQueryResult>
```
DvirDukhan, line 195: should we add this `using vecsim_stl::vector<VecSimQueryResult>` at vecsim stl?

DvirDukhan (approved):

PR #413, opened 2023-08-04
Provide GPU-accelerated vector indexes with RAFT

wphicks:
This PR is a replacement for #377, which fell out of date with changes on main. This PR provides access to GPU-accelerated index types provided by the RAPIDS RAFT library. It introduces two new index types: IVF and Tiered IVF, which provide access to both IVF Flat and IVFPQ indexes depending on how they are configured.

The intention of this initial integration is to provide GPU-based performance improvements for building indexes and batched searches. A later PR will introduce another index type which will offer perf improvement (relative to CPU HNSW) for single-query searches as well.

This PR does not substantively modify existing objects. A `clear` method is added to the brute force index to allow it to be reset with a single call.

**This PR is still a work in progress.** Remaining tasks:

- [ ] Update index-to-index transfer in tiered index to make use of asynchronous job infrastructure (Work on this was started but not completed)
- [ ] Add vector deletion once enabled by RAFT (https://github.com/rapidsai/raft/issues/1565)
- [ ] Add tests
- [ ] Add source files instantiating the index types provided by the current headers
- [ ] Switch back to mainline RAFT after merge of https://github.com/rapidsai/raft/pull/1716

Tasks to be tackled in later PRs:

- Integrating CAGRA indexes
- Multivalue support
- Vector update support

Questions for this PR:

- Should the IVF indexes be named in such a way as to indicate that they are GPU indexes in order to allow for possible CPU IVF support in the future?
- Based on initial feedback for 377, I have merged IVF Flat and IVF PQ into a single index type, with the choice of algorithm determined by the configuration parameters. Does this design meet the needs mentioned, or do we prefer the earlier separation?

**Mark if applicable**

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,9 @@
-cmake_minimum_required(VERSION 3.10)
+option(USE_CUDA "Build Cuda code" On)
+if(USE_CUDA)
+  cmake_minimum_required(VERSION 3.23.1 FATAL_ERROR)
+else()
+  cmake_minimum_required(VERSION 3.10)
+endif()
 cmake_policy(SET CMP0077 NEW)
 
 set(CMAKE_CXX_STANDARD 20)
@@ -23,6 +28,18 @@ include(cmake/san.cmake)
 
 # ----------------------------------------------------------------------------------------------
 project(VectorSimilarity)
+
+if (USE_CUDA)
+    # Enable CUDA compilation for this project
+    enable_language(CUDA)
+ 	# Add definition for conditional compilation of CUDA components
+	add_definitions(-DUSE_CUDA)
+    # Perform all RAFT-specific CMake setup
+	include(cmake/raft.cmake)
+    # Required flags for compiling RAFT
+    set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --expt-extended-lambda --expt-relaxed-constexpr")
+    set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -O3 -std=c++17")
+endif()
 
 # Only do these if this is the main project, and not if it is included through add_subdirectory
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 option(USE_CUDA "Build Cuda code" On)
 if(USE_CUDA)
-  cmake_minimum_required(VERSION 3.23.1 FATAL_ERROR)
+  cmake_minimum_required(VERSION 3.26.4 FATAL_ERROR)
 else()
   cmake_minimum_required(VERSION 3.10)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,7 +44,7 @@ endif()
 # Only do these if this is the main project, and not if it is included through add_subdirectory
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fexceptions -fPIC ${CLANG_SAN_FLAGS} ${LLVM_CXX_FLAGS} ${COV_CXX_FLAGS}")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fexceptions -fPIC -pthread ${CLANG_SAN_FLAGS} ${LLVM_CXX_FLAGS} ${COV_CXX_FLAGS}")
 set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_LINKER_FLAGS} ${LLVM_LD_FLAGS}")
 set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_LINKER_FLAGS} ${LLVM_LD_FLAGS}")
 
--- a/src/VecSim/algorithms/brute_force/brute_force.h
+++ b/src/VecSim/algorithms/brute_force/brute_force.h
@@ -35,6 +35,11 @@ protected:
 public:
     BruteForceIndex(const BFParams *params, const AbstractIndexInitParams &abstractInitParams);
 
+    void clear() {
+      idToLabelMapping.clear();
+      vectorBlocks.clear();
+      count = idType{};
+    }
     size_t indexSize() const override;
     size_t indexCapacity() const override;
     vecsim_stl::vector<DistType> computeBlockScores(const DataBlock &block, const void *queryBlob,
--- a/src/VecSim/algorithms/brute_force/brute_force.h
+++ b/src/VecSim/algorithms/brute_force/brute_force.h
@@ -36,9 +36,9 @@ public:
     BruteForceIndex(const BFParams *params, const AbstractIndexInitParams &abstractInitParams);
 
     void clear() {
-      idToLabelMapping.clear();
-      vectorBlocks.clear();
-      count = idType{};
+        idToLabelMapping.clear();
+        vectorBlocks.clear();
+        count = idType{};
     }
     size_t indexSize() const override;
     size_t indexCapacity() const override;
--- a/src/VecSim/algorithms/ivf/ivf.cuh
+++ b/src/VecSim/algorithms/ivf/ivf.cuh
@@ -87,7 +87,7 @@ private:
 public:
     IVFIndex(const IVFParams *ivfParams, const AbstractIndexInitParams & commonParams)
         : VecSimIndexAbstract<dist_type>{commonParams},
-          res_{}, //TODO(wphicks): Construct smartly
+          res_{raft::resource_manager::get_device_resources()},
           build_params_{[ivfParams](){
             auto result = ivfParams->pqBits > 0 ?
               build_params_t{std::in_place_index<1>} : 
@@ -132,7 +132,7 @@ public:
                   bool overwrite_allowed = true) override {
         return addVectorBatch(vector_data, &label, 1, overwrite_allowed);
     }
-    auto addVectorBatch(const void *vector_data, labelType *label, size_t batch_size,
+    auto addVectorBatchAsync(const void *vector_data, labelType *label, size_t batch_size,
                        bool overwrite_allowed = true) {
       // Allocate memory on device to hold vectors to be added
       auto vector_data_gpu =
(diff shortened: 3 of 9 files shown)


Spartee:
@wphicks also, probably need to sign the CLA as well.

src/VecSim/vec_sim_common.h
```diff
@@ -130,9 +136,46 @@ typedef struct {
...
+                  // compression. If set to 0, a heuristic will be used to
+                  // select the dimensionality.
+
+    IVFPQCodebookKind codebookKind;
+    CudaType lutType;
+    CudaType internalDistanceType;
+    double preferredShmemCarveout; // Fraction of GPU's unified memory / L1
+                                   // cache to be used as shared memory
+
+} IVFParams;
```
DvirDukhan, line 168: can we `RAFTIVFParams` this?

src/VecSim/vec_sim_common.h
```diff
@@ -130,9 +136,46 @@ typedef struct {
...
+
+typedef struct {
```
DvirDukhan, line 170: I don't believe we need this struct definition as is
you can send the `IVFParams` as `primaryIndexParams` when creating `TieredIndexParams`
if you need something specific for you tiered index management layer, this is the case where you need to create a specific struct 
see 
```
    union {
        TieredHNSWParams tieredHnswParams;
    } specificParams;
``` 
at `TieredIndexParams`

src/VecSim/algorithms/brute_force/brute_force.h
```diff
@@ -35,6 +35,11 @@ class BruteForceIndex : public VecSimIndexAbstract<DistType> {
 public:
     BruteForceIndex(const BFParams *params, const AbstractIndexInitParams &abstractInitParams);
 
+    void clear() {
```
DvirDukhan, line 38: when do we need to use this?

src/VecSim/algorithms/ivf/ivf.cuh
DvirDukhan, line 1: rename to raft_ivf please

src/VecSim/algorithms/ivf/ivf.cuh
```diff
@@ -0,0 +1,288 @@
...
+
+#include <raft/core/device_resources.hpp>
+#include <raft/core/error.hpp>
+#include <raft/distance/distance_types.hpp>
+#include <raft/neighbors/ivf_flat.cuh>
+#include <raft/neighbors/ivf_flat_types.hpp>
+#include <raft/neighbors/ivf_pq.cuh>
+#include <raft/neighbors/ivf_pq_types.hpp>
+
+#pragma once
```
DvirDukhan, line 25: is this the right place?

src/VecSim/algorithms/ivf/ivf.cuh
```diff
@@ -0,0 +1,288 @@
...
+
+inline auto constexpr GetRaftDistanceType(VecSimMetric vsm) {
+    auto result = raft::distance::DistanceType{};
+    switch (vsm) {
+    case VecSimMetric_L2:
+        result = raft::distance::DistanceType::L2Expanded;
+        break;
+    case VecSimMetric_IP:
+        result = raft::distance::DistanceType::InnerProduct;
```
DvirDukhan, line 34: side note: our IP is actually 1-IP
we can deal with this later

src/VecSim/algorithms/ivf/ivf.cuh
```diff
@@ -0,0 +1,288 @@
...
+        // Allocate memory on device to hold vector labels
+        auto label_gpu = raft::make_device_vector<labelType>(res_, batch_size);
+
+        // Copy vector data to previously allocated device buffer
+        raft::copy(vector_data_gpu.data_handle(), static_cast<DataType const *>(vector_data),
+                   this->dim * batch_size, res_.get_stream());
+        // Copy label data to previously allocated device buffer
+        raft::copy(label_gpu.data_handle(), label, batch_size, res_.get_stream());
+
+        if (std::holds_alternative<raft::neighbors::ivf_flat::index_params>(build_params_)) {
```
DvirDukhan, line 152: Is there a way to avoid this `if`?

src/VecSim/algorithms/ivf/ivf.cuh
```diff
@@ -0,0 +1,288 @@
...
+                          inner.preferred_shmem_carvout = ivfParams->preferredShmemCarveout;
+                      }
+                  },
+                  result);
+              return result;
+          }()},
+          index_{std::nullopt} {}
+    auto addVector(const void *vector_data, labelType label,
+                   bool overwrite_allowed = true) override {
+        return addVectorBatch(vector_data, &label, 1, overwrite_allowed);
```
DvirDukhan, line 136: Mandatory due to our requirements? I guess batch size 1 is not efficient here?

src/VecSim/algorithms/ivf/ivf.cuh
```diff
@@ -0,0 +1,288 @@
...
+            // Allocate memory on device for neighbor results
+            auto neighbors_gpu = raft::make_device_vector<labelType>(res_, k);
+            // Allocate memory on device for distance results
+            auto distances_gpu = raft::make_device_vector<dist_type>(res_, k);
+            // Copy query vector to device
+            raft::copy(vector_data_gpu.data_handle(), static_cast<data_type>(queryBlob), this->dim,
+                       res_.get_stream());
+
+            // Perform correct search based on index type
+            if (std::holds_alternative<raft::neighbors::ivf_flat::index>(index_)) {
```
DvirDukhan, line 214: same question about this `if` here

src/VecSim/algorithms/ivf/ivf_tiered.cuh
```diff
@@ -0,0 +1,85 @@
...
+        // TODO(wphicks)
+        // If in flatIndex, delete
+        // If being transferred to backend, wait for transfer
+        // If in backendIndex, delete
+    }
+
+    auto indexSize() {
+        auto frontend_lock = std::scoped_lock(this->flatIndexGuard);
+        auto backend_lock = std::scoped_lock(this->mainIndexGuard);
+        return (getBackendIndex().indexSize() + this->frontendIndex.indexSize());
```
DvirDukhan, line 34: Assuming transfer is "atomic" and does not cause duplications?

src/VecSim/algorithms/ivf/ivf_tiered.cuh
```diff
@@ -0,0 +1,85 @@
...
+        for (auto block_id = 0; block_id < vectorBlocks.size(); ++block_id) {
+            auto *in_begin = reinterpret_cast<DataType *>(vectorBlocks[block_id].getElement(0));
+            auto length = vectorBlocks[block_id].getLength();
+            std::copy(in_begin, in_begin + length, out);
+            out += length;
+        }
+
+        auto backend_lock = std::scoped_lock(this->mainIndexGuard);
+        this->flatBuffer->clear();
+        frontend_lock.unlock();
```
DvirDukhan, line 81: needs to be called after `addVectorBatch` otherwise we can have both indexes empty

DvirDukhan (commented):
Thanks for the work so far
Please see comments

src/VecSim/algorithms/ivf/ivf.cuh
wphicks: Not unless we want to split IVF flat and IVF PQ back out into separate indexes again.

One possible alternate spelling would be to use `std::visit` with a constexpr if to distinguish the types. This has the disadvantage that compilers sometimes have trouble creating an efficient `std::visit` over small variants, but it has the advantage of providing obviously idiomatic access to the variant.

If performance is what we're worried about, I would avoid `std::visit` but add a compile-time assumption (via `__builtin_unreachable`, `__assume` or similar based on the compiler) that the variant contains the expected type prior to the `std::get` call. That usually allows the compiler to avoid the check that it must ordinarily make to see if it needs to throw a `std::bad_variant_access`.

If readability is what we are concerned about, the bodies of those if clauses could be separated into their own functions.

src/VecSim/algorithms/ivf/ivf.cuh
wphicks: See above

CMakeLists.txt
```diff
@@ -24,10 +29,22 @@ include(cmake/san.cmake)
...
+    # Required flags for compiling RAFT
+    set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --expt-extended-lambda --expt-relaxed-constexpr")
+    set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -O3 -std=c++17")
+endif()
+
 # Only do these if this is the main project, and not if it is included through add_subdirectory
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fexceptions -fPIC ${CLANG_SAN_FLAGS} ${LLVM_CXX_FLAGS} ${COV_CXX_FLAGS}")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fexceptions -fPIC -pthread ${CLANG_SAN_FLAGS} ${LLVM_CXX_FLAGS} ${COV_CXX_FLAGS}")
```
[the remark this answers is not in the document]
wphicks: I actually meant to ask you about that. I was having trouble compiling without it (even before making any of my other changes), but I'll have to recreate the failing build log. Will either post it here or as a separate issue.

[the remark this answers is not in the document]
DvirDukhan: please do, thanks!

src/VecSim/algorithms/ivf/ivf.cuh
DvirDukhan: Can you please share a suggestion for the last proposal, so we can iterate? (could be here or on slack).  I'm thinking about modularity and code reusability here.

src/VecSim/algorithms/ivf/ivf_tiered.cuh
DvirDukhan: Yes you are right, I missed the backend lock,
 Additional question regarding this:
 when calling 
 ``` c++
     void clear() {
        idToLabelMapping.clear();
        vectorBlocks.clear();
        count = idType{};
    }
```
you are not clearing the label set of the frontend index. Can this cause issues when calling multiple times to `transferToBackend` with an increasing set of IDs?

File: src/VecSim/algorithms/brute_force/brute_force.h (hunk 3 above)
Comment (lowener) [reply src/VecSim/algorithms/brute_force/brute_force.h]: The `clear()` function is used in the Tiered Raft IVF index: The vectors from the flat buffers are all transferred to the backend index, and then the flat buffer is cleared.

**Review by alonre24 — COMMENTED** [review]
Some small questions/comments after doing a high-level review - before we move on to POC

File: src/VecSim/algorithms/raft_ivf/ivf_tiered.h [hunk 12]
```diff
@@ -0,0 +1,246 @@
+#pragma once
+
+#include <mutex>
```
GuyAv46, line 3: `<algorithm>` is missing and it messes up with `std::set_union` and `std::copy_n`
```suggestion
#include <mutex>
#include <algorithm>
```

PR #419, opened 2023-08-24
Reduce locks for adhoc BF 

alonre24:

In ad-hoc BF mode of the tiered index, we call internally to `getDistanceFrom` of the flat and HNSW indexes. These calls access the indexes' data, and it is not safe to run insert/delete operations in parallel. Until now we were acquiring the locks internally, but since this is usually called for every vector individually, we see that the overhead of acquiring and releasing the locks is significant in that case.
Hence we introduce the new API for acquiring and releasing the locks for shared ownership and assume that the caller acquires the locks while calling the tiered index `getDistanceFrom`.

**Mark if applicable**

### DIFF
--- a/src/VecSim/algorithms/hnsw/hnsw_multi.h
+++ b/src/VecSim/algorithms/hnsw/hnsw_multi.h
@@ -130,20 +130,20 @@ double HNSWIndex_Multi<DataType, DistType>::getDistanceFromInternal(labelType la
     DistType dist = INVALID_SCORE;
 
     // Check if the label exists in the index, return invalid score if not.
-//    if (Safe)
-//        this->indexDataGuard.lock_shared();
+    //    if (Safe)
+    //        this->indexDataGuard.lock_shared();
     auto it = this->labelLookup.find(label);
     if (it == this->labelLookup.end()) {
-//        if (Safe)
-//            this->indexDataGuard.unlock_shared();
+        //        if (Safe)
+        //            this->indexDataGuard.unlock_shared();
         return dist;
     }
 
     // Get the vector of ids associated with the label.
     // Get a copy if `Safe` is true, otherwise get a reference.
     decltype(auto) IDs = getCopyOrReference<Safe>(it->second);
-//    if (Safe)
-//        this->indexDataGuard.unlock_shared();
+    //    if (Safe)
+    //        this->indexDataGuard.unlock_shared();
 
     // Iterate over the ids and find the minimum distance.
     for (auto id : IDs) {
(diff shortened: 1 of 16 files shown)


src/VecSim/vec_sim.h
```diff
@@ -199,6 +199,10 @@ void VecSimTieredIndex_GC(VecSimIndex *index);
 bool VecSimIndex_PreferAdHocSearch(VecSimIndex *index, size_t subsetSize, size_t k,
                                    bool initial_check);
 
+void VecSimTieredIndex_AcquireSharedLocks(VecSimIndex *index);
```
meiravgri, line 202: I suggest changing `VecSimIndex_GetDistanceFrom` to `VecSimIndex_GetDistanceFrom_Unsafe` and mentioning in the function description that the shared locks should be acquired

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -803,14 +814,18 @@ int TieredHNSWIndex<DataType, DistType>::deleteVector(labelType label) {
 // 3. label exists in both indexes - we may have some of the vectors with the same label in the flat
 //    buffer only and some in the Main index only (and maybe temporal duplications).
 //    So, we get the distance from both indexes and return the minimum.
+
+// IMPORTANT: this should be called when the *tiered index locks are locked for shared ownership*,
+// along with HNSW index data guard lock. That is since the internal getDistanceFrom calls access
+// the indexes' data, and it is not safe to run insert/delete operation in parallel. Also, we avoid
+// acquiring the locks internally, since this is usually called for every vector individually, and
+// the overhead of acquiring and releasing the locks is significant in that case.
```
meiravgri, line 822: I don't think that this is the (only) location for this comment. Should be part of the API.

src/VecSim/algorithms/hnsw/hnsw_multi.h
```diff
@@ -124,26 +120,19 @@ constexpr decltype(auto) getCopyOrReference(Arg &&arg) {
...
     auto it = this->labelLookup.find(label);
     if (it == this->labelLookup.end()) {
-        if (Safe)
-            this->indexDataGuard.unlock_shared();
         return dist;
     }
 
     // Get the vector of ids associated with the label.
     // Get a copy if `Safe` is true, otherwise get a reference.
-    decltype(auto) IDs = getCopyOrReference<Safe>(it->second);
```
meiravgri, line 144: can this function be removed if not in use anymore?

meiravgri (commented):
See comments below :)
The main suggestion is to include `UnSafe` in the API function's name so if we have an alternative safe solution in the future, we won't have to break the API again.
Also, since this can cause a bottleneck in writing requests, I think it is worth examining the option to first accumulate results up to some limit (on Redisearch side)  and then calculate the accumulated vectors distances.

DvirDukhan (approved):

meiravgri (approved):

PR #432, opened 2024-01-29
Dump neighbors in HNSW debug command [MOD-6496]

alonre24:

Add an API that will return all neighbors in HNSW for a given label in an HNSW index (multi is currently not supported). This API should be used for debugging purposes.
This also includes adding a method in HNSW for returning the ids of a label, and it is re-used in `markDelete` calls.  

**Mark if applicable**


GuyAv46 (commented):
1. I think we should move the debug API to a different h file, so it is obvious these commands are not part of the regular API
2. I think we should move the HNSW debug API to an abstract class and not as part of any index API, and then remove the HNSW debug command implementation from the brute force index and implement it on HNSW (and tiered) only

GuyAv46 (approved):

PR #438, opened 2024-04-03
Fix hardware optimization compilation - [MOD-6567]

GuyAv46:

We previously discovered that the way we add architecture optimization flags was wrong, and we "leaked" higher optimizations onto lower optimization level functions (e.g. an SSE function was compiled with the `-mavx` flag, and yielded a binary that an SSE only machine cannot use).

from [GCC documentation](https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html):

> GCC depresses SSEx instructions when -mavx is used. Instead, it generates new AVX instructions or AVX equivalence for all SSEx instructions when needed.
> 
> These options enable GCC to use these extended instructions in generated code, even without -mfpmath=sse. Applications that perform run-time CPU detection must compile separate files for each supported architecture, using the appropriate flags. In particular, the file containing the CPU detection code should be compiled without these options. 

This PR fixes the ill-formed `CMakeList.txt` files and adds new `.cpp` files for choosing the optimization function, which will each be compiled with the right flags

Note that a regression is expected for the SSE functions, as they got "too optimized"

**Mark if applicable**


src/VecSim/spaces/IP/IP_SSE_FP32.h
```diff
@@ -29,10 +29,10 @@ float FP32_InnerProductSIMD16_SSE(const void *pVect1v, const void *pVect2v, size
...
         if (residual % 4 == 3) {
             // Load 3 floats and set the last one to 0
-            v1 = _mm_load_ps(pVect1); // load 4 floats
-            v2 = _mm_load_ps(pVect2);
-            v1 = _mm_blend_ps(_mm_setzero_ps(), v1, 7); // set the last one to 0
-            v2 = _mm_blend_ps(_mm_setzero_ps(), v2, 7);
+            v1 = _mm_load_ss(pVect1); // load 1 float, set the rest to 0
+            v2 = _mm_load_ss(pVect2);
+            v1 = _mm_loadh_pi(v1, (__m64 *)(pVect1 + 1));
+            v2 = _mm_loadh_pi(v2, (__m64 *)(pVect2 + 1));
```
meiravgri, line 35: why IP and L2 are different?

meiravgri (commented):
Great!
Consider to add this to the description:
https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html

> GCC depresses SSEx instructions when -mavx is used. Instead, it generates new AVX instructions or AVX equivalence for all SSEx instructions when needed.
> 
> These options enable GCC to use these extended instructions in generated code, even without -mfpmath=sse. Applications that perform run-time CPU detection must compile separate files for each supported architecture, using the appropriate flags. In particular, the file containing the CPU detection code should be compiled without these options.

alonre24 (commented):
All good! just simplify a bit benchmarks where possible

src/VecSim/spaces/CMakeLists.txt
```diff
@@ -10,14 +10,17 @@ include(CheckCXXCompilerFlag)
 project(VectorSimilarity_Spaces)
 
 # TODO: Remove this once cpu_features get support for M1
-if(NOT APPLE)
-	include(${root}/cmake/cpu_features.cmake)
-elseif(CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "x86_64")
+if((NOT APPLE) OR (CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "x86_64"))
```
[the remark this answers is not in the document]
GuyAv46: No, because an Apple machine with Intel CPU supports the optimizations as well. only `ARM` machines cannot.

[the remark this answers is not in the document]
meiravgri: so isn't checking x86_64 enough?

src/VecSim/spaces/IP_space.cpp
```diff
@@ -30,21 +30,21 @@ dist_func_t<float> IP_FP32_GetDistFunc(size_t dim, const Arch_Optimization arch_
 
     switch (arch_opt) {
     case ARCH_OPT_AVX512_F:
-#ifdef __AVX512F__
+#ifdef OPT_AVX512F
```
meiravgri, line 33: why not use the compiler's built-in flags?

src/VecSim/spaces/CMakeLists.txt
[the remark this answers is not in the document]
GuyAv46: here, by checking `NOT APPLE` we support Linux on ARM (like the Ubuntu Linux we have on CI), which has support for the runtime CPU features. Currently, it doesn't help much since we don't have optimizations for ARM but in general, we want to include the runtime CPU feature to all the OSs and architectures we have basic support for.

src/VecSim/spaces/IP_space.cpp
GuyAv46: because we actually don't want to compile this file with the optimization flags, even though it worked until now. We don't want an unsupported instruction to slip in and cause a crash on non-supporting machines.
Using our own indication of whether we compiled the optimizations seems clean. Can we turn on these flags without using the optimizations (`-mavx`)? Is it good practice to `define` these flags without using the optimizations?

meiravgri: Thanks. i agree this is a better practice.

meiravgri (approved):

alonre24 (approved):

PR #453, opened 2024-04-14
FP16 distance functions

alonre24:

This PR introduces distance calculation between float16 vectors.

float16 is a typedef for unsigned short int, and the conversion functions from float32 to float16 and vice versa have been added in a new file: src/types/float16.h.

The process of distance calculation involves three steps:

Conversion of a chunk of float16 elements into a float32 vector (with or without intel intrinsic optimizations)
Calculation of the distance between the float32 vectors.
Return of the calculated float32 distance.
The implementation varies based on the register size and is optimized for different instruction sets:

256-bit registers: Requires AVX + F16C + FMA
512-bit registers: Requires AVX512_BW & AVX512_VL

The appropriate implementation is chosen at runtime based on the instructions supported by the hardware and the compiler. 

In addition to the implementation, this PR also includes benchmark and unit tests for the float16 distance calculation function.

**Mark if applicable**


src/VecSim/spaces/space_aux.cpp
```diff
@@ -16,10 +16,14 @@ Arch_Optimization getArchitectureOptimization() {
...
     cpu_features::X86Features features = cpu_features::GetX86Info().features;
-    if (features.avx512bw && features.avx512vbmi2) {
+    if (features.avx512bw && features.avx512vl) {
+        return ARCH_OPT_AVX512_BW_VL;
+    } else if (features.avx512bw && features.avx512vbmi2) {
         return ARCH_OPT_AVX512_BW_VBMI2;
     } else if (features.avx512f) {
         return ARCH_OPT_AVX512_F;
+    } else if (features.avx && features.f16c && features.fma3) {
+        return ARCH_OPT_F16C;
```
meiravgri, line 26: if these flags and avx2 exist, ARCH_OPT_F16C is set.
See comment in L2_space.cpp

src/VecSim/spaces/L2_space.cpp
```diff
@@ -145,6 +154,7 @@ dist_func_t<float> L2_BF16_GetDistFunc(size_t dim, const Arch_Optimization arch_
             *alignment = 16 * sizeof(bfloat16); // align to 256 bits.
         break;
 #endif
+    case ARCH_OPT_F16C:
```
meiravgri, line 157: since f16c is above avx2 in spaces_aux, i won't get here the most optimized function.

meiravgri (commented):
IMO, before we proceed, let's neglect the ARCH_OPT enum.

src/VecSim/types/float16.h
```diff
@@ -0,0 +1,102 @@
...
+
+inline float _interpret_as_float(uint32_t num) {
+    void *num_ptr = &num;
+    return *(float *)num_ptr;
+}
+
+inline int32_t _interpret_as_int(float num) {
+    void *num_ptr = &num;
+    return *(int32_t *)num_ptr;
+}
```
meiravgri, line 24:
```suggestion
#include <algorithm>
#include <bit>
namespace vecsim_types {

using float16 = uint16_t;

inline float _interpret_as_float(uint32_t num) {
    return std::bit_cast<float>(num);
}

inline int32_t _interpret_as_int(float num) {
    return std::bit_cast<int32_t>(num);
}
```

meiravgri (commented):
some astetic comments :)

src/VecSim/spaces/functions/implementation_chooser.h
```diff
@@ -45,6 +45,6 @@
 #define CHOOSE_IMPLEMENTATION(out, dim, chunk, func)                                               \
     do {                                                                                           \
         decltype(out) __ret_dist_func;                                                             \
-        switch (dim % chunk) { CASES##chunk(X, func) }                                             \
+        switch ((dim) % (chunk)) { CASES##chunk(X, func) }                                         \
```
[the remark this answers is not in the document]
alonre24: I worked on this in parallel to you (then revert my changes), I left this as it is a good practice to use parenthesis around macro variables

src/VecSim/types/float16.h
alonre24: nice!
so from which GCC version it is supported?

meiravgri: this is c++20 and was added as of GCC 11

alonre24: so we cannot support in all OSs :(

meiravgri (approved):
exciting

PR #454, opened 2024-04-16
[MOD-6764] Introduce bfloat16 datatype

meiravgri:
This PR introduces a new data type, BFLOAT16, to the VecSim API.

## New 
*  Implementation of BFLOAT16 in the indices factories.
* Addition of BFLOAT16 utilities (`BFLOAT16_STRING` etc.).
* Implementation of a normalization function for BFLOAT16.
* Unit and flow tests for BFLOAT16.
* Benchmarks for BFLOAT16, including a serialized HNSW index file uploaded to the AWS bucket.
## Changes
* Added Datatype to `VecSimIndexAbstract` template arguments.
* Changed `SetDistFunc` to `GetDistFunc`, now returning the distance function instead of assigning it to an output parameter.
* Moved normalization function from vec_utils to spaces directory.
* Flow tests now require Python 3.10.

**Mark if applicable**


tests/unit/test_bf16.cpp
```diff
@@ -0,0 +1,1101 @@
...
+#include "VecSim/vec_sim_debug.h"
+#include "VecSim/spaces/L2/L2.h"
+
+#include <random>
+
+using bfloat16 = vecsim_types::bfloat16;
+
+class BF16Test : public ::testing::Test {
+protected:
+    virtual void SetUp(HNSWParams &params) {
```
alonre24, line 19: Can't you just set it to be pure virtual? (=0)

alonre24 (commented):
Very clean and thorough work, well done!

src/VecSim/vec_sim_index.h
```diff
@@ -84,12 +85,11 @@ struct VecSimIndexAbstract : public VecSimIndexInterface {
     VecSimIndexAbstract(const AbstractIndexInitParams &params)
         : VecSimIndexInterface(params.allocator), dim(params.dim), vecType(params.vecType),
           dataSize(dim * VecSimType_sizeof(vecType)), metric(params.metric),
-          blockSize(params.blockSize ? params.blockSize : DEFAULT_BLOCK_SIZE), alignment(0),
-          lastMode(EMPTY_MODE), isMulti(params.multi), logCallbackCtx(params.logCtx) {
+          blockSize(params.blockSize ? params.blockSize : DEFAULT_BLOCK_SIZE),
+          distFunc(spaces::GetDistFunc<DataType, DistType>(metric, dim, &alignment)), alignment(0),
```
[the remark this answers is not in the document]
meiravgri: good catch

src/VecSim/algorithms/hnsw/hnsw_single_tests_friends.h
```diff
@@ -13,3 +13,5 @@ INDEX_TEST_FRIEND_CLASS(IndexAllocatorTest_test_hnsw_reclaim_memory_Test)
 INDEX_TEST_FRIEND_CLASS(HNSWTestParallel_parallelInsertSearch_Test)
 INDEX_TEST_FRIEND_CLASS(HNSWTieredIndexTest_testSizeEstimation_Test)
 INDEX_TEST_FRIEND_CLASS(HNSWTieredIndexTest_swapJobBasic_Test)
+friend class BF16HNSWTest_testSizeEstimation_Test;
```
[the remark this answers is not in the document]
meiravgri: the macro is for templated classes

src/python_bindings/bindings.cpp
```diff
@@ -92,18 +95,35 @@ class PyBatchIterator {
...
+            }
+            py::capsule free_when_done(data_numpy,
+                                       [](void *vector_data) { delete[](float *) vector_data; });
+            return py::array_t<float>(
+                {n_vectors, dim}, // shape
+                {dim * sizeof(float),
+                 sizeof(float)}, // C-style contiguous strides for the data type
+                data_numpy,      // the data pointer
+                free_when_done);
+        }
```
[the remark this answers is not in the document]
meiravgri: no, because DataType in this case is bfloat16 and we want to return float array

tests/unit/test_bf16.cpp
meiravgri: no because the argument type is different

alonre24 (approved):

src/VecSim/vec_sim_tiered_index.h
```diff
@@ -96,8 +96,15 @@ class VecSimTieredIndex : public VecSimIndexInterface {
     static VecSimWriteMode getWriteMode() { return VecSimIndexInterface::asyncWriteMode; }
 
 #ifdef BUILD_TESTS
-    inline VecSimIndexAbstract<DistType> *getFlatBufferIndex() { return this->frontendIndex; }
+    inline VecSimIndexAbstract<DataType, DistType> *getFlatBufferIndex() {
+        return this->frontendIndex;
+    }
     inline size_t getFlatBufferLimit() { return this->flatBufferLimit; }
+
+    virtual void fitMemory() {
```
GuyAv46, line 104: isn't it being used in testing? can we get it covered?

src/VecSim/types/bfloat16.h
```diff
@@ -8,17 +8,33 @@
...
 }
 
+static inline float bfloat16_to_float32(bfloat16 val) {
+    float result = 0;
+    bfloat16 *p_result = (bfloat16 *)&result + 1;
+    *p_result = val;
+    return result;
+}
+
+static inline float bfloat16_to_float32_bigEndian(bfloat16 val) {
```
GuyAv46, line 33: consider having a `constexpr` function that checks if the target arch is big/little endian, and use it to make `bfloat16_to_float32` fit for both (by adding the `1` offset in a constexpr manner according to the function result.
This way we don't need the code duplication and the recoverable implemetation

src/VecSim/spaces/spaces.h
```diff
@@ -13,12 +13,17 @@ namespace spaces {
...
-void SetDistFunc(VecSimMetric metric, size_t dim, dist_func_t<double> *index_dist_func,
-                 unsigned char *alignment);
+// Set the distance function for a given data type, metric and dimension. The alignment hint is
+// determined according to the chosen implementation and available optimizations.
+
+template <typename DataType, typename DistType>
+dist_func_t<DistType> GetDistFunc(VecSimMetric metric, size_t dim, unsigned char *alignment);
+
+template <typename DataType>
+using normalizeVector_f = void (*)(void *input_vector, const size_t dim);
```
GuyAv46, line 23: what does `_f` stand for? consider aligning the type name with `dist_func_t` or at least decide between camel and snake casing

src/VecSim/spaces/spaces.cpp
```diff
@@ -4,36 +4,77 @@
...
+template <>
+dist_func_t<double> GetDistFunc<double, double>(VecSimMetric metric, size_t dim,
+                                                unsigned char *alignment) {
+    switch (metric) {
+    case VecSimMetric_Cosine:
+    case VecSimMetric_IP:
+        return IP_FP64_GetDistFunc(dim, nullptr, alignment);
+    case VecSimMetric_L2:
+        return L2_FP64_GetDistFunc(dim, nullptr, alignment);
+    default:
```
GuyAv46, line 57: Why add a default here? Don't we compile with "warn incomplete switch enum"?

src/VecSim/algorithms/hnsw/hnsw.h
```diff
@@ -366,6 +366,10 @@ class HNSWIndex : public VecSimIndexAbstract<DistType>,
      */
     virtual void getDataByLabel(labelType label,
                                 std::vector<std::vector<DataType>> &vectors_output) const = 0;
+    virtual void fitMemory() {
```
GuyAv46, line 369: Can we cover it?

src/VecSim/algorithms/brute_force/brute_force.h
```diff
@@ -83,6 +83,10 @@ class BruteForceIndex : public VecSimIndexAbstract<DistType> {
      */
     virtual void getDataByLabel(labelType label,
                                 std::vector<std::vector<DataType>> &vectors_output) const = 0;
+    virtual void fitMemory() {
```
GuyAv46, line 86: can we cover it?

src/VecSim/vec_sim_tiered_index.h
meiravgri: only in the benchmark
i can write a unit test

src/VecSim/types/bfloat16.h
meiravgri: constexpr is checked at compilation time and we want it to be tested on runtime.
The best i can do is to unify big and little function under one templated function (similar to what i did in the normalization function)
but the test should take place at run time

src/VecSim/spaces/spaces.h
meiravgri: function 🤷 
it's not new, i just moved it from vec_utils

src/VecSim/spaces/spaces.cpp
meiravgri: The compiler will complain that the function doesn't return a value

src/VecSim/algorithms/brute_force/brute_force.h
meiravgri: BESEDER

src/VecSim/types/bfloat16.h
GuyAv46: is the compiled binary compatible with other arches? I don't think so. And if not, it is something we can decide in compile time. Can we verify that?

src/VecSim/spaces/spaces.cpp
GuyAv46: so consider moving the exception outside the switch, so future metrics won't automatically get covered by the default in the future

meiravgri: right good point

src/VecSim/types/bfloat16.h
meiravgri: I pushed a fix that minimizes code duplications

GuyAv46 (approved):

PR #470, opened 2024-05-26
Test alignment decision - [MOD-6998]

GuyAv46:

Add tests for the alignment hint calculation (both logic [unit] and when creating an index [flow])

**Mark if applicable**

### DIFF
--- a/src/VecSim/spaces/spaces.cpp
+++ b/src/VecSim/spaces/spaces.cpp
@@ -24,9 +24,9 @@ dist_func_t<float> GetDistFunc<vecsim_types::bfloat16, float>(VecSimMetric metri
     switch (metric) {
     case VecSimMetric_Cosine:
     case VecSimMetric_IP:
-        return IP_BF16_GetDistFunc(dim, nullptr, alignment);
+        return IP_BF16_GetDistFunc(dim, alignment);
     case VecSimMetric_L2:
-        return L2_BF16_GetDistFunc(dim, nullptr, alignment);
+        return L2_BF16_GetDistFunc(dim, alignment);
     }
     throw std::invalid_argument("Invalid metric");
 }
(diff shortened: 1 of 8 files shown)


src/VecSim/spaces/spaces.cpp
```diff
@@ -24,9 +24,9 @@ dist_func_t<float> GetDistFunc<vecsim_types::bfloat16, float>(VecSimMetric metri
     switch (metric) {
     case VecSimMetric_Cosine:
     case VecSimMetric_IP:
-        return IP_BF16_GetDistFunc(dim, nullptr, alignment);
+        return IP_BF16_GetDistFunc(dim, alignment);
```
meiravgri, line 27: noice

tests/unit/test_spaces.cpp
```diff
@@ -315,36 +328,50 @@ TEST_P(FP32SpacesOptimizationTest, FP32InnerProductTest) {
         v2[i] = (float)(i + 1.5);
     }
 
+    auto expected_alignment = [](size_t reg_bit_size, size_t dim) {
+        size_t elements_in_reg = reg_bit_size / sizeof(float) / 8;
+        return (dim % elements_in_reg == 0) ? elements_in_reg * sizeof(float) : 0;
+    };
```
meiravgri, line 334: The alignment is actually independent of the type, no? 
`alignment = reg_bit_size / 8;`

meiravgri (commented):
Clean as always.
one lambda declaration is enough as the alignment can be concluded from the register size.

tests/unit/test_spaces.cpp
GuyAv46: The alignment itself yes, but we only align if the `dim * sizeof(element)` is a multiplication of the register byte size

meiravgri: sure, i just meant that there is no need for a dedicated lambda for each type

meiravgri (approved):

PR #473, opened 2024-06-04
[MOD-7177] Add compilers support for CPU flags check in spaces test

meiravgri:
This PR introduces compiler **support in cpu flags checks in the spaces test**. This ensures that any code not supported by the compiler is excluded at compile time. 
The process of checking compiler CPU flags and creating macros accordingly during compilation has been moved to a separate file. This file is now included in every project's CMake that relies on these checks and macros. 

In addition to this, the PR also introduces several other changes:
**GCC Version Upgrade in Rocky8:** The GCC version in Rocky8 has been upgraded to version 11. This version includes an assembler that supports AVX512_BF16 assembly instructions.

**Deprecation of Bionic 18:** This includes removing the Bionic workflow and the installation script.

**Mark if applicable**


.github/workflows/rocky8.yml
```diff
@@ -7,3 +7,6 @@ jobs:
     uses: ./.github/workflows/task-unit-test.yml
     with:
       container: rockylinux:8
+    defaults:
+      run:
+        shell: bash -l -eo pipefail {0}
```
GuyAv46, line 12: Is this needed? We set it again in `task-unit-test.yml`

.install/ubuntu_18.04.sh
GuyAv46, line 1: This file should be deleted no?

cmake/x86_64InstructionFlags.cmake
GuyAv46, line 1: consider renaming to something that indicates it sets flags **and** add definitions

src/VecSim/spaces/CMakeLists.txt
```diff
@@ -17,74 +16,56 @@ set(OPTIMIZATIONS "")
...
-	CHECK_CXX_COMPILER_FLAG(-mavx512bw CXX_AVX512BW)
-	CHECK_CXX_COMPILER_FLAG(-mavx512vbmi2 CXX_AVX512VBMI2)
-	CHECK_CXX_COMPILER_FLAG(-mavx512f CXX_AVX512F)
-	CHECK_CXX_COMPILER_FLAG(-mavx2 CXX_AVX2)
-	CHECK_CXX_COMPILER_FLAG(-mavx CXX_AVX)
-	CHECK_CXX_COMPILER_FLAG(-mf16c CXX_F16C)
-	CHECK_CXX_COMPILER_FLAG(-mfma CXX_FMA)
-	CHECK_CXX_COMPILER_FLAG(-msse3 CXX_SSE3)
-	CHECK_CXX_COMPILER_FLAG(-msse CXX_SSE)
+	include(${root}/cmake/x86_64InstructionFlags.cmake)
```
GuyAv46, line 19: If you decide not to rename the file, consider commenting that it will define the `CXX_*` conditions we use here and the compilation defiitions

tests/unit/test_spaces.cpp
```diff
@@ -282,6 +282,7 @@ TEST_P(FP32SpacesOptimizationTest, FP32L2SqrTest) {
 
     dist_func_t<float> arch_opt_func;
     float baseline = FP32_L2Sqr(v, v2, dim);
+#ifdef OPT_AVX512F
     if (optimization.avx512f) {
```
GuyAv46, line 286: Nice! We should have done it in this order on the spaces files as well :(

.github/workflows/rocky8.yml
meiravgri: leftovers

.install/ubuntu_18.04.sh
meiravgri: yes

cmake/x86_64InstructionFlags.cmake
meiravgri: The file's name is general and doesn't imply which specific operations are done.
I'll add a comment

tests/unit/test_spaces.cpp
meiravgri: Yeah i wondered if there's a good reason for that
I'll change it 🤫

GuyAv46 (approved):

PR #475, opened 2024-06-20
[MOD-7259] Restore bionic workkflow

meiravgri:
This PR reinstates support for Ubuntu 18, which was previously removed in [PR #472](https://github.com/RedisAI/VectorSimilarity/pull/472).

The assembler provided by the binutils package in Ubuntu 18 does not support the assembly instructions included in the AVX512BF16 CPU flag. As a result, compilation errors are raised when attempting to compile code that requires this flag.

To avoid these errors, this PR disables the compilation of any code that requires the AVX512BF16 CPU flag when building on Ubuntu 18.

### DIFF
--- a/.github/workflows/flow-temp.yml
+++ b/.github/workflows/flow-temp.yml
@@ -9,37 +9,10 @@ name: temporary testing
 
 on:
   push:
-    branches-ignore: ['**'] # ignore all branches. Comment this line to run your workflow below on every push.
+    # branches-ignore: ['**'] # ignore all branches. Comment this line to run your workflow below on every push.
 jobs:
-  macos:
+  bionic:
     uses: ./.github/workflows/task-unit-test.yml
     with:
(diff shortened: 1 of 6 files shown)


.github/workflows/flow-temp.yml
```diff
@@ -9,37 +9,10 @@ name: temporary testing
 
 on:
   push:
-    branches-ignore: ['**'] # ignore all branches. Comment this line to run your workflow below on every push.
+    # branches-ignore: ['**'] # ignore all branches. Comment this line to run your workflow below on every push.
```
GuyAv46, line 12: don't forget to revert

GuyAv46 (approved):

.github/workflows/flow-temp.yml
meiravgri: 🤭

PR #477, opened 2024-06-26
[MOD - 6778] AVX512_FP16 distance functions

meiravgri:
This PR introduces enhancements to distance calculation by leveraging the `AVX512_FP16` instruction set, which is supported by CPUs with the `AVX512_FP16` flag.

## Minimum dimension
`_mm512_reduce_add_ph` is called once,  at the end of the distance calculation function, to reduce a 512-bit register into a single `_FLoat16` variable.

It was found that this function incurs a significant overhead, leading to a substantial degradation in distance calculation performance compared to alternative implementations. This overhead becomes less significant at higher dimensions, as illustrated in the following graphs:
image
image

Therefore, we opt for the distance calculation function with this optimization only when the dimension exceeds a certain threshold:
**for IP**: **500** and higher
**for L2**: **440** and higher 

## GCC version
`GCC12` introduced two important features for this PR:
1.  `_Float` support.
2. `AVX512_FP16` instruction support.
Therefore, GCC is upgraded on platforms that support GCC12+.
Note: The default `gcov` version is also set to align with the GCC version. 

## Tests
1. Unit tests for the new implementations are done on higher dimensions to ensure the right function is chosen.
2. A new test unit has been added to compare the conversion from `float` to the `_Float` native type against our manual conversion implementation, and ensure they are identical.

## Additional changes
1. We now use explicit constructors for `float16` and `bfloat16` to avoid implicit casts. Previously, attempting to initialize a `float16`/`bfloat16` variable with a `float` value resulted in an implicit conversion to uint16_t, leading to a garbage value.  The explicit constructor prevents this issue by failing the compilation if an implicit conversion is attempted.
[MOD-7335](https://redislabs.atlassian.net/browse/MOD-7335) aims to introduce an additional constructor that expect `float` and will call the correct conversion function.
2. The names of avx512f function sets were changed from AVX512 to AVX512F to align with the naming convention of _FLAG_TYPE_.
3. The `override` keyword has been added to virtual methods in fp16 unit tests to eliminate compilation warnings

[MOD-7335]: https://redislabs.atlassian.net/browse/MOD-7335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/flow-temp.yml
```diff
@@ -11,8 +11,21 @@ on:
...
-      container: ubuntu:bionic
+      container: rockylinux:8
       run-valgrind: false
+      run-codecov: false
+  rocky9:
+    uses: ./.github/workflows/task-unit-test.yml
+    with:
+      container: rockylinux:9
+      run-valgrind: false
+      run-codecov: false
```
DvirDukhan, line 25: still needed?

.github/workflows/task-unit-test.yml
```diff
@@ -45,3 +49,6 @@ jobs:
         with:
           name: valgrind tests reports
           path: bin/Linux-x86_64-debug/unit_tests/Testing/Temporary/
+      - name: codecov
+        if: ${{ inputs.run-codecov }}
+        run: make coverage
```
DvirDukhan, line 54: I think this one is reduendent, no?

.install/ubuntu_22.04.sh
```diff
@@ -4,5 +4,8 @@ export DEBIAN_FRONTEND=noninteractive
 MODE=$1 # whether to install using sudo or not
 
 $MODE apt-get update -qq || true
-$MODE apt-get install -yqq git wget build-essential valgrind lcov
+$MODE apt-get install -yqq gcc-12 g++-12 git wget build-essential valgrind lcov
```
DvirDukhan, line 7: why not 13?

src/VecSim/spaces/IP_space.cpp
```diff
@@ -168,11 +169,18 @@ dist_func_t<float> IP_FP16_GetDistFunc(size_t dim, unsigned char *alignment, con
     auto features = (arch_opt == nullptr)
                         ? cpu_features::GetX86Info().features
                         : *static_cast<const cpu_features::X86Features *>(arch_opt);
+#ifdef OPT_AVX512_FP16
+    if (features.avx512_fp16 && dim >= 500) {
```
GuyAv46, line 173: Consider linking to the PR here (so we can see the graphs and have an explanation for the dimension condition

.github/workflows/task-unit-test.yml
```diff
@@ -17,6 +17,10 @@ on:
         description: 'Run valgrind tests'
         type: boolean
         default: true
+      run-codecov:
```
GuyAv46, line 20: is this temporary?

src/VecSim/spaces/IP/IP_AVX512FP16_FP16.h
```diff
@@ -0,0 +1,49 @@
...
+                                         size_t dimension) {
+    auto *pVect1 = (float16 *)pVect1v;
+    auto *pVect2 = (float16 *)pVect2v;
+
+    const float16 *pEnd1 = pVect1 + dimension;
+
+    __m512h sum = _mm512_setzero_ph();
+
+    if constexpr (residual) {
+        __mmask32 mask = (1LU << residual) - 1;
```
GuyAv46, line 34:
```suggestion
        constexpr __mmask32 mask = (1LU << residual) - 1;
```

src/VecSim/spaces/L2/L2_AVX512FP16_FP16.h
```diff
@@ -0,0 +1,52 @@
...
+float FP16_L2SqrSIMD32_AVX512FP16(const void *pVect1v, const void *pVect2v, size_t dimension) {
+    auto *pVect1 = (float16 *)pVect1v;
+    auto *pVect2 = (float16 *)pVect2v;
+
+    const float16 *pEnd1 = pVect1 + dimension;
+
+    __m512h sum = _mm512_setzero_ph();
+
+    if constexpr (residual) {
+        __mmask32 mask = (1LU << residual) - 1;
```
GuyAv46, line 35:
```suggestion
        constexpr __mmask32 mask = (1LU << residual) - 1;
```

src/VecSim/spaces/L2_space.cpp
```diff
@@ -161,11 +162,18 @@ dist_func_t<float> L2_FP16_GetDistFunc(size_t dim, unsigned char *alignment, con
     auto features = (arch_opt == nullptr)
                         ? cpu_features::GetX86Info().features
                         : *static_cast<const cpu_features::X86Features *>(arch_opt);
+#ifdef OPT_AVX512_FP16
+    if (features.avx512_fp16 && dim >= 440) {
```
GuyAv46, line 166: Consider linking to the PR here (so we can see the graphs and have an explanation for the dimension condition

tests/unit/test_spaces.cpp
```diff
@@ -742,4 +745,105 @@ TEST_P(FP16SpacesOptimizationTest, FP16L2SqrTest) {
 INSTANTIATE_TEST_SUITE_P(FP16OptFuncs, FP16SpacesOptimizationTest,
                          testing::Range(32UL, 32 * 2UL + 1));
 
+/** Since we are handling floats, the order of summation affect on the final result.
+ * This is very sifnificant when the entries are half precision floats, since the accumulated
```
DvirDukhan, line 749:
```suggestion
 * This is very siעnificant when the entries are half precision floats, since the accumulated
```

.github/workflows/flow-temp.yml
meiravgri: i disabled the workflow

tests/unit/test_spaces.cpp
```diff
@@ -742,4 +745,105 @@ TEST_P(FP16SpacesOptimizationTest, FP16L2SqrTest) {
...
+ * error is much higher than in single precision floats.
+ * In the following tests the error between the naive calculation to SIMD optimization function
+ * is allowed to be up to 1%. If we wanted to be accurate, we could have done the baseline
+ * calculations accumulating the results in a SIMD size vector and reduce the final result to float,
+ * but this is too complicated for the scope of this test.
+ * Special attention should be given to the implementation of the SIMD reduce function for float16,
+ * that has diffrent logic than the float32 and float64 reduce functions.
```
DvirDukhan, line 756:
```suggestion
 * that has different logic than the float32 and float64 reduce functions.
```

.github/workflows/task-unit-test.yml
meiravgri: it defaults to false and might be useful in the future

.github/workflows/task-unit-test.yml
meiravgri: it defaults to false and might be useful in the future

.install/ubuntu_22.04.sh
meiravgri: GCC13 requires ppa and i wanted the keep the installation as simple as possible

tests/unit/test_spaces.cpp
```diff
@@ -742,4 +745,105 @@ TEST_P(FP16SpacesOptimizationTest, FP16L2SqrTest) {
...
+ * For more info, refer to intel's intrinsinc guide.
```
DvirDukhan, line 757:
```suggestion
 * For more info, refer to intel's intrinsics guide.
```

src/VecSim/spaces/IP/IP_AVX512FP16_FP16.h
meiravgri: aww going to check if i havn't miss more of these in #481

tests/unit/test_spaces.cpp
```diff
@@ -742,4 +745,105 @@ TEST_P(FP16SpacesOptimizationTest, FP16L2SqrTest) {
...
+ */
+#ifdef OPT_AVX512_FP16
+// avx512_fp16 flag functions are only chosen for high dimensions
+class FP16SpacesOptimizationTestAdvanced : public testing::TestWithParam<size_t> {};
+
+TEST_P(FP16SpacesOptimizationTestAdvanced, FP16InnerProductTestAdv) {
+    auto optimization = cpu_features::GetX86Info().features;
+    if (optimization.avx512_fp16) {
+        size_t dim = GetParam();
+        float16 v1[dim], v2[dim];
```
DvirDukhan, line 767: don't we want to stop doing this?

tests/unit/test_spaces.cpp
```diff
@@ -742,4 +745,105 @@ TEST_P(FP16SpacesOptimizationTest, FP16L2SqrTest) {
...
+        dist_func_t<float> arch_opt_func;
+        unsigned char alignment = 0;
+        arch_opt_func = IP_FP16_GetDistFunc(dim, &alignment, &optimization);
+        ASSERT_EQ(arch_opt_func, Choose_FP16_IP_implementation_AVX512FP16(dim))
+            << "Unexpected distance function chosen for dim " << dim;
+        float dist = arch_opt_func(v1, v2, dim);
+        float f_baseline = baseline;
+        float error = std::abs((dist / f_baseline) - 1);
+        // Alow 1% error
+        ASSERT_LE(error, 0.01) << "AVX512 with dim " << dim << ", baseline: " << f_baseline
```
DvirDukhan, line 797:
```suggestion
        ASSERT_LE(error, 0.01) << "AVX512_FP16 with dim " << dim << ", baseline: " << f_baseline
```

tests/unit/test_spaces.cpp
```diff
@@ -742,4 +745,105 @@ TEST_P(FP16SpacesOptimizationTest, FP16L2SqrTest) {
...
+                               << ", dist: " << dist;
+        ASSERT_EQ(alignment, expected_alignment(512, dim)) << "AVX512 with dim " << dim;
```
DvirDukhan, line 799:
```suggestion
        ASSERT_EQ(alignment, expected_alignment(512, dim)) << "AVX512_FP16 with dim " << dim;
```

tests/unit/test_spaces.cpp
```diff
@@ -742,4 +745,105 @@ TEST_P(FP16SpacesOptimizationTest, FP16L2SqrTest) {
...
+        arch_opt_func = L2_FP16_GetDistFunc(dim, &alignment, &optimization);
+        ASSERT_EQ(arch_opt_func, Choose_FP16_L2_implementation_AVX512FP16(dim))
+            << "Unexpected distance function chosen for dim " << dim;
+        float dist = arch_opt_func(v1, v2, dim);
+        float f_baseline = baseline;
+        float error = std::abs((dist / f_baseline) - 1);
+        // Alow 1% error
+        ASSERT_LE(error, 0.01) << "AVX512 with dim " << dim << ", baseline: " << f_baseline
+                               << ", dist: " << dist;
+        ASSERT_EQ(alignment, expected_alignment(512, dim)) << "AVX512 with dim " << dim;
```
DvirDukhan, line 839: same as above

tests/unit/test_spaces.cpp
meiravgri: This is better 😄

tests/unit/test_types.cpp
```diff
@@ -0,0 +1,34 @@
...
+    for (size_t i = 0; i < runs; i++) {
+        float val = dis(gen);
+        _Float16 nativeCast = static_cast<_Float16>(val);
+        vecsim_types::float16 vecsimCast = vecsim_types::FP32_to_FP16(val);
+
+        uint16_t nativeBits, vecsimBits;
+        std::memcpy(&nativeBits, &nativeCast, sizeof(_Float16));
+        std::memcpy(&vecsimBits, &vecsimCast, sizeof(vecsim_types::float16));
+
+        ASSERT_EQ(nativeBits, vecsimBits) << "failed at iteration " << i << "with value " << val;
```
DvirDukhan, line 30: can't it flake?

.github/workflows/task-unit-test.yml
DvirDukhan: I don't like it. Workflows should be clean

meiravgri: ok :(

tests/unit/test_types.cpp
meiravgri: why? we expect to get identical results

tests/unit/test_spaces.cpp
meiravgri: We don't specify the exact cpu flag in any of the tests. It can be inferred by the line number

src/VecSim/spaces/IP/IP_AVX512FP16_FP16.h
meiravgri: We are all good 🥵

tests/unit/test_spaces.cpp
meiravgri: yes, we have an open issue to fix that

tests/unit/test_types.cpp
DvirDukhan: and if we don't?

meiravgri: Then we are lucky that we have this test

GuyAv46 (approved):
Nicely done

GuyAv46 (approved):
Even better

DvirDukhan (approved):

PR #486, opened 2024-07-07
Enhance bm files download and add dispatcher to bm flow

meiravgri:
1. Add workflow dispatcher that allows triggering the benchmarks manually
2. parallelize benchmark files download. Downloading all the benchmarks files takes ~5 min instead of more than 20 min.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -10,12 +10,23 @@ on:
         type: choice
         options:
           - benchmarks-all
+          - benchmarks-default
           - bm-basics-fp32-single
           - bm-basics-fp32-multi
+          - bm-basics-fp64-single
+          - bm-basics-fp64-multi
           - bm-basics-bf16-single
           - bm-basics-bf16-multi
           - bm-basics-fp16-single
           - bm-basics-fp16-multi
+          - bm-batch-iter-fp32-single
+          - bm-batch-iter-fp32-multi
+          - bm-batch-iter-fp64-single
+          - bm-batch-iter-fp64-multi
+          - bm-batch-iter-bf16-single
+          - bm-batch-iter-bf16-multi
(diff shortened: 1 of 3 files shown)


.github/workflows/benchmark.yml
```diff
@@ -10,23 +10,12 @@ on:
...
-          - bm-batch-iter-fp32-single
-          - bm-batch-iter-fp32-multi
-          - bm-batch-iter-fp64-single
-          - bm-batch-iter-fp64-multi
-          - bm-batch-iter-bf16-single
-          - bm-batch-iter-bf16-multi
-          - bm-batch-iter-fp16-single
-          - bm-batch-iter-fp16-multi
           - bm-updated-fp32-single
           - bm-spaces
```
GuyAv46, line 20: why not include the full list?

meiravgri: I read that it is limited to 10 but then read again and they are talking about the input list

GuyAv46: Really? Good to know

tests/benchmark/bm_files.sh
```diff
@@ -32,4 +32,4 @@ elif [ "$BM_TYPE" = "bm-updated-fp32-single" ]; then
     file_name="updated"
 fi
 
-wget --no-check-certificate -q -i tests/benchmark/data/hnsw_indices/hnsw_indices_$file_name.txt -P tests/benchmark/data
```
GuyAv46, line 35: Intentionally removing the `-q`?

meiravgri: Yes, it doesn't affect the time and might be useful if something is wrong

GuyAv46 (dismissed):

GuyAv46 (approved):

PR #487, opened 2024-07-07
Fix nightly, backport and merge-queue

GuyAv46:

Fixed unit-test flow for OSs that don't support node20
Added slack notification on nightly failure
Update backport flow (aligned with RediSearch)

**Mark if applicable**

### DIFF
--- a/.github/workflows/task-unit-test.yml
+++ b/.github/workflows/task-unit-test.yml
@@ -56,15 +56,21 @@ jobs:
       - name: valgrind
         if: ${{ inputs.run-valgrind }}
         run: make valgrind
+      - name: Set Artifact Names
+        # Artifact names have to be unique, so we base them on the environment.
+        # We also remove invalid characters from the name.
+        id: artifact-names
+        run: | # Invalid characters include: Double quote ", Colon :, Less than <, Greater than >, Vertical bar |, Asterisk *, Question mark ?
+          echo "name=$(echo "${{ inputs.container || inputs.env }} ${{ runner.arch }}" | sed -e 's/[":\/\\<>\|*?]/_/g')" >> $GITHUB_OUTPUT
       - name: Archive valgrind tests reports
-        if: ${{ inputs.run-valgrind }} && failure() && steps.node20.outputs.supported == 'true'
+        if: ${{ inputs.run-valgrind && failure() && steps.node20.outputs.supported == 'true' }}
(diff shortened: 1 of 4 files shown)


.github/workflows/task-unit-test.yml
```diff
@@ -30,7 +30,24 @@ jobs:
       - name: pre-checkout script
         if: ${{ inputs.pre-checkout-script }}
         run: ${{ inputs.pre-checkout-script }}
+      - name: Check for node20 support
+        id: node20
+        run: |
+          for os in amazonlinux:2 ubuntu:bionic; do
```
meiravgri, line 36: is there a more general way to test support?
something similar to `which node20`?

.github/workflows/task-unit-test.yml
```diff
@@ -40,8 +57,14 @@ jobs:
...
-        if: ${{ inputs.run-valgrind }} && failure()
-        uses: actions/upload-artifact@v3
+        if: ${{ inputs.run-valgrind }} && failure() && steps.node20.outputs.supported == 'true'
+        uses: actions/upload-artifact@v4
+        with:
+          name: valgrind tests reports
+          path: bin/Linux-x86_64-debug/unit_tests/Testing/Temporary/
+      - name: Archive valgrind tests reports (fallback)
+        if: ${{ inputs.run-valgrind }} && failure() && steps.node20.outputs.supported == 'false'
+        uses: actions/upload-artifact@v4
```
meiravgri, line 67: 3?

.github/workflows/task-unit-test.yml
GuyAv46: I think this is the simplest way (as we do in RediSearch). The node20 comes from the runner but fails in runtime due to a missing and unsupported glibc version, so I don't think `which node20` will fail/return nothing

.github/workflows/task-unit-test.yml
meiravgri: maybe trigger the temp flow?

.github/workflows/task-unit-test.yml
GuyAv46: This is a temporary solution regardless. We hope to drop the support for these OSs soon. When we do, I hope it is clear that this step and all the ones that need its conclusion should be removed

meiravgri (dismissed):

meiravgri (approved):

PR #494, opened 2024-07-08
fix flow tests

meiravgri:
## Poetry
use ml_dtypes for bfloat
update numpy to latest

 ## bindings/cmake:
 * add min version
 * make VecSim library PRIVATE

## Makefile
 * add verbose option to make flow_test

## Installations
* install cmake 3.29

**Mark if applicable**


DvirDukhan (approved):

src/python_bindings/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
+cmake_minimum_required(VERSION 3.25)
```
DvirDukhan, line 1: do we need to change also in RediSearch?

meiravgri: not used by RediSearch

DvirDukhan (approved):

PR #512, opened 2024-07-30
Introduce raw data container interface and data blocks implementation + use it in BF [MOD-7461]

alonre24:

According to the index refactor as discussed - creating the generic interface for raw data `key:value` container that allows the basic CRUD operations over data items.

Create an implementation for this interface that uses a vector of data blocks, and uses it in BF index through the new abstraction layer.  

TODO in next PR(s):
- Apply this in HNSW for the raw vectors' data
- Move the container to the base index struct

**Mark if applicable**


src/VecSim/containers/data_blocks_container.h
```diff
@@ -0,0 +1,30 @@
...
+
+#include "data_block.h"
+#include "raw_data_container_interface.h"
+#include "VecSim/memory/vecsim_malloc.h"
+#include "VecSim/utils/vecsim_stl.h"
+
+class DataBlocksContainer : public VecsimBaseObject, public RawDataContainer {
+
+    size_t element_bytes_count;           // Element size in bytes
+    size_t len;                           // Number of items in the container
```
GuyAv46, line 11: consider renaming to `size` or `element_count`

src/VecSim/containers/data_blocks_container.cpp
```diff
@@ -0,0 +1,52 @@
...
+    }
+    blocks.back().addElement(element);
+    len++;
+    return RAW_DATA_CONTAINER_OK;
+}
+
+const char *DataBlocksContainer::getElement(size_t id) const {
+    if (id >= len) {
+        return nullptr;
+    }
```
GuyAv46, line 26: consider using assert instead (can we ever call this with an incorrect id?)

src/VecSim/containers/data_blocks_container.cpp
```diff
@@ -0,0 +1,52 @@
...
+    return blocks.at(id / this->block_size).getElement(id % this->block_size);
+}
+
+RawDataContainer_Status DataBlocksContainer::removeElement(size_t id) {
+    if (id >= len) {
+        return RAW_DATA_CONTAINER_ID_NOT_EXIST;
+    }
+    if (id < len - 1) {
+        return RAW_DATA_CONTAINER_ERR; // only the last element can be removed
```
GuyAv46, line 35: can we add coverage for this? can it ever happen? if not, consider using an assertion

src/VecSim/containers/data_blocks_container.cpp
```diff
@@ -0,0 +1,52 @@
...
+const char *DataBlocksContainer::getElement(size_t id) const {
+    if (id >= len) {
+        return nullptr;
+    }
+    return blocks.at(id / this->block_size).getElement(id % this->block_size);
+}
+
+RawDataContainer_Status DataBlocksContainer::removeElement(size_t id) {
+    if (id >= len) {
+        return RAW_DATA_CONTAINER_ID_NOT_EXIST;
```
GuyAv46, line 32: can this ever happen? if not, consider using an assertion

src/VecSim/containers/data_blocks_container.cpp
```diff
@@ -0,0 +1,52 @@
...
+      len(0), blocks(allocator), block_size(blockSize), alignment(_alignment) {}
+
+DataBlocksContainer::~DataBlocksContainer() = default;
+
+size_t DataBlocksContainer::size() const { return len; }
+
+RawDataContainer_Status DataBlocksContainer::appendElement(const void *element) {
+    if (len % block_size == 0) {
+        blocks.push_back(DataBlock(this->block_size, this->element_bytes_count, this->allocator,
+                                   this->alignment));
```
GuyAv46, line 16: consider using `emplace_back` instead

src/VecSim/algorithms/brute_force/brute_force.h
```diff
@@ -96,50 +100,37 @@ class BruteForceIndex : public VecSimIndexAbstract<DataType, DistType> {
...
         resizeLabelLookup(idToLabelMapping.size());
     }
 
-    inline void shrinkByBlock() {
+    void shrinkByBlock() {
         assert(indexCapacity() > 0); // should not be called when index is empty
 
-        // remove last block (should be empty)
-        assert(vectorBlocks.size() > 0 && vectorBlocks.back().getLength() == 0);
-        vectorBlocks.pop_back();
```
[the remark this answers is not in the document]
alonre24: No, this is part of the internal container to decide when to grow/shrink

src/VecSim/algorithms/brute_force/brute_force.h
```diff
@@ -96,50 +100,37 @@ class BruteForceIndex : public VecSimIndexAbstract<DataType, DistType> {
     // Private internal function that implements generic single vector deletion.
     virtual void removeVector(idType id);
 
-    inline void growByBlock() {
-        assert(vectorBlocks.size() == 0 || vectorBlocks.back().getLength() == this->blockSize);
-        vectorBlocks.emplace_back(this->blockSize, this->dataSize, this->allocator,
-                                  this->alignment);
```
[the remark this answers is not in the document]
alonre24: No, this is part of the internal container to decide when to grow/shrink

src/VecSim/algorithms/brute_force/brute_force.h
```diff
@@ -22,41 +24,43 @@
...
     idType count;
 
 public:
     BruteForceIndex(const BFParams *params, const AbstractIndexInitParams &abstractInitParams);
 
     size_t indexSize() const override;
     size_t indexCapacity() const override;
-    vecsim_stl::vector<DistType> computeBlockScores(const DataBlock &block, const void *queryBlob,
-                                                    void *timeoutCtx,
+    vecsim_stl::vector<DistType> computeBlockScores(idType start_id, idType end_id,
```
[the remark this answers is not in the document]
alonre24: Basically no. Using an iterator for the `RawDataContainer` is probably a more robust solution. Not sure that in the scope of this PR

src/VecSim/containers/data_block.cpp
```diff
@@ -5,7 +5,7 @@
  */
 
 #include "data_block.h"
-#include "VecSim/memory/vecsim_malloc.h"
+#include "../memory/vecsim_malloc.h"
```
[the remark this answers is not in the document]
alonre24: Redundant, all of these `..` were added when I moved between directories...

src/VecSim/containers/raw_data_container_interface.h
```diff
@@ -0,0 +1,42 @@
...
+     * @return status
+     */
+    virtual RawDataContainer_Status removeElement(size_t id) = 0;
+    /**
+     * @param id to change its asociated data
+     * @param element the new raw data to associate with id
+     * @return status
+     */
+    virtual RawDataContainer_Status updateElement(size_t id, const void *element) = 0;
+};
```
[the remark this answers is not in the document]
alonre24: We havethe generic `removeElement` for that

src/VecSim/containers/raw_data_container_interface.h
```diff
@@ -0,0 +1,42 @@
...
+    RAW_DATA_CONTAINER_ERR
+} RawDataContainer_Status;
+
+struct RawDataContainer {
+    /**
+     * This is an abstract interface, constructor/destructor should be implemented by the derived
+     * classes
+     */
+    RawDataContainer() = default;
+    virtual ~RawDataContainer() = default;
```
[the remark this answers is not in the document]
alonre24: Not mandatory, it is for readability and better practice as I know...
https://www.geeksforgeeks.org/virtual-destructor/
"As a guideline, any time you have a virtual function in a class, you should immediately add a virtual destructor (even if it does nothing). This way, you ensure against any surprises later."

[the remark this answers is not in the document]
GuyAv46: We should make the constructor virtual as well, no?

src/VecSim/containers/data_blocks_container.cpp
```diff
@@ -4,48 +4,40 @@ DataBlocksContainer::DataBlocksContainer(size_t blockSize, size_t elementBytesCo
...
 }
 
 RawDataContainer_Status DataBlocksContainer::removeElement(size_t id) {
-    if (id >= len) {
-        return RAW_DATA_CONTAINER_ID_NOT_EXIST;
-    }
-    if (id < len - 1) {
-        return RAW_DATA_CONTAINER_ERR; // only the last element can be removed
-    }
+    assert(id == element_count - 1); // only the last element can be removed
```
GuyAv46, line 30: Why not implement the remove as an internal swap with the last element?

src/VecSim/containers/raw_data_container_interface.h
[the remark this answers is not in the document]
alonre24: I don't think that there is such a concept of virtual constructor in C++

src/VecSim/containers/data_blocks_container.cpp
alonre24: The ID is an external handle. If we would swap, then you would loose the original mapping (or you would have to maintain another mapping, which is not desired in this case...)

src/VecSim/algorithms/brute_force/bfm_batch_iterator.h
```diff
@@ -19,34 +19,30 @@ class BFM_BatchIterator : public BF_BatchIterator<DataType, DistType> {
...
-            auto block_scores = this->index->computeBlockScores(
-                curr_id, last_id, this->getQueryBlob(), this->getTimeoutCtx(), &rc);
-            if (VecSim_OK != rc) {
-                return rc;
+        auto *vectors_it = this->index->getVectorsIterator();
+        auto DistFunc = this->index->getDistFunc();
+        while (vectors_it->hasNext()) {
+            // Compute the scores for every vector and extend the scores array.
+            if (VECSIM_TIMEOUT(this->getTimeoutCtx())) {
+                return VecSim_QueryReply_TimedOut;
```
GuyAv46, line 34: this is probably OK, but we initially did not check for timeout after every vector to avoid the overhead (only per block). We should at least benchmark to see we don't add a regression here

src/VecSim/algorithms/brute_force/brute_force.h
```diff
@@ -251,33 +234,33 @@ BruteForceIndex<DataType, DistType>::topKQuery(const void *queryBlob, size_t k,
...
+
+    // For vector, compute its scores and update the Top candidates max heap
+    auto *vectors_it = vectors->getIterator();
     idType curr_id = 0;
-    while (curr_id < this->count) {
-        idType last_id = MIN(curr_id + this->blockSize - 1, this->count - 1);
-        auto scores = computeBlockScores(curr_id, last_id, queryBlob, timeoutCtx, &rep->code);
-        if (VecSim_OK != rep->code) {
+    while (vectors_it->hasNext()) {
+        if (VECSIM_TIMEOUT(timeoutCtx)) {
```
GuyAv46, line 243: same comment about regression

src/VecSim/algorithms/brute_force/brute_force.h
```diff
@@ -297,19 +280,18 @@ BruteForceIndex<DataType, DistType>::rangeQuery(const void *queryBlob, double ra
...
 
     DistType radius_ = DistType(radius);
+    auto *vectors_it = vectors->getIterator();
     idType curr_id = 0;
-    while (curr_id < this->count) {
-        idType last_id = MIN(curr_id + this->blockSize - 1, this->count - 1);
-        auto scores = computeBlockScores(curr_id, last_id, queryBlob, timeoutCtx, &rep->code);
-        if (VecSim_OK != rep->code) {
+    while (vectors_it->hasNext()) {
+        if (VECSIM_TIMEOUT(timeoutCtx)) {
```
GuyAv46, line 287: same comment about regression

src/VecSim/containers/raw_data_container_interface.h
```diff
@@ -41,4 +43,24 @@ struct RawDataContainer {
...
+     */
+    RawDataContainer_Iterator() = default;
+    virtual ~RawDataContainer_Iterator() = default;
+
+    /**
+     * The basic iterator operations API
+     */
+    virtual bool hasNext() = 0;
+    virtual const char *next() = 0;
+    virtual void reset() = 0;
```
GuyAv46, line 65: Any reason not to adapt to C++ API (`++` and `*` operators)?
Also consider making this class a sub-class of the main one

src/VecSim/algorithms/brute_force/bfm_batch_iterator.h
alonre24: ARe you sure? We had the timeout check in the compute block score function before, in every iteration:

```    
for (size_t i = 0; i < len; i++) {
        if (VECSIM_TIMEOUT(timeoutCtx)) {
            *rc = VecSim_QueryReply_TimedOut;
            return scores;
        }
         ...
}
```

src/VecSim/containers/raw_data_container_interface.h
alonre24: I moved the iterator to be a sub-class. Regarding the operators, it's nice and sophisticated but doesn't make much difference... we already have iterators in this format which is clearer IMO.

src/VecSim/algorithms/brute_force/bfm_batch_iterator.h
GuyAv46: OK maybe here we did. But for pure KNN queries we didn't

src/VecSim/containers/raw_data_container_interface.h
GuyAv46: I don't think that a C API is clearer on a C++ project. I also don't like that we have both `hasNext` and `next` but `next` calls `hasNext` anyways

GuyAv46 (approved):

src/VecSim/algorithms/brute_force/brute_force.h
```diff
@@ -201,10 +178,7 @@ void BruteForceIndex<DataType, DistType>::removeVector(idType id_to_delete) {
     labelType last_idx_label = getVectorLabel(last_idx);
 
     // Get last vector data.
-    DataBlock &last_vector_block = vectorBlocks.back();
-    assert(&last_vector_block == &getVectorVectorBlock(last_idx));
-
-    void *last_vector_data = last_vector_block.removeAndFetchLastElement();
+    const char *last_vector_data = vectors->getElement(last_idx);
```
meiravgri, line 181: this can be inside the if, no?
or replaced with pop + fetch + remove `vectors->removeElement(last_idx);`

src/VecSim/algorithms/brute_force/brute_force_single.h
```diff
@@ -52,49 +52,36 @@ class BruteForceIndex_Single : public BruteForceIndex<DataType, DistType> {
 #endif
 protected:
     // inline definitions
+    void setVectorId(labelType label, idType id) override { labelToIdLookup.emplace(label, id); }
```
meiravgri, line 55: no required changes in brute_force_multi?

tests/unit/test_bruteforce.cpp
```diff
@@ -344,7 +344,7 @@ TYPED_TEST(BruteForceTest, brute_force_reindexing_same_vector) {
     ASSERT_EQ(VecSimIndex_IndexSize(index), 0);
 
     // The vector block should be removed.
-    ASSERT_EQ(bf_index->getVectorBlocks().size(), 0);
+    ASSERT_EQ(bf_index->getVectorsContainer()->size(), 0);
```
meiravgri, line 347: can't comment in the relevant location.
I have verfied that all the friend declaration for the tests are still needed and `brute_force_reindexing_same_vector_different_id` is redundant, probably not related to this PR but can be fixed on the way :)

meiravgri (commented):
Cool! Well done

Some suggestions:
Consider running benchmarks to ensure no or minimal perforamnce hit.

Also, now that we are not initailizng any block, can we change `BruteForceTest/testSizeEstimation` to compare **equality** between estimation and actual?

src/VecSim/containers/data_block.h
```diff
@@ -40,15 +38,20 @@ struct DataBlock : public VecsimBaseObject {
...
 
-    inline char *removeAndFetchLastElement() {
+    void popLastElement() {
+        assert(this->length > 0);
+        this->length--;
+    }
+
+    char *removeAndFetchLastElement() {
         return this->data + ((--this->length) * element_bytes_count);
     }
```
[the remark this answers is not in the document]
alonre24: `removeAndFetchLastElement` should be removed completely once we use raw_data_layer in HNSW as well - the generic API will use for fetching the last element with `getElement` and then we remove it (as we did in flat now)

src/VecSim/algorithms/brute_force/brute_force.h
```diff
@@ -6,57 +6,56 @@
...
+#include <sys/param.h>
 
 using spaces::dist_func_t;
 
 template <typename DataType, typename DistType>
 class BruteForceIndex : public VecSimIndexAbstract<DataType, DistType> {
 protected:
     vecsim_stl::vector<labelType> idToLabelMapping;
-    vecsim_stl::vector<DataBlock> vectorBlocks;
+    RawDataContainer *vectors;
```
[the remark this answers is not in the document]
alonre24: This should move to the base index object in the next PR, in which we cannot know the dynamic conatiner object (so it is a pointer here already as a preparation)

src/VecSim/algorithms/brute_force/brute_force.h
```diff
@@ -154,12 +140,13 @@ template <typename DataType, typename DistType>
 BruteForceIndex<DataType, DistType>::BruteForceIndex(
     const BFParams *params, const AbstractIndexInitParams &abstractInitParams)
     : VecSimIndexAbstract<DataType, DistType>(abstractInitParams),
-      idToLabelMapping(this->allocator), vectorBlocks(this->allocator), count(0) {
+      idToLabelMapping(this->allocator), count(0) {
     assert(VecSimType_sizeof(this->vecType) == sizeof(DataType));
     // Round up the initial capacity to the nearest multiple of the block size.
     size_t initialCapacity = RoundUpInitialCapacity(params->initialCapacity, this->blockSize);
     this->idToLabelMapping.resize(initialCapacity);
-    this->vectorBlocks.reserve(initialCapacity / this->blockSize);
```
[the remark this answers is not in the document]
alonre24: yes, since we decided to deprecate this argument anyway (we see it as an internal implementation decision)

src/VecSim/algorithms/brute_force/brute_force_single.h
alonre24: No, as we do not update vectors in multi once we get a new vector for some existing label, we append it instead

src/VecSim/containers/data_blocks_container.cpp
```diff
@@ -0,0 +1,80 @@
...
+RawDataContainer::Status DataBlocksContainer::addElement(const void *element, size_t id) {
+    assert(id == element_count); // we can only append new elements
+    if (element_count % block_size == 0) {
+        blocks.emplace_back(this->block_size, this->element_bytes_count, this->allocator,
+                            this->alignment);
+    }
+    blocks.back().addElement(element);
+    element_count++;
+    return OK;
+}
```
meiravgri, line 26: Please use explicit scoping
```suggestion
    return Status::OK;
```
A child class might include an enum with the same enumaration (`OK`)
The easiest way to identify all the locations that require explicit scoping is to add 
`enum Status2 { OK = 0, ID_ALREADY_EXIST, ID_NOT_EXIST, ERR };` to `DataBlocksContainer` and see where the comilation fails

src/VecSim/containers/raw_data_container_interface.h
```diff
@@ -0,0 +1,59 @@
+#pragma once
+
+struct RawDataContainer {
+    enum Status { OK = 0, ID_ALREADY_EXIST, ID_NOT_EXIST, ERR };
```
meiravgri, line 4: It is a better practice to use scoped enumaration (`enum class`) that doesn't allow implicit conversion to integral types.

tests/unit/test_common.cpp
```diff
@@ -383,6 +383,26 @@ TYPED_TEST(UtilsTests, results_containers) {
     VecSimQueryReply_Free(res2);
 }
 
+TYPED_TEST(UtilsTests, data_blocks_container) {
+    std::shared_ptr<VecSimAllocator> allocator = VecSimAllocator::newVecsimAllocator();
+    // Create a simple data blocks container of chars with block of size 1.
+    auto chars_container = DataBlocksContainer(1, 1, allocator, 64);
+    ASSERT_EQ(chars_container.size(), 0);
+    ASSERT_EQ(chars_container.addElement(std::string("a").c_str(), 0), RawDataContainer::OK);
+    ASSERT_EQ(chars_container.size(), 1);
```
meiravgri, line 392: using `enum class` requires using `RawDataContainer::Status::OK` here

meiravgri (commented):
some small comments regarding the enum

meiravgri (dismissed):

GuyAv46 (approved):

PR #517, opened 2024-08-07
Use pull request target in gha

alonre24:

Replace pull_request with pull_request_target to enable forked PRs to run our basic flow which includes creating EC2 instance for  coverage flow (forked PRs still required maintainer approval to run)
https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target

**Mark if applicable**

### DIFF
--- a/.github/workflows/benchmark-labeled.yml
+++ b/.github/workflows/benchmark-labeled.yml
@@ -1,7 +1,7 @@
 name: benchmark
 
 on:
-  pull_request:
+  pull_request_target:
     types: [opened, reopened, synchronize, labeled] # Default ([opened, reopened, synchronize]) + labeled
     branches-ignore:
       - '*docs'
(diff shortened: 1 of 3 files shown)


GuyAv46 (commented):
Nice
Shouldn't we fix the benchmark flow as well?

alonre24:
> Nice Shouldn't we fix the benchmark flow as well? Also, the CI doesn't seem to run. Can we verify that it's because the target branch doesn't have the `pull_request_target` yet?

Will update the benchmark 
I verified - this is the reason why CI didn't run

GuyAv46 (approved):
LGTM
Requires force push right?

alonre24:
> LGTM Requires force push right?

yes...

PR #522, opened 2024-08-12
[MOD-7369] Add AVX512VL flag to FP16 flag to enhance performance.

meiravgri:
### Background
In PR https://github.com/RedisAI/VectorSimilarity/pull/477 we introduced distance functions leveraging `AVX512FP16` intrinsic to enhance `FLOAT16` distance calculations.
We observed that `_mm512_reduce_add_ph` incurs significant overhead, leading to a substantial degradation in distance calculation performance compared to alternative implementations. However, this overhead becomes less significant at higher dimensions. Therefore, `AVX512FP16` instruction set was enabled only for dimensions exceeding a certain threshold.

### Current PR
This PR adds the `-mavx512vl` compilation flag to mitigate the latency issues observed with `_mm512_reduce_add_ph`.
Explanation: 
The reduce instruction is a complex instruction,  translating into a sequence of instructions under the hood. 
**With the flag**, the compiler uses instructions: 128/256-bit `vaddph`. **Without the flag**, the compiler uses instructions: `vaddsh`. 
As demonstrated in the attached performance graph, the first option leads to improved performance across all dimensions.
image
image

#### Additional changes
Expand LOW_DIM spaces benchmarks range. Starts from dim = 55 instead of 100. Still satisfies a no-residual dimension for all types (at dim = 160)


src/VecSim/spaces/functions/AVX512FP16_VL.cpp
GuyAv46, line 1: How is this file and its functions completely new? Shouldn’t we have other functions to rename or remove?

meiravgri: I think i fixed it

GuyAv46 (approved):
Awesome

alonre24 (approved):

PR #525, opened 2024-08-14
[MOD-7474] fix valgrind on linux x86

meiravgri:
## Better resources management in Valgrind
This update addresses an issue where Valgrind hangs when running `HNSWTestParalle`l on x86 machines. The exact cause is unclear but appears related to Valgrind's thread management. By setting the `--fair-sched=yes` flag, Valgrind uses a different algorithm to synchronize threads, potentially avoiding the hang.

## Modifying Parallel Tests
- Added a test destructor to free the index to avoid false leak alerts.
- Print neighbors in case the search fails to identify unreachable nodes.
- Use `testing::Test::HasFatalFailure()` to halt tests early if they fail within a subroutine or thread.
- Modified `parallelInsertSearch` to continue running queries until 95% of vectors are indexed, rather than just a single query during indexing.
- Split the previous loop running on` is_multi = [false, true]` into two distinct tests—one for single index and one for multi-index scenarios.

### DIFF
--- a/tests/unit/test_hnsw_parallel.cpp
+++ b/tests/unit/test_hnsw_parallel.cpp
@@ -369,19 +369,28 @@ TYPED_TEST(HNSWTestParallel, parallelInsert) {
         // label and the query val (n/2, n/2-1, n/2+1, n/2-2, n/2+2, ...) The score is the L2
         // distance between the vectors that correspond the ids.
         size_t diff_id = std::abs(int(id - n / 2));
-        if (diff_id != (res_index + 1) / 2) {
+        if (diff_id != (res_index + 1 + 7) / 2) {
+            ADD_FAILURE() << "Expected diff_id: " << (res_index + 1 + 7) / 2 << " got: " << diff_id;
             auto *hnsw_index = this->CastToHNSW(parallel_index);
             int **neighbors_output;
             VecSimDebug_GetElementNeighborsInHNSWGraph(hnsw_index, id, &neighbors_output);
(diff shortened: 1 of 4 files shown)


tests/unit/test_hnsw_parallel.cpp
```diff
@@ -435,82 +484,90 @@ TYPED_TEST(HNSWTestParallel, parallelInsertSearch) {
...
                 // We expect to get the results with increasing order of the distance between the
                 // res label and the query val (n/4, n/4-1, n/4+1, n/4-2, n/4+2, ...) The score is
                 // the L2 distance between the vectors that correspond the ids.
                 size_t diff_id = std::abs(int(id - query_val));
-                ASSERT_EQ(diff_id, (res_index + 1) / 2);
-                ASSERT_EQ(score, (dim * (diff_id * diff_id)));
+                if (is_multi) {
+                    ASSERT_EQ(diff_id, (res_index + 1) / 2);
+                    ASSERT_EQ(score, (dim * (diff_id * diff_id)));
+                    return;
```
alonre24, line 529: Why different behavior for multi?

alonre24 (commented):
Nice
What about enabling valgrind in alpine?

tests/unit/test_hnsw_parallel.cpp
meiravgri: from `VecSimDebug_GetElementNeighborsInHNSWGraph` description:

> TODO: Implement the full version that supports MULTI as well.

alonre24 (approved):

PR #534, opened 2024-08-27
[MOD-7643] Handle pending jobs after delete in-place in tiered index [Bug]

alonre24:

In tiered HNSW, we support insert/delete vectors asynchronously and in place (while holding the global index lock exclusively). However, these modes might be combined, and in case we move from deleting vectors asynchronously to deleting vectors in place, we may have some pending repair jobs and swap jobs for the deleted vector and the id of the last vector, which is going to take the deleted vector's id after the deletion.

Until now, we have failed to handle these pending jobs and remained with an undefined behavior. We fix this mechanism and invalidate the associated jobs properly, as we do for asynchronous deletion (with the "swap job" routine).

Also, we cannot assume that upon deletion in-place we can always remove the last id's label in `SwapLastIdWithDeletedId` since it might just be marked as deleted by asynchronous delete before (hence, not existing).
Moreover, avoid using the `has_marked_deleted` flag in HNSW functions, as it is not safe - elements can be marked as deleted anytime during the graph searches (which do not hold the global index data lock).

**Mark if applicable**


src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -800,7 +825,7 @@ int TieredHNSWIndex<DataType, DistType>::deleteVector(labelType label) {
     } else {
         // delete in place.
         this->mainIndexGuard.lock();
-        num_deleted_vectors += this->backendIndex->deleteVector(label);
+        num_deleted_vectors += this->deleteLabelInplace(label);
```
meiravgri, line 828: what lock protects the write mode?

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -3535,3 +3535,64 @@ TYPED_TEST(HNSWTieredIndexTestBasic, FitMemoryTest) {
...
+                         .multi = isMulti};
+    VecSimParams hnsw_params = CreateParams(params);
+
+    auto mock_thread_pool = tieredIndexMock();
+
+    auto *tiered_index = this->CreateTieredHNSWIndex(hnsw_params, mock_thread_pool);
+    auto allocator = tiered_index->getAllocator();
+
+    // Insert one vector to HNSW.
+    GenerateAndAddVector<TEST_DATA_T>(tiered_index->backendIndex, dim, 0);
```
meiravgri, line 3556: add multiple vectors per label if multi

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -486,6 +488,25 @@ idType TieredHNSWIndex<DataType, DistType>::setAndSaveInvalidJob(AsyncJob *job)
...
 }
 
+template <typename DataType, typename DistType>
+int TieredHNSWIndex<DataType, DistType>::deleteLabelInplace(labelType label) {
+    auto ids = this->getHNSWIndex()->getElementIds(label);
+    int num_deleted_vectors = this->backendIndex->deleteVector(label);
+    // dispoase pending repair and swap jobs for the removed ids.
+    vecsim_stl::vector<idType> idsToRemove(this->allocator);
+    readySwapJobs += ids.size(); // account for the current ids that are going to be removed.
+    for (auto id : ids) {
```
meiravgri, line 498: should executeSwapJob be dne i a seperate loop?

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -3535,3 +3535,64 @@ TYPED_TEST(HNSWTieredIndexTestBasic, FitMemoryTest) {
...
+    ASSERT_EQ(((HNSWRepairJob *)(mock_thread_pool.jobQ.front().job))->node_id, 0);
+    ASSERT_EQ(((HNSWRepairJob *)(mock_thread_pool.jobQ.front().job))->level, 0);
+    ASSERT_EQ(tiered_index->idToRepairJobs.size(), 1);
+    ASSERT_GE(tiered_index->idToRepairJobs.at(0).size(), 1);
+    ASSERT_EQ(tiered_index->idToRepairJobs.at(0)[0]->associatedSwapJobs.size(), 1);
+    ASSERT_EQ(tiered_index->idToRepairJobs.at(0)[0]->associatedSwapJobs[0]->deleted_id, 1);
+
+    // Add one more vector and remove it, expect that the same repair job for 0 would be created
+    // for reapring 0->2.
+    GenerateAndAddVector<TEST_DATA_T>(tiered_index->backendIndex, dim, 2, 2);
```
meiravgri, line 3575:
```suggestion
    // for repairing 0->2.
```

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -126,6 +126,9 @@ class TieredHNSWIndex : public VecSimTieredIndex<DataType, DistType> {
     // Returns the id that the job was stored under (to be set in the job id field).
     idType setAndSaveInvalidJob(AsyncJob *job);
 
+    // Handle deletion of vector inplace condering that async deletion might occurred beforhand.
```
meiravgri, line 129: ?
```suggestion
    // Handle deletion of vector inplace considering that async deletion might occurred beforehand.
```

meiravgri (commented):
Good catch!
1. Correct me if i'm wrng but `has_marked_deleted` can never be asumed? maybe all the `has_marked_deleted` templated functions should be reviewed? 
I noticed that in some places it is uses to differntiate between tiered hnsw and a simple hnsw. If this is the only job of this template, cosider renaming.

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -674,10 +695,14 @@ int TieredHNSWIndex<DataType, DistType>::addVector(const void *blob, labelType l
     int ret = 1;
     auto hnsw_index = this->getHNSWIndex();
     if (this->getWriteMode() == VecSim_WriteInPlace) {
+        // First, check if we need to overwrite the vector in-place for single (from both indexes).
+        if (!this->backendIndex->isMultiValue()) {
+            ret -= this->deleteVector(label);
+        }
+        // Insert the vector to the HNSW index. Internally, we will never have to overrite the label
+        // since we already checked it outside.
```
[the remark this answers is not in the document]
alonre24: As discussed - we are checking it here (line 700), and it is required to handle deletion in place in case of overwrites (instead of letting HNSW handle the overwrite without invalidating the jobs)

tests/unit/test_hnsw_tiered.cpp
alonre24: what?

src/VecSim/algorithms/hnsw/hnsw_tiered.h
alonre24: Sorry, I sat on my keyboard I guess ;)
The suggestion is right, is there anything else unclear?

tests/unit/test_hnsw_tiered.cpp
meiravgri: 😓 got messed up
just fixed the Spelling of repairing

src/VecSim/algorithms/hnsw/hnsw_tiered.h
meiravgri: just wasn't sure you meant "considering"

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -486,6 +488,25 @@ idType TieredHNSWIndex<DataType, DistType>::setAndSaveInvalidJob(AsyncJob *job)
...
+int TieredHNSWIndex<DataType, DistType>::deleteLabelInplace(labelType label) {
+    auto ids = this->getHNSWIndex()->getElementIds(label);
+    int num_deleted_vectors = this->backendIndex->deleteVector(label);
+    // dispoase pending repair and swap jobs for the removed ids.
+    vecsim_stl::vector<idType> idsToRemove(this->allocator);
+    readySwapJobs += ids.size(); // account for the current ids that are going to be removed.
+    for (auto id : ids) {
+        auto swap_job = HNSWSwapJob(this->allocator, id);
+        idToSwapJob[id] = &swap_job;
+        this->executeSwapJob(&swap_job, idsToRemove);
```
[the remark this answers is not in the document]
alonre24: I let `executeSwapJob` recieve the deleted id and not the job so we do not need to create the temp swap job

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -486,6 +494,26 @@ idType TieredHNSWIndex<DataType, DistType>::setAndSaveInvalidJob(AsyncJob *job)
     return curInvalidId;
 }
 
+template <typename DataType, typename DistType>
+int TieredHNSWIndex<DataType, DistType>::deleteLabelInplace(labelType label) {
```
meiravgri, line 498: consider renaming to `deleteLabelFromHNSWInplace`

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -3535,3 +3535,123 @@ TYPED_TEST(HNSWTieredIndexTestBasic, FitMemoryTest) {
...
+    tiered_index->setWriteMode(VecSim_WriteInPlace);
+    // Delete inplace, expect that the repair job for 0->1 and 0->2 will not be valid anymore.
+    ASSERT_EQ(tiered_index->deleteVector(0), 2);
+    ASSERT_EQ(tiered_index->indexSize(), 3);
+    ASSERT_FALSE(mock_thread_pool.jobQ.front().job->isValid);
+    mock_thread_pool.thread_iteration();
+    // the next job in the queue shoule be the repair job for 2->3 and invalidate as well.
+    ASSERT_FALSE(mock_thread_pool.jobQ.front().job->isValid);
+    mock_thread_pool.thread_iteration();
+    // the next job in the queue shoule be the repair job for 4->3, we validate that the id of 4
```
meiravgri, line 3645: capital letter

meiravgri (commented):
cool :)

src/VecSim/algorithms/hnsw/hnsw.h
```diff
@@ -1973,13 +1958,8 @@ VecSimQueryReply *HNSWIndex<DataType, DistType>::topKQuery(const void *query_dat
...
     // We now oun the results heap, we need to free (delete) it when we done
     candidatesLabelsMaxHeap<DistType> *results;
-    if (this->numMarkedDeleted) {
-        results = searchBottomLayer_WithTimeout<true>(
-            bottom_layer_ep, query_data, std::max(query_ef, k), k, timeoutCtx, &rep->code);
-    } else {
-        results = searchBottomLayer_WithTimeout<false>(
-            bottom_layer_ep, query_data, std::max(query_ef, k), k, timeoutCtx, &rep->code);
-    }
+    results = searchBottomLayer_WithTimeout(bottom_layer_ep, query_data, std::max(query_ef, k), k,
```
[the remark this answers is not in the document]
alonre24: I believe it was unsafe to do it to begin with, since the number of marked deleted might change anytime during the graph search... @GuyAv46 WDYT?

src/VecSim/vec_sim.h
```diff
@@ -235,6 +235,8 @@ void VecSim_SetLogCallbackFunction(logCallbackFunction callback);
 /**
  * @brief Allow 3rd party to set the write mode for tiered index - async insert/delete using
  * background jobs, or insert/delete inplace.
+ * !! In tiered index scenario, Should we called from main thread only !! (that is, the thread that
```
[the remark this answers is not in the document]
alonre24: Its implicit, wanted to give more info :)

src/VecSim/algorithms/hnsw/hnsw.h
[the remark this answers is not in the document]
GuyAv46: It could be changed with the read lock right? So yeah I think it wasn't safe. Can we see the performance hit here?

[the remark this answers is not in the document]
alonre24: Running micro-benchmarks - let's see

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -255,44 +258,48 @@ void TieredHNSWIndex<DataType, DistType>::executeRepairJobWrapper(AsyncJob *job)
...
-        idToSwapJob.at(prev_last_id)->deleted_id = job->deleted_id;
-        idToSwapJob.at(job->deleted_id) = idToSwapJob.at(prev_last_id);
+        idToSwapJob.at(prev_last_id)->deleted_id = deleted_id;
+        if (idToSwapJob.contains(deleted_id)) {
+            idToSwapJob.at(deleted_id) = idToSwapJob.at(prev_last_id);
+        } else {
+            // Id was deleted in-place so there was no swap job for it, create a new one for the
+            // swapped id
+            idToSwapJob.insert({deleted_id, idToSwapJob.at(prev_last_id)});
+        }
```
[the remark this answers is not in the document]
alonre24: I thought that we cannot do `[]` with `vecsim_stl` map, but it seems to work :)

GuyAv46 (approved):

PR #572, opened 2024-12-17
Override missing intrinsincs in gcc <11

meiravgri:
There are missing implmnetation for several intrinsincs in gcc < 11 

full list and suggested alternatives for each missing function can be found here:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95483

The missing functions were added in gcc11.

in this PR we override `_mm256_loadu_epi8` with `mm256_maskz_loadu_epi8`, using ~0 mask as recommended by the gcc team.

### DIFF
--- a/src/VecSim/spaces/space_includes.h
+++ b/src/VecSim/spaces/space_includes.h
@@ -16,6 +16,9 @@
 #if defined(__AVX512F__) || defined(__AVX__) || defined(__SSE__)
 #if defined(__GNUC__)
 #include <x86intrin.h>
+// Override missing implementations in GCC < 11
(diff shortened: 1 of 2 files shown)


alonre24 (approved):

src/VecSim/spaces/space_includes.h
```diff
@@ -16,6 +16,9 @@
 #if defined(__AVX512F__) || defined(__AVX__) || defined(__SSE__)
 #if defined(__GNUC__)
 #include <x86intrin.h>
+#if (__GNUC__ < 11)
```
alonre24, line 19: Add an explanation  or the link to the issue in a comment here

PR #573, opened 2024-12-19
[MOD-8206] INT8 flow tests

meiravgri:
This pull request introduces flow tests for `VecSimType_INT8` data type.

### Generalized Test Classes
* Implemented `GeneralTest` class for running common tests on BF and HNSW indices. Currently only used by `TestINT8`.
* Created `TestINT8` class, inherits from `GeneralTest`

### Bindings support
* Added support for `VecSimType_INT8` in the `PyHNSWLibIndex`:
  * `HNSWIndex::saveIndex`
  * `HNSWIndex::checkIntegrity`

### New Helper Functions
* Introduced `create_flat_index` and `create_add_vectors` functions to facilitate the creation and addition of vectors to an index.
* Added a function `create_int8_vectors` to generate INT8 vectors for testing purposes.
* `fp32_expand_and_calc_cosine_dist` takes 2 numpy arrays and converts them to `np.float32` to avoid overflow in distance calculations. 

### Additional Improvements
* Updated `IndexCtx` class to include `type_to_dtype` mapping for various data types to their numpy type.
* Modified `__init__` method in `IndexCtx` to accept a `create_data_func` parameter, allowing custom data creation functions to be passed.

### DIFF
--- a/tests/flow/common.py
+++ b/tests/flow/common.py
@@ -24,6 +24,7 @@ def create_hnsw_params(dim, num_elements, metric, data_type, ef_construction=200
     hnsw_params.multi = is_multi
 
     return hnsw_params
+
 # Helper function for creating an index,uses the default HNSW parameters if not specified.
 def create_hnsw_index(dim, num_elements, metric, data_type, ef_construction=200, m=16, ef_runtime=10, epsilon=0.01,
                       is_multi=False):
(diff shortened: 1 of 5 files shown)


tests/flow/common.py
```diff
@@ -40,6 +41,23 @@ def create_hnsw_index(dim, num_elements, metric, data_type, ef_construction=200,
...
+    bfparams = BFParams()
+
+    bfparams.dim = dim
+    bfparams.type = data_type
+    bfparams.metric = metric
+    bfparams.multi = is_multi
+
+    return BFIndex(bfparams)
+
+def create_add_vectors(index, vectors):
```
GuyAv46, line 55: why is it called "create"?

GuyAv46 (approved):
@alonre24 We should refactor our flow tests, to be more generic, and also remove irrelevant tests (or move to a `test_bindings.py` file if all they do is to check the bindings)

tests/flow/common.py
meiravgri: It creates also the list of (key, vector) tuples

PR #583, opened 2024-12-25
[MOD-8204] int8 micro benchmarks

lerman25:
This pull request introduces support for INT8 data type in benchmarks. The changes include adding new benchmark types, modifying existing scripts to handle the new data type INT8 vectors.


tests/benchmark/bm_initialization/bm_basics_initialize_int8.h
```diff
@@ -0,0 +1,82 @@
...
+        ->Args({65, 1})                                                                            \
+        ->Args({65, 10})                                                                           \
+        ->Args({65, 100})                                                                          \
+        ->Args({80, 1})                                                                            \
+        ->Args({80, 10})                                                                           \
+        ->Args({80, 100})                                                                          \
+        ->ArgNames({"radiusX100", "epsilonX1000"})                                                 \
+        ->Iterations(10)                                                                           \
+        ->Unit(benchmark::kMillisecond)
+// Memory BF
```
alonre24, line 21: Why use different args for benchmarking range with int8 and not the common ones we use for all other types?

tests/benchmark/bm_files.sh
```diff
@@ -26,6 +26,12 @@ elif [ "$BM_TYPE" = "bm-basics-fp16-single" ] \
 || [ "$BM_TYPE" = "bm-batch-iter-fp16-multi" ]
 then
     file_name="basic_fp16"
+elif [ "$BM_TYPE" = "bm-basics-int8-single" ] \
+|| [ "$BM_TYPE" = "bm-basics-int8-multi" ] \
```
alonre24, line 30: Benchmarks for multi will come in a different PR no?

alonre24 (commented):
Good job! Few small comments. Also, we need to understand why benchmarks flow did not run upon every push to the branch

tests/benchmark/bm_initialization/bm_basics_initialize_int8.h
lerman25: In the dataset I used (wikipedia-int8), the previous range query  resulted in no result (i.e no vectors in X range from query vector), so I used bigger range to fit the current dataset.
The range query results were also validated with a python script.

tests/benchmark/bm_files.sh
lerman25: Yes but it will be needed, so I added it now.

tests/benchmark/bm_initialization/bm_basics_initialize_int8.h
alonre24: I see. Perhaps we can make these arguments our common ones for all types in range search to reduce complexity. WDYT @meiravgri ?

tests/benchmark/bm_files.sh
alonre24: Just for the good order, I would suggest adding it with the PR  that will introduce multi and not now.

tests/benchmark/bm_initialization/bm_basics_initialize_int8.h
meiravgri: IMO We should not touch the current benchmarks values.
We can change the range definition macro to receive additional arguments that will determine the range query parameters.
something like:
```
#define REGISTER_Range(BM_FUNC, radius, epsilons)
->ArgsRange(combine(radiuses, epsilons))
```
I believe google benchmark suggests a suitable API to achieve values combinations.

src/python_bindings/bindings.cpp
```diff
@@ -439,6 +441,7 @@ class PyHNSWLibIndex : public PyVecSimIndex {
             return dynamic_cast<HNSWIndex<int8_t, float> *>(this->index.get())
                 ->checkIntegrity()
                 .valid_state;
+
```
meiravgri, line 444: remove line

tests/benchmark/bm_files.sh
```diff
@@ -26,6 +26,10 @@ elif [ "$BM_TYPE" = "bm-basics-fp16-single" ] \
 || [ "$BM_TYPE" = "bm-batch-iter-fp16-multi" ]
 then
     file_name="basic_fp16"
+elif [ "$BM_TYPE" = "bm-basics-int8-single" ] \
+|| [ "$BM_TYPE" = "bm-batch-iter-int8-single" ] \
```
meiravgri, line 30:
```suggestion
|| [ "$BM_TYPE" = "bm-batch-iter-int8-single" ] 
```

tests/benchmark/benchmarks.sh
```diff
@@ -3,17 +3,20 @@ BM_TYPE=$1;
 if [ -z "$BM_TYPE"  ] || [ "$BM_TYPE" = "benchmarks-all" ]; then
     for bm_class in basics batch_iterator; do
         for type in single multi; do
-            for data_type in fp32 fp64 bf16 fp16; do
+            for data_type in fp32 fp64 bf16 fp16 ; do
```
meiravgri, line 6: revert unnecessary changes please :)

tests/benchmark/bm_initialization/bm_basics_initialize_int8.h
meiravgri: For now a better location for this defintion would be in bm_basics_single_int8.cpp

meiravgri: Why we didn't have to change this for brute force as well? Worth checking if we even get any results there

meiravgri (commented):
Great work :) !
Few little comments and added my shnkel to Alon's concerns.

alonre24 (approved):

tests/benchmark/run_files/bm_basics_single_int8.cpp
```diff
@@ -25,6 +25,32 @@ const char *BM_VecSimGeneral::test_queries_file =
 #define BM_ADD_LABEL_ASYNC          AddLabel_Async_Single
 #define BM_DELETE_LABEL_ASYNC       DeleteLabel_Async_Single
 
+// Larger Range query value are required for wikipedia dataset.
```
meiravgri, line 28:
```suggestion
// Larger Range query values are required for wikipedia dataset.
```

meiravgri (commented):
just a small typo fix and we are good to go

meiravgri (approved):

PR #584, opened 2024-12-26
Implement UINT8 vector type - [MOD-8230, MOD-8408]

GuyAv46:
## Implement support for new type `UINT8`

This PR follows the implementation of the `INT8` type and implements all the functionality required for the `UINT8` type.

**Mark if applicable**

### DIFF
--- a/src/VecSim/spaces/IP/IP.cpp
+++ b/src/VecSim/spaces/IP/IP.cpp
@@ -90,3 +90,28 @@ float INT8_Cosine(const void *pVect1v, const void *pVect2v, size_t dimension) {
         *reinterpret_cast<const float *>(static_cast<const int8_t *>(pVect2v) + dimension);
     return 1.0f - float(INT8_InnerProductImp(pVect1v, pVect2v, dimension)) / (norm_v1 * norm_v2);
 }
+
+static inline int UINT8_InnerProductImp(const void *pVect1v, const void *pVect2v,
+                                        size_t dimension) {
+    uint8_t *pVect1 = (uint8_t *)pVect1v;
+    uint8_t *pVect2 = (uint8_t *)pVect2v;
+
+    int res = 0; // signed int is used to avoid overflow and unsigned arithmetic
+    for (size_t i = 0; i < dimension; i++) {
+        res += pVect1[i] * pVect2[i];
+    }
+    return res;
+}
+
+float UINT8_InnerProduct(const void *pVect1v, const void *pVect2v, size_t dimension) {
+    return 1 - UINT8_InnerProductImp(pVect1v, pVect2v, dimension);
+}
+
+float UINT8_Cosine(const void *pVect1v, const void *pVect2v, size_t dimension) {
+    // We expect the vectors' norm to be stored at the end of the vector.
+    float norm_v1 =
+        *reinterpret_cast<const float *>(static_cast<const uint8_t *>(pVect1v) + dimension);
+    float norm_v2 =
+        *reinterpret_cast<const float *>(static_cast<const uint8_t *>(pVect2v) + dimension);
+    return 1.0f - float(UINT8_InnerProductImp(pVect1v, pVect2v, dimension)) / (norm_v1 * norm_v2);
+}
--- a/src/VecSim/spaces/IP/IP.cpp
+++ b/src/VecSim/spaces/IP/IP.cpp
@@ -67,11 +67,18 @@ float FP16_InnerProduct(const void *pVect1, const void *pVect2, size_t dimension
     return 1.0f - res;
 }
 
-static inline int INT8_InnerProductImp(const void *pVect1v, const void *pVect2v, size_t dimension) {
-    int8_t *pVect1 = (int8_t *)pVect1v;
-    int8_t *pVect2 = (int8_t *)pVect2v;
+// Return type for the inner product functions.
+// The type should be able to hold `dimension * MAX_VAL(int_elem_t) * MAX_VAL(int_elem_t)`.
+// To support dimension up to 2^16, we need the difference between the type and int_elem_t to be at
+// least 2 bytes. We assert that in the implementation.
+template <typename int_elem_t>
+using ret_t = std::conditional_t<sizeof(int_elem_t) == 1, int, long long>;
 
-    int res = 0;
+template <typename int_elem_t>
+static inline ret_t<int_elem_t>
+INTEGER_InnerProductImp(const int_elem_t *pVect1, const int_elem_t *pVect2, size_t dimension) {
+    static_assert(sizeof(ret_t<int_elem_t>) - sizeof(int_elem_t) * 2 >= sizeof(uint16_t));
+    ret_t<int_elem_t> res = 0;
     for (size_t i = 0; i < dimension; i++) {
         res += pVect1[i] * pVect2[i];
     }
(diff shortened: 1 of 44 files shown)


tests/unit/unit_test_utils.cpp
```diff
@@ -46,6 +51,9 @@ VecSimQueryParams CreateQueryParams(const HNSWRuntimeParams &RuntimeParams) {
 
 static bool is_async_index(VecSimIndex *index) {
     return dynamic_cast<VecSimTieredIndex<float, float> *>(index) != nullptr ||
+           dynamic_cast<VecSimTieredIndex<bfloat16, float> *>(index) != nullptr ||
+           dynamic_cast<VecSimTieredIndex<float16, float> *>(index) != nullptr ||
```
alonre24, line 55: There is a ticket for fixing this in all relevant versions (adding the BF16 / FP16). I guess we can have it here, just remember to add it to 0.8 as well

src/VecSim/spaces/L2/L2_AVX512F_BW_VL_VNNI_UINT8.h
```diff
@@ -0,0 +1,78 @@
...
+    sum = _mm512_dpwssd_epi32(sum, diff_lo, diff_lo);
+
+    __m512i va_hi = _mm512_unpackhi_epi8(va, _mm512_setzero_si512()); // AVX512BW
+    __m512i vb_hi = _mm512_unpackhi_epi8(vb, _mm512_setzero_si512());
+    __m512i diff_hi = _mm512_sub_epi16(va_hi, vb_hi);
+    sum = _mm512_dpwssd_epi32(sum, diff_hi, diff_hi);
+
+    // _mm512_dpwssd_epi32(src, a, b)
+    // Multiply groups of 2 adjacent pairs of signed 16-bit integers in `a` with corresponding
+    // 16-bit integers in `b`, producing 2 intermediate signed 32-bit results. Sum these 2 results
```
alonre24, line 28: Is there an unsigned equivalent for `_mm512_dpwssd_epi32`?

src/VecSim/spaces/IP_space.cpp
```diff
@@ -252,4 +252,62 @@ dist_func_t<float> Cosine_INT8_GetDistFunc(size_t dim, unsigned char *alignment,
...
+    auto features = (arch_opt == nullptr)
+                        ? cpu_features::GetX86Info().features
+                        : *static_cast<const cpu_features::X86Features *>(arch_opt);
+#ifdef OPT_AVX512_F_BW_VL_VNNI
+    if (features.avx512f && features.avx512bw && features.avx512vl && features.avx512vnni) {
+        // For uint8 vectors with cosine distance, the extra float for the norm shifts alignment to
+        // `(dim + sizeof(float)) % 32`.
+        // Vectors satisfying this have a residual, causing offset loads during calculation.
+        // To avoid complexity, we skip alignment here, assuming the performance impact is
+        // negligible.
```
alonre24, line 305: I need to better understand the considerations here (I see it follows INT8 as well)

tests/flow/test_bruteforce.py
```diff
@@ -652,7 +652,7 @@ def test_batch_iterator(self):
     def range_query(self, dist_func):
         bfindex = self.create_index(VecSimMetric_Cosine)
         label_to_vec_list = self.create_add_vectors(bfindex)
-        radius = 0.7
+        radius = bfindex.knn_query(self.query_data[0], k=100)[1][0][-1] # get the distance of the 100th closest vector as the radius
```
alonre24, line 655: why?

tests/flow/test_bruteforce.py
```diff
@@ -663,8 +663,8 @@ def range_query(self, dist_func):
         # Verify that we got exactly all vectors within the range
         results, keys = get_ground_truth_results(dist_func, self.query_data[0], label_to_vec_list, res_num)
 
-        assert_allclose(max(bf_distances[0]), results[res_num-1]["dist"], rtol=1e-05)
-        assert np.array_equal(np.array(bf_labels[0]), np.array(keys))
+        assert_allclose(bf_distances[0], [results[i]["dist"] for i in range(res_num)], rtol=1e-05)
+        assert self.compute_correct(bf_labels[0], bf_distances[0], keys, results) == res_num
```
alonre24, line 667: What's the change here exactly?

tests/unit/unit_test_utils.cpp
GuyAv46: Hmm ok. So I’ll remove it here as we won’t BP this PR to 0.8

src/VecSim/spaces/IP_space.cpp
GuyAv46: 👍🏼. I also thought about it and if we put the norm at the beginning of the vector we will be able to use alignment

tests/flow/test_bruteforce.py
GuyAv46: because the radius is related to the data type and data set, and fixing it to a constant value makes no sense (and it also failed as with `UINT8` we found no results).
This way, we get a radius we know matches at least 100 results, regardless of the vectors and their type

tests/flow/test_bruteforce.py
GuyAv46: Check that all the results are matching the expected, and not only that the last one is within the range

GuyAv46: Here now, we also handle ties, where the order of the labels may be different between sorting implementations (we only sort by the distance)

src/VecSim/spaces/L2/L2_AVX512F_BW_VL_VNNI_UINT8.h
GuyAv46: No, but we don't need (or want) one.
We cast the `uint8_t` elements into **signed** `int16_t` (still with zero extension), and then we want to perform signed arithmetic operations.
for example, if we have elements 0 and 200, we want the result to be `(0-200)^2 = -200^2 = 40,000`, and not `(0-200)^2 = (uint16_t)(-200)^2 = 65,336^2 = 4,268,792,896`

tests/unit/unit_test_utils.cpp
GuyAv46: Since we already changed this function on main and 8.0, we will have conflicts with 0.8 anyway. I'll keep this change and we will need to fix it on 0.8 only.

src/VecSim/spaces/normalize/normalize_naive.h
```diff
@@ -74,10 +74,11 @@ static inline void float16_normalizeVector(void *vec, const size_t dim) {
     }
 }
 
-static inline void int8_normalizeVector(void *vec, const size_t dim) {
-    int8_t *input_vector = static_cast<int8_t *>(vec);
+template <typename DataType>
+static inline void int_normalizeVector(void *vec, const size_t dim) {
```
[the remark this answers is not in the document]
GuyAv46: I think it is better to have the name aligned with the previous API set (`float16_normalizeVector`, `bfloat16_normalizeVector`...)

src/python_bindings/bindings.cpp
```diff
@@ -305,6 +305,9 @@ class PyHNSWLibIndex : public PyVecSimIndex {
         } else if (type == VecSimType_INT8) {
             auto *hnsw = dynamic_cast<HNSWIndex<int8_t, float> *>(index.get());
             hnsw->saveIndex(location);
+        } else if (type == VecSimType_UINT8) {
```
meiravgri, line 308: i can't comment at the relevant location :unamused:
why did you remove `VecSimType_INT8` from `getVector(labelType label)` (instead of adding `VecSimType_UINT8`)

tests/benchmark/spaces_benchmarks/bm_spaces_int8.cpp
```diff
@@ -46,11 +46,12 @@ bool avx512_f_bw_vl_vnni_supported = opt.avx512f && opt.avx512bw && opt.avx512vl
 INITIALIZE_BENCHMARKS_SET_L2_IP(BM_VecSimSpaces_Integers_INT8, INT8, AVX512F_BW_VL_VNNI, 32,
                                 avx512_f_bw_vl_vnni_supported);
 INITIALIZE_BENCHMARKS_SET_Cosine(BM_VecSimSpaces_Integers_INT8, INT8, AVX512F_BW_VL_VNNI, 32,
```
meiravgri, line 48: The `;` is redundant no?

tests/benchmark/benchmarks.sh
```diff
@@ -15,8 +15,8 @@ if [ -z "$BM_TYPE"  ] || [ "$BM_TYPE" = "benchmarks-all" ]; then
     echo spaces_fp64
     echo spaces_bf16
     echo spaces_fp16
-    echo spaces_int8    
-
+    echo spaces_int8
+    echo spaces_uint8
```
meiravgri, line 19: let's keep the spaces?

tests/flow/test_bruteforce.py
meiravgri: For better readability, consider splitting this into steps, something like
k100_dists = bfindex.knn_query(self.query_data[0], k=100)[1][0]
last_dist = k100_dists[-1]

you can do better with the names

meiravgri (commented):
cool

tests/benchmark/spaces_benchmarks/bm_spaces_int8.cpp
GuyAv46: I think so, but it messed with the clang-format so I added it

src/VecSim/spaces/IP_space.cpp
alonre24: As discussed - let's open a (small) spike to try that

tests/flow/test_bruteforce.py
GuyAv46: I don't think it makes it more readable

src/VecSim/spaces/IP/IP_AVX512F_BW_VL_VNNI_UINT8.h
```diff
@@ -0,0 +1,91 @@
...
+        pVect1 += residual % 32;
+
+        __m256i temp_b = _mm256_maskz_loadu_epi8(mask, pVect2);
+        __m512i vb = _mm512_cvtepu8_epi16(temp_b);
+        pVect2 += residual % 32;
+
+        sum = _mm512_dpwssd_epi32(sum, va, vb);
+    }
+
+    // TODO: unify this and the above steps for dim>64 with a single mask loading?
```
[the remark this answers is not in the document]
GuyAv46: Done. Benchmarks seems OK

src/VecSim/spaces/IP/IP.cpp
```diff
@@ -67,51 +67,50 @@ float FP16_InnerProduct(const void *pVect1, const void *pVect2, size_t dimension
...
+// To support dimension up to 2^16, we need the difference between the type and int_elem_t to be at
+// least 2 bytes. We assert that in the implementation.
+template <typename int_elem_t>
+using ret_t = std::conditional_t<sizeof(int_elem_t) == 1, int, long long>;
+
+template <typename int_elem_t>
+static inline ret_t<int_elem_t>
+INTEGER_InnerProductImp(const int_elem_t *pVect1, const int_elem_t *pVect2, size_t dimension) {
+    static_assert(sizeof(ret_t<int_elem_t>) - sizeof(int_elem_t) * 2 >= sizeof(uint16_t));
+    ret_t<int_elem_t> res = 0;
```
alonre24, line 81: This generic introduced for possible future integer types larger than 1 byte?

meiravgri (dismissed):

src/VecSim/spaces/IP/IP.cpp
GuyAv46: Yes, somewhat (at least 16-bit types).
More important is the static assertions to make sure nobody uses this templated function with unfitting types

alonre24 (approved):

meiravgri (approved):

PR #603, opened 2025-03-11
[MOD-8438] Micro-benchmarks for multi - int8 + benchmark ranges generalization 

lerman25:
This PR change/adds the following:

- Using the apply function to pass benchmark (BF/HNSW) arguments, enabling iteration over radius and epsilon ranges instead of manually specifying combinations.

- Introducing a benchmark_range template struct for better generalization, with static functions returning appropriate arrays based on index data type (default provided if not specialized for type).

- Adding multi int8 benchmarking with a generated dataset.

- Including a single_to_multi Python script (used for generating the int8 multi dataset) for future  use.

### DIFF
--- a/tests/benchmark/data/scripts/single_to_multi.py
+++ b/tests/benchmark/data/scripts/single_to_multi.py
@@ -1,6 +1,6 @@
 # This file is a template file for generating multi datasets from single
 # Using HNSW-knn for labeling
-# In this version, it create a multi dataset from the "wipedia_single" dataset used for int8
+# In this version, it creates a multi dataset from the "wipedia_single" dataset used for int8
 # Refrain from pushing changes unless necessary
 import h5py
 import numpy as np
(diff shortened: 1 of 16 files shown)


tests/benchmark/data/scripts/single_to_multi.py
```diff
@@ -0,0 +1,88 @@
+# This file is a template file for generating multi datasets from single
+# Using HNSW-knn for labeling
+# In this version, it create a multi dataset from the "wipedia_single" dataset used for int8
```
meiravgri, line 3:
```suggestion
# In this version, it creates a multi dataset from the "wipedia_single" dataset used for int8
```

tests/benchmark/data/scripts/single_to_multi.py
```diff
@@ -0,0 +1,88 @@
...
+
+inverse_groups = {}
+for vct,lbl in groups.items():
+    if lbl not in inverse_groups:
+        inverse_groups[lbl] = []
+    inverse_groups[lbl].append(vct)
+
+for i in range(n_vectors):
+    if i not in groups:
+        print("Error: Some vectors appear are missing!")
```
meiravgri, line 66:
```suggestion
        print("Error: Some vectors appear to be missing!")
```

meiravgri (commented):
awesome!

meiravgri (approved):

PR #605, opened 2025-03-13
[MOD-8439] Benchmark uint8 - single&multi

lerman25:
This PR adds the following:

Benchmarks for uint8 datatype, including single and multi datasets.
The single dataset is a int8->uint8 type conversion of int8 single dataset.
The multi dataset is generated with the single_to_multi.py file adjusted for uint8.
The type conversion script template is added with the uint-fy.py file.
The ranges for uint8 were selected by observing the cosine-distance distribution from a sample of the test set
and a sample of the dataset.


meiravgri (approved):
Perfect

PR #619, opened 2025-03-27
Add SVE/SVE2 support for uint8 and int8 data type [MOD-9080]

lerman25:
This PR add int8 and uint8 support for SVE & SVE2 ARM cpu flags. 

Important note
-  SVE2 has no improvements to offer comparing to SVE in regard for IP/L2 calculation, so it uses the same function.


dor-forer (dismissed):

src/VecSim/spaces/IP/IP_SVE_INT8.h
```diff
@@ -0,0 +1,101 @@
...
+
+static void InnerProductStep(const int8_t *&pVect1, const int8_t *&pVect2, size_t &offset,
+                             svfloat32_t &sum) {
+    svbool_t pg = svptrue_b8();
+
+    // Load int8 vectors
+    svint8_t v1_i8 = svld1_s8(pg, pVect1 + offset);
+    svint8_t v2_i8 = svld1_s8(pg, pVect2 + offset);
+
+    svfloat32_t ipf32 = svcvt_f32_s32_x(pg, svdot_s32(svdup_s32(0), v1_i8, v2_i8));
```
GuyAv46, line 18: why not accumulate the results with the expected `s32` type, and cast it to float only at the end?

src/VecSim/spaces/IP/IP_SVE_INT8.h
```diff
@@ -0,0 +1,101 @@
+/*
+ *Copyright Redis Ltd. 2021 - present
+ *Licensed under your choice of the Redis Source Available License 2.0 (RSALv2) or
+ *the Server Side Public License v1 (SSPLv1).
+ */
+#pragma once
+#include "VecSim/spaces/space_includes.h"
+#include <arm_sve.h>
+
+static void InnerProductStep(const int8_t *&pVect1, const int8_t *&pVect2, size_t &offset,
```
GuyAv46, line 10:
```suggestion
inline void InnerProductStep(const int8_t *&pVect1, const int8_t *&pVect2, size_t &offset,
```

src/VecSim/spaces/IP/IP_SVE_UINT8.h
```diff
@@ -0,0 +1,101 @@
+/*
+ *Copyright Redis Ltd. 2021 - present
+ *Licensed under your choice of the Redis Source Available License 2.0 (RSALv2) or
+ *the Server Side Public License v1 (SSPLv1).
+ */
+#pragma once
+#include "VecSim/spaces/space_includes.h"
+#include <arm_sve.h>
+
+static void InnerProductStep(const uint8_t *&pVect1, const uint8_t *&pVect2, size_t &offset,
```
GuyAv46, line 10:
```suggestion
inline void InnerProductStep(const uint8_t *&pVect1, const uint8_t *&pVect2, size_t &offset,
```

src/VecSim/spaces/IP/IP_SVE_UINT8.h
```diff
@@ -0,0 +1,101 @@
...
+                             svfloat32_t &sum) {
+    svbool_t pg = svptrue_b8();
+
+    // Load uint8 vectors
+    svuint8_t v1_ui8 = svld1_u8(pg, pVect1 + offset);
+    svuint8_t v2_ui8 = svld1_u8(pg, pVect2 + offset);
+
+    svfloat32_t ipf32 = svcvt_f32_u32_x(pg, svdot_u32(svdup_u32(0), v1_ui8, v2_ui8));
```
GuyAv46, line 18: why not accumulate the results with the expected `s32` type, and cast it to float only at the end?

src/VecSim/spaces/IP/IP_SVE_UINT8.h
```diff
@@ -0,0 +1,101 @@
...
+
+    sum = svadd_f32_x(pg, sum, ipf32);
+
+    offset += svcntb(); // Move to the next set of uint8 elements
```
GuyAv46, line 22: I wonder if this has a good performance. This function always returns the same value in runtime, but I'm not sure if the compiler can deduce that and only call it once. Have you tried to call it once at the beginning of the function and use its value, instead of calling it here?

src/VecSim/spaces/L2/L2_SVE_UINT8.h
```diff
@@ -0,0 +1,139 @@
...
+    svint16_t v1_16_l = svreinterpret_s16(svunpklo_u16(v1_ui8));
+    svint16_t v1_16_h = svreinterpret_s16(svunpkhi_u16(v1_ui8));
+    svint16_t v2_16_l = svreinterpret_s16(svunpklo_u16(v2_ui8));
+    svint16_t v2_16_h = svreinterpret_s16(svunpkhi_u16(v2_ui8));
+
+    // Calculate difference and square for low part
+    svint16_t diff_l = svsub_s16_x(pg, v1_16_l, v2_16_l);
+
+    svint32_t diff32_l_l = svunpklo_s32(diff_l);
+    svint32_t diff32_l_h = svunpkhi_s32(diff_l);
```
GuyAv46, line 29: why is this needed? the square diff of `u8` can fit in a `u16`

lerman25: Because values of diffs might be negative, and while modulo arithmetic make it so that the bits value is correct, when we'll multiply the value will no longer be true.

dor-forer (approved):

src/VecSim/spaces/L2/L2_SVE2_INT8.h
```diff
@@ -0,0 +1,105 @@
...
+    svint16_t diff_e = svsublb_s16(v1_i8, v2_i8);
+
+    // Subtract v2 from v1 and widen the results to int16 for the odd indexes
+    svint16_t diff_o = svsublt_s16(v1_i8, v2_i8);
+
+    sum = svmlalb_s32(sum, diff_e, diff_e);
+    sum = svmlalt_s32(sum, diff_e, diff_e);
+
+    sum = svmlalb_s32(sum, diff_o, diff_o);
+    sum = svmlalt_s32(sum, diff_o, diff_o);
```
GuyAv46, line 30: Are there no intrinsics like `svdot_s32` to multiply and fold i16 into i32?

src/VecSim/spaces/L2/L2_SVE2_UINT8.h
```diff
@@ -0,0 +1,105 @@
...
+                         svint32_t &sum, const size_t chunk) {
+    svbool_t pg = svptrue_b8();
+    // Note: Because all the bits are 1, the extention to 16 and 32 bits does not make a difference
+    // Otherwise, pg should be recalculated for 16 and 32 operations
+
+    svuint8_t v1_i8 = svld1_u8(pg, pVect1 + offset); // Load int8 vectors from pVect1
+    svuint8_t v2_i8 = svld1_u8(pg, pVect2 + offset); // Load int8 vectors from pVect2
+
+    // Subtract v2 from v1 and widen the results to int16 for the even indexes
+    svint16_t diff_e = svreinterpret_s16(svsublb_u16(v1_i8, v2_i8));
```
GuyAv46, line 21: is reinterpret cast the correct type of cast here? (and in all other L2 implementations)

GuyAv46 (commented):
Very nice and clean work

src/VecSim/spaces/IP/IP_SVE_INT8.h
```diff
@@ -0,0 +1,102 @@
...
+        if constexpr (additional_steps >= 2) {
+            InnerProductStep(pVect1, pVect2, offset, sum1, vl);
+        }
+        if constexpr (additional_steps >= 3) {
+            InnerProductStep(pVect1, pVect2, offset, sum2, vl);
+        }
+    }
+
+    if constexpr (partial_chunk) {
+        svbool_t pg = svwhilelt_b8(offset, dimension);
```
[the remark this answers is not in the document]
lerman25: Applied to all

src/VecSim/spaces/L2/L2_SVE2_INT8.h
lerman25: svdot is SVE, can be used only for  i8 -> i32
There is an i16 -> i64,  but I doubt it will be faster due to needing type conversion/ accumulating on i64, I suggest to keep it as  a TODO.

src/VecSim/spaces/L2/L2_SVE2_UINT8.h
lerman25: As we discussed, it's ok to reinterpret,  but adding comment to explain and  unit_test for future cases

GuyAv46 (approved):
Great work

dor-forer (approved):

GuyAv46 (approved):

PR #622, opened 2025-03-30
Install git to enable submodule fetch

alonre24:

After merging #621 - the checkout@v4 action cannot work without installing git before

**Mark if applicable**

### DIFF
--- a/.github/workflows/benchmark-runner.yml
+++ b/.github/workflows/benchmark-runner.yml
@@ -45,6 +45,8 @@ jobs:
     needs: start-runner
     runs-on: ${{ needs.start-runner.outputs.runner_label }}
     steps:
+      - name: Pre checkout deps
+        run:  sudo apt-get update && sudo apt-get -y install git
       - name: Checkout code
         uses: actions/checkout@v4
         with:
--- a/.github/workflows/alpine3.yml
+++ b/.github/workflows/alpine3.yml
@@ -7,5 +7,5 @@ jobs:
     uses: ./.github/workflows/task-unit-test.yml
     with:
       container: alpine:3
-      pre-checkout-script: apk add bash
+      pre-checkout-script: apk add bash git
       run-valgrind: true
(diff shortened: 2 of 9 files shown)


.github/workflows/benchmark-runner.yml
```diff
@@ -45,6 +45,8 @@ jobs:
     needs: start-runner
     runs-on: ${{ needs.start-runner.outputs.runner_label }}
     steps:
+      - name: Pre checkout deps
+        run:  sudo apt-get update && sudo apt-get -y install git
```
GuyAv46, line 49: We always use Ubuntu/Debian here, right?

GuyAv46 (approved):

.github/workflows/benchmark-runner.yml
alonre24: as far as I know...

.github/workflows/task-unit-test.yml
```diff
@@ -25,11 +25,10 @@ jobs:
     container: ${{ inputs.container || null }}
     defaults:
       run:
-        shell: bash -l -eo pipefail {0}
+        # Use bash as default shell for every OS other than alpine
+        shell: ${{ inputs.container && contains(inputs.container, 'alpine') && 'sh -l -eo pipefail {0}' || 'bash -l -eo pipefail {0}' }}
```
alonre24, line 29: Ugly, but couldn't make it work otherwise (Alpine doesn't have bash by default, so it fails immdietly, other os cannot accept the `sh -l -eo pipefail {0} command since their default shell do not support -o pipefail)

GuyAv46: you can use the steps we have on RediSearch, we handles it in 2 steps

alonre24: I see. Frankly, I don't think that one of the approaches is better than the other (both a bit ugly...)

GuyAv46 (approved):

PR #624, opened 2025-04-01
pin cmake version before brew update

meiravgri:

A clear and concise description of what the PR is solving.

1. #...
2. MOD...

1. ...
2. ...

**Mark if applicable**


.install/macos.sh
```diff
@@ -1,4 +1,11 @@
 #!/bin/bash
+
+# Without pinning cmake, it will install the latest version(>= 4.0)
+# This leads to to an error:
+# Compatibility with CMake < 3.5 has been removed from CMake.
+# For now we went with pinning cmake to 3.31.6 which is the version that is exists in the current mac OS docker image we use
```
GuyAv46, line 6:
```suggestion
# For now we went with pinning cmake to 3.31.6 which is the version that exists in the current MacOS image we use
```

PR #626, opened 2025-04-02
Remove pull_request_target and add submodules fetch where needed

alonre24:

**Mark if applicable**

### DIFF
--- a/.github/workflows/benchmark-labeled.yml
+++ b/.github/workflows/benchmark-labeled.yml
@@ -1,7 +1,7 @@
 name: benchmark
 
 on:
-  pull_request_target:
+  pull_request:
     types: [opened, reopened, synchronize, labeled] # Default ([opened, reopened, synchronize]) + labeled
     branches-ignore:
       - '*docs'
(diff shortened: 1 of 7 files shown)


.install/macos.sh
```diff
@@ -1,4 +1,11 @@
 #!/bin/bash
+
+# Without pinning cmake, it will install the latest version(>= 4.0)
+# This leads to to an error:
+# Compatibility with CMake < 3.5 has been removed from CMake.
+# For now we went with pinning cmake to 3.31.6 which is the version that is exists in the current mac OS docker image we use
```
meiravgri, line 6:
```suggestion
# For now we went with pinning cmake to 3.31.6 which is the version that is exists in the current macOS image we use
```

.github/workflows/flow-temp.yml
```diff
@@ -9,7 +9,7 @@ name: temporary testing
 
 on:
   push:
-    branches-ignore: ['**'] # ignore all branches. Comment this line to run your workflow below on every push.
+    # branches-ignore: ['**'] # ignore all branches. Comment this line to run your workflow below on every push.
```
meiravgri, line 12:
```suggestion
    branches-ignore: ['**'] # ignore all branches. Comment this line to run your workflow below on every push.
```

meiravgri (approved):

PR #665, opened 2025-04-27
[MOD-9557]  Fix incorrect vector blob size calculation

meiravgri:
# **Description:**

This PR fixes a bug where vector data was copied using an incorrect size, leading to truncated vectors and corrupted distance calculations in certain configurations.

## Bug Summary:

Previously, when copying a vector blob, we computed the size **manually** using:

```cpp
memcpy(blob_copy.get(), this->frontendIndex->getDataByInternalId(job->id),
       this->frontendIndex->getDim() * sizeof(DataType));
```
This approach ignores additional data stored alongside the vector. In particular, when using `int8` with `cosine` distance, a `float`-typed norm (4 bytes) is appended to the end of each vector. By not including this in the copy, the norm was lost.

This led to:

* Invalid norm values (0 or garbage) during distance computation.
* `-inf` similarity scores due to division by invalid norms.
* Many vectors selecting the same neighbors.
* Lock contention on shared nodes in multithreaded environments, significantly impacting performance.

## Fix:

We now use the correct size returned by `getDataSize()` for the copy:

```cpp
auto blob_copy = this->getAllocator()->allocate_unique(this->frontendIndex->getDataSize());

memcpy(blob_copy.get(), this->frontendIndex->getDataByInternalId(job->id),
       this->frontendIndex->getDataSize());
```

# Miscellaneous Improvements
### Eliminated unnecessary memory allocation and copying in `TieredHNSW_BatchIterator`
*  Previously, `TieredHNSW_BatchIterator` allocated memory and copied the original query blob, but never used this copy directly.
* The internal batch iterators  (`flat_iterator` and `hnsw_iterator`) create their own copies:
  * `flat_iterator` copy is created during `TieredHNSW_BatchIterator` construction
  * When `TieredHNSW_BatchIterator::getNextResults()` is called and `hnsw_iterator` is not initialized, it takes the blob from `flat_iterator`

### Test utilities
* **Added `getStoredVectorDataByLabel` Method**: Introduced a new method in `BruteForceIndex`, `HNSWIndex`, and their derived classes.  This new method retrieves the full stored vector, including appended metadata (e.g., norms).
* This complements the existing `getDataByLabel` method, which retrieves only the vector elements. 

## 📝 Note:

This fix also reveals a **bug** in vector preprocessing where the code assumes the input blob and processed blob have the same size:

```cpp
storage_blob = this->allocator->allocate(processed_bytes_count);
memcpy(storage_blob, original_blob, processed_bytes_count);
```
This is problematic because:

- For int8 vectors with cosine similarity,  `processed_bytes_count` includes space for the norm
- But `original_blob` might only contain the vector elements `(dim * sizeof(DataType))`
- **This may result in out-of-bounds `memcpy` reads.**

## This issue will be fully addressed in a follow-up PR, see [MOD-9576](https://redislabs.atlassian.net/browse/MOD-9576).
In this PR, preliminary steps have been taken:
1. TODO added: A comment has been placed in the preprocessor logic to flag the need for handling size mismatches:
`// TODO: handle original_blob_size != processed_bytes_count`

2.  🔄 Batch Iterator Modification:

To eliminate duplicated, potentially unsafe copy logic and ensure consistent handling of blob size and memory allocation, the responsibility for copying the query blob has been delegated from the batch iterator to the preprocessor.

Previously, the batch iterator manually allocated memory and copied the input before invoking `preprocessQueryInPlace()`:

```cpp
auto *queryBlobCopy =
    this->allocator->allocate_aligned(this->dataSize, this->preprocessors->getAlignment());
memcpy(queryBlobCopy, queryBlob, this->dim * sizeof(DataType));
this->preprocessQueryInPlace(queryBlobCopy);
```
This has now been replaced with a call to the existing `preprocessQuery()` method with `force_copy = true`:

```cpp
auto queryBlobCopy = this->preprocessQuery(queryBlob, true);
auto *queryBlobCopyPtr = queryBlobCopy.release();  // Take ownership
```

[MOD-9576]: https://redislabs.atlassian.net/browse/MOD-9576?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/VecSim/algorithms/brute_force/brute_force_multi.h
```diff
@@ -47,10 +47,30 @@ class BruteForceIndex_Multi : public BruteForceIndex<DataType, DistType> {
...
         for (idType id : ids->second) {
             auto vec = std::vector<DataType>(this->dim);
+            // Only copy the vector data (dim * sizeof(DataType)), not any additional metadata like
+            // the norm
             memcpy(vec.data(), this->getDataByInternalId(id), this->dim * sizeof(DataType));
             vectors_output.push_back(vec);
         }
     }
+
+    std::vector<std::vector<char>> getStoredVectorDataByLabel(labelType label) const override {
```
GuyAv46, line 57: Not used by any test?

src/VecSim/algorithms/brute_force/brute_force.h
```diff
@@ -77,15 +77,41 @@ class BruteForceIndex : public VecSimIndexAbstract<DataType, DistType> {
...
+     * - For other vector types or future implementations, this will include any additional data
+     *   that might be stored with the vector
+     *
+     * Use this method when you need access to the complete vector data as it is stored internally.
+     *
+     * @param label The label to retrieve data for
+     * @return A vector containing the complete vector data (elements + metadata) for the given
+     * label
+     */
+    virtual std::vector<std::vector<char>> getStoredVectorDataByLabel(labelType label) const = 0;
```
GuyAv46, line 113: 1. Can we move the two helpers under `VecSimIndexAbstract` (test only), so we don't need to redefine it in `HNSW`?
2. Consider renaming - I'd expect `getDataByLabel` to return all the data. Maybe `getDataByLabel` and `getVectorByLabel`, or something similar

src/VecSim/spaces/computer/preprocessor_container.cpp
```diff
@@ -32,15 +31,16 @@ void PreprocessorsContainerAbstract::preprocessQueryInPlace(void *blob,
...
-            return this->wrapAllocated(aligned_mem);
-        }
+MemoryUtils::unique_blob PreprocessorsContainerAbstract::maybeCopyToAlignedMem(
+    const void *original_blob, size_t blob_bytes_count, bool force_copy) const {
+    bool needs_copy =
+        force_copy || (this->alignment && ((uintptr_t)original_blob % this->alignment != 0));
+
+    if (needs_copy) {
+        auto aligned_mem = this->allocator->allocate_aligned(blob_bytes_count, this->alignment);
+        // TODO: handle original_blob_size != processed_bytes_count
```
GuyAv46, line 41: When? Is there something that will fail a PR if we forget to support it?

src/VecSim/spaces/computer/preprocessors.h
```diff
@@ -57,9 +57,11 @@ class CosinePreprocessor : public PreprocessorInterface {
             // If one of them is null, allocate memory for it and copy the original_blob to it.
             if (storage_blob == nullptr) {
                 storage_blob = this->allocator->allocate(processed_bytes_count);
+                // TODO: handle original_blob_size != processed_bytes_count
```
GuyAv46, line 60: When? Is there something that will fail a PR if we forget to support it? (relevant to all the changes in the file)

GuyAv46 (commented):
Nice catch and nice tests!

src/VecSim/spaces/computer/preprocessor_container.cpp
meiravgri: ASAP. it is undefined potentially it can lead to a crash

src/VecSim/algorithms/brute_force/brute_force.h
meiravgri: moved

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -924,9 +923,10 @@ double TieredHNSWIndex<DataType, DistType>::getDistanceFrom_Unsafe(labelType lab
 
 template <typename DataType, typename DistType>
 TieredHNSWIndex<DataType, DistType>::TieredHNSW_BatchIterator::TieredHNSW_BatchIterator(
-    void *query_vector, const TieredHNSWIndex<DataType, DistType> *index,
+    const void *query_vector, const TieredHNSWIndex<DataType, DistType> *index,
     VecSimQueryParams *queryParams, std::shared_ptr<VecSimAllocator> allocator)
-    : VecSimBatchIterator(query_vector, queryParams ? queryParams->timeoutCtx : nullptr,
+    // Tiered batch iterator doesn't hold its own copy of the query vector.
```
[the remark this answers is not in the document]
meiravgri: done

src/VecSim/vec_sim_tiered_index.h
```diff
@@ -110,6 +110,9 @@ class VecSimTieredIndex : public VecSimIndexInterface {
     inline VecSimIndexAbstract<DataType, DistType> *getFlatBufferIndex() {
         return this->frontendIndex;
     }
+    inline BruteForceIndex<DataType, DistType> *getFlatBufferIndexAsBruteForce() {
```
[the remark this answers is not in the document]
meiravgri: makes sense

GuyAv46 (approved):
Nicely done!
Lets not forget the renaming of the `getData` helpers

alonre24 (dismissed):

GuyAv46 (approved):

PR #677, opened 2025-05-21
Solve issues with mutexes while running tests on MacOS M1 

JoanFM:

2 issues are handled in this PR:

1- An `abort` was identified while running a test related to `PyBatchIterator`. The abort happened when the test was being shutdown. The error comes from the `deleter`passed to the `PyBatchIterator` `shared_ptr` which captures the `PyIndex`by reference. The problem is that there is no guarantee that the `PyIndex`lifetime will survive the `Iterator`. In that deleter a `mutex` is released (which could be in an invalid state). The solution is to have the `mutex`be wrapped in a shared_ptr that will share the ownership with the `PyBatchIterator` deleter as it captures it by value (incrementing its reference count).

2- Deadlock found on `tests/flow/test_hnsw_parallel.py::test_parallel_insert_batch_search`:

- Thread #N-1 has the `indexGuard` locked in shared mode, and tries to recover the GIL in the getNextResult
- Thread #N tries to acquire `indexGuard` exclusively, while N-1 has the indexGuard in shared mode. (WAIT)
- Thread #N+1 is another iterator that want `indexGuard` in shared mode (It goes in the queue) but cannot get it in shared mode (I guess is System dependant if they would schedule a reader to bypass the writer ) potential starvation. Thread N+1 has the GIL.

This is solved by releasing the GIL in `Thread N+1` before asking for `indexGuard`.

### DIFF
--- a/src/python_bindings/bindings.cpp
+++ b/src/python_bindings/bindings.cpp
@@ -258,9 +258,10 @@ private:
                 if (ind >= n_queries) {
                     break;
                 }
-                indexGuard.lock_shared();
-                results[ind] = queryFunc((const char *)items.data(ind), param, query_params);
-                indexGuard.unlock_shared();
+                {
+                    std::shared_lock<std::shared_mutex> lock(indexGuard);
+                    results[ind] = queryFunc((const char *)items.data(ind), param, query_params);
+                }
             }
         };
         std::thread thread_objs[n_threads];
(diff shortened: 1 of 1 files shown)


src/python_bindings/bindings.cpp
```diff
@@ -395,23 +398,21 @@ class PyHNSWLibIndex : public PyVecSimIndex {
...
-                        exclusive = false;
+                        // Read lock for normal operations
+                        std::shared_lock<std::shared_mutex> index_guard(*indexGuard);
+                        this->addVectorInternal((const char *)data.data(ind), labels.at(ind));
                     } else {
-                        // Lock exclusively if we are performing resizing due to a new block.
-                        indexGuard.lock();
+                        // Exclusive lock for block resizing operations
+                        std::unique_lock<std::shared_mutex> index_guard(*indexGuard);
+                        this->addVectorInternal((const char *)data.data(ind), labels.at(ind));
```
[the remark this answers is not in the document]
JoanFM: will revert

src/python_bindings/bindings.cpp
```diff
@@ -459,12 +463,15 @@ class PyHNSWLibIndex : public PyVecSimIndex {
     }
     PyBatchIterator createBatchIterator(const py::object &input,
                                         VecSimQueryParams *query_params) override {
+
         py::array query(input);
-        auto del = [&](VecSimBatchIterator *pyBatchIter) {
+        py::gil_scoped_release py_gil;
```
alonre24, line 468: Nice catch!
Let's also document when declaring `indexGuard` that it must be held after releasing Python GIL to avoid deadlocks

JoanFM: Done!

alonre24 (approved):

PR #680, opened 2025-05-25
add to gcc11-bullseye to workflows - [MOD-9529]

BenGoldberger:
test redisearch on debian-bullseye with gcc>10


.github/workflows/event-merge-to-queue.yml
```diff
@@ -49,6 +49,7 @@ jobs:
       - focal
 #      - bionic
       - bullseye
+      - bullseye-gcc11
```
alonre24, line 52: I don't think we should have both "regular" bullseye and gcc11 bullseye on every PR merge. If you believe we need to have both flows, than let's keep it only in nightly

.github/workflows/debian11.yml
```diff
@@ -8,3 +8,9 @@ jobs:
     with:
       container: debian:bullseye
       pre-checkout-script: apt-get update && apt-get -y install git
+
+  bullseye-gcc11:
```
alonre24, line 12: Please add a comment to the PR with a link to a successful run of this new job

BenGoldberger: https://github.com/RedisAI/VectorSimilarity/pull/680#issuecomment-2909403334

BenGoldberger: https://github.com/RedisAI/VectorSimilarity/pull/680#issuecomment-2909513795

.github/workflows/event-merge-to-queue.yml
BenGoldberger: ok. I wasnt sure

alonre24 (dismissed):

.github/workflows/event-merge-to-queue.yml
BenGoldberger: when I tried to add a file the event-nightly flow was angry for using too many files : [here](https://github.com/RedisAI/VectorSimilarity/actions/runs/15255841541)

So I added a boolean option and as you can see the merge-queue doesnt run the job [here](https://github.com/RedisAI/VectorSimilarity/actions/runs/15270874514)

PR #689, opened 2025-06-03
Enabler for proper de-duplication of query results in tiered index with compression

meiravgri:
**Describe the changes in the pull request**

A clear and concise description of what the PR is solving.

**Which issues this PR fixes**
1. #...
2. MOD...

**Main objects this PR modified**
1. ...
2. ...

**Mark if applicable**

- [ ] This PR introduces API changes
- [ ] This PR introduces serialization changes

PR #692, opened 2025-06-08
Enabler for proper de-duplication of query results in tiered index with compression

alonre24:

(replacing #689 so we can merge this directly to main)
This PR changes the tiered index KNN and range search implementation to properly de-duplicate query results when the two underlying indexes have the same label with different scores. This is possible in multi, butt will also be possible for indexes with compression where full size vectors are stored in flat bugger and compressed vectors are in the backend index.

To test this change, we create a tiered HNSW index (the only tiered index that exists today) and inject labels with different vectors into its two tiers. 

This also includes:
* Exposing the API for getting the number of unique labels in an index, and relaxing the locking in the tiered HNSW implementation. 
* Removal of the expected result count parameter from runTopKSearchTes and the equivalent for range search, adding a new  `runTopKTieredIndexSearchTest` call that calls the internal search function that can search `withSet` and filter duplicate labels results with different scores.
* Install the latest clang-18 in CI to the local format will suits the remote one.

**Mark if applicable**

### DIFF
--- a/.install/ubuntu_24.04.sh
+++ b/.install/ubuntu_24.04.sh
@@ -4,6 +4,14 @@ export DEBIAN_FRONTEND=noninteractive
 MODE=$1 # whether to install using sudo or not
 
 $MODE apt update -qq
+sudo apt install -y wget lsb-release gnupg
+
+# Add the LLVM APT repo for the latest clang-18 version
+wget https://apt.llvm.org/llvm.sh
+chmod +x llvm.sh
+$MODE ./llvm.sh 18
+clang-format-18 --version
+
 $MODE apt install -yqq git wget build-essential lcov openssl libssl-dev \
-    python3 python3-venv python3-dev unzip rsync clang curl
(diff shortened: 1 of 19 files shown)


rfsaliev:
Hi!
I have the question, how the 'Singe-Vector' mode 'topKQuery{k=2}' should work in the following case:
* let `q` - query vector
* let `v1; v2; v3; v4; v5` are vectors which distance to `q` in appropriate order: `v1` is closer than `v2`, etc.
* let assume, that during some index updates, we got the case when: frontend contains: { `id1:v4`; `id2:v5` }, backend: { `id1:v1`; `id2:v2`; `id3:v3` }
* top2 frontend: {`id1`, `id2`}, top2 backend: {`id1`, `id2`}.
* `id1` and `id2` in backend are outdated, so we should select vectors from frontend: {v4, v5}, but backend contains valid vector {v3} which is closer than v4 and v5, but not received from backend top2 query.

PR #702, opened 2025-06-19
Expose SVS Vamana defaults + debug info [MOD-9759]

alonre24:

Expose SVS Vamana default build parameters to be used by redisearch in ft.info, and implement the `debugInfo` API for svs index and tiered svs. 
This includes adding unit tests for the svs factory to validate that both default and non-default parameters are passed properly, and refining the debug info call for the tiered index to compute the effective label set count internally, thereby avoiding the need to reacquire locks in `indexLabelCount()`. The last change was possible due to moving the `getLabelSet` virtual call to the abstract index level (all derived indexes were already implementing this call)

**Mark if applicable**


src/VecSim/vec_sim_common.h
```diff
@@ -17,21 +17,37 @@ extern "C" {
...
 #define HNSW_INVALID_LEVEL SIZE_MAX
 #define INVALID_JOB_ID     UINT_MAX
 #define INVALID_INFO       UINT_MAX
 
+// SVS-Vamana default parameters
+#define SVS_VAMANA_DEFAULT_ALPHA_L2                 1.2f
+#define SVS_VAMANA_DEFAULT_ALPHA_IP                 0.95f
+#define SVS_VAMANA_DEFAULT_GRAPH_MAX_DEGREE         32
+#define SVS_VAMANA_DEFAULT_CONSTRUCTION_WINDOW_SIZE 200
+#define SVS_VAMANA_DEFAULT_MAX_CANDIDATE_POOL_SIZE  (3 * SVS_VAMANA_DEFAULT_CONSTRUCTION_WINDOW_SIZE)
```
meiravgri, line 40: what if SVS_VAMANA_DEFAULT_CONSTRUCTION_WINDOW_SIZE is defined by the user differently than the default, but SVS_VAMANA_DEFAULT_MAX_CANDIDATE_POOL_SIZE is not?

src/VecSim/vec_sim_common.h
```diff
@@ -17,21 +17,37 @@ extern "C" {
...
+#define SVS_VAMANA_DEFAULT_PRUNE_TO                 (SVS_VAMANA_DEFAULT_GRAPH_MAX_DEGREE - 4)
```
meiravgri, line 41: what if SVS_VAMANA_DEFAULT_GRAPH_MAX_DEGREE is defined by the user differently than the default, but SVS_VAMANA_DEFAULT_PRUNE_TO is not?

src/VecSim/vec_sim_common.h
```diff
@@ -17,21 +17,37 @@ extern "C" {
...
+#define SVS_VAMANA_DEFAULT_USE_SEARCH_HISTORY       true
+#define SVS_VAMANA_DEFAULT_NUM_THREADS              1
+// NOTE: optimal training threshold may depend on the SVSIndex compression mode.
+// it might be good to implement a utility to compute default threshold based on index parameters
+// DEFAULT_BLOCK_SIZE is used to round the training threshold to FLAT index blocks
+#define SVS_VAMANA_DEFAULT_TRAINING_THRESHOLD (10 * DEFAULT_BLOCK_SIZE) // 10 * 1024 vectors
```
meiravgri, line 47: so if I understood correctly your comment, the actual value might be different than what is defined here?

src/VecSim/algorithms/svs/svs_utils.h
```diff
@@ -24,10 +24,6 @@
 #include <string>
 #include <utility>
 
-// NOTE: optimal training threshold may depend on the SVSIndex compression mode.
-// it might be good to implement an utility to compute default threshold based on index parameters
-// DEFAULT_BLOCK_SIZE is used to round the training threshold to FLAT index blocks
-constexpr size_t SVS_DEFAULT_TRAINING_THRESHOLD = 10 * DEFAULT_BLOCK_SIZE; // 10 * 1024 vectors
 // Maximum training threshold for SVS index, used to limit the size of training data
 constexpr size_t SVS_MAX_TRAINING_THRESHOLD = 100 * DEFAULT_BLOCK_SIZE; // 100 * 1024 vectors
```
meiravgri, line 28: Why wasn't this moved as well?

src/VecSim/algorithms/svs/svs_utils.h
```diff
@@ -102,18 +98,20 @@ makeVamanaBuildParameters(const SVSParams &params) {
...
     return svs::index::vamana::VamanaBuildParameters{
-        getOrDefault(params.alpha, (params.metric == VecSimMetric_L2 ? 1.2f : 0.95f)),
+        getOrDefault(params.alpha, (params.metric == VecSimMetric_L2 ?
+            SVS_VAMANA_DEFAULT_ALPHA_L2 : SVS_VAMANA_DEFAULT_ALPHA_IP)),
         graph_max_degree,
         construction_window_size,
         getOrDefault(params.max_candidate_pool_size, construction_window_size * 3),
         getOrDefault(params.prune_to, graph_max_degree - 4),
-        params.use_search_history != VecSimOption_DISABLE
+        params.use_search_history == VecSimOption_AUTO ? SVS_VAMANA_DEFAULT_USE_SEARCH_HISTORY :
```
meiravgri, line 113: isn't `VecSimOption_AUTO` is a valid configuration that in the future might not necessarily `SVS_VAMANA_DEFAULT_USE_SEARCH_HISTORY?`

i believe that it was intended to implement something like:
```
if (params.use_search_history == VecSimOption_AUTO) ? choose_best_search_history_config()
else  if (params.use_search_history == VecSimOption_ENABLE) true
else /* VecSimOption_DISABLE */ false
```

src/VecSim/algorithms/svs/svs_utils.h
```diff
@@ -102,18 +98,20 @@ makeVamanaBuildParameters(const SVSParams &params) {
...
+        (params.use_search_history == VecSimOption_ENABLE)
```
meiravgri, line 114: although VecSimOption_ENABLE == 1 that translates to false (but not in rust :D ! ), it is confusing that the actual meaning is VecSimOption_DISABLE
considering that this is currently a boolean i would do 
```suggestion
        !SVS_VAMANA_DEFAULT_USE_SEARCH_HISTORY
```

src/VecSim/vec_sim_common.h
alonre24: It's a comment from a previous PR that introduces this, we may want to do it in the future (not now)

src/VecSim/vec_sim_common.h
alonre24: same as for SVS_VAMANA_DEFAULT_MAX_CANDIDATE_POOL_SIZE...

src/VecSim/algorithms/svs/svs_utils.h
alonre24: right, but VecSimOption_AUTO is also the default value (0), and there is no `choose_best_search_history_config()` implentation yet (the default is always true)

src/VecSim/vec_sim_common.h
meiravgri: Do you mean it shouldn’t be configurable at all, or just not exposed via the info command?
In any case, if we do expose it in the info output, I think it should reflect the correct runtime-derived value. So maybe the options are:
1. Don’t expose it now and possibly introduce a runtime fetch API later if needed, or
2. Remove it altogether if it's not meant to be user-facing.

meiravgri (dismissed):

tests/unit/test_svs.cpp
```diff
@@ -82,6 +82,45 @@ class SVSTest : public ::testing::Test {
         return VecSimSvsQuant_Scalar ==
                std::get<0>(svs_details::isSVSQuantBitsSupported(index_type_t::get_quant_bits()));
     }
+
+    // Expected values structure for parameter validation
+    struct ExpectedSVSValues {
```
meiravgri, line 87: what is the difference between this struct and svsInfoStruct?

tests/unit/test_svs.cpp
```diff
@@ -82,6 +82,45 @@ class SVSTest : public ::testing::Test {
...
+        EXPECT_EQ(info.svsInfo.maxCandidatePoolSize, expected.max_candidate_pool_size);
+        EXPECT_EQ(info.svsInfo.pruneTo, expected.prune_to);
+        EXPECT_EQ(info.svsInfo.useSearchHistory, expected.use_search_history);
+        EXPECT_EQ(info.svsInfo.numThreads, expected.num_threads);
+        EXPECT_EQ(info.svsInfo.searchWindowSize, expected.search_window_size);
+        EXPECT_DOUBLE_EQ(info.svsInfo.epsilon, expected.epsilon);
+
+        // numberOfMarkedDeletedNodes should be 0 for a newly created index
+        EXPECT_EQ(info.svsInfo.numberOfMarkedDeletedNodes, 0);
+    }
```
meiravgri, line 123: can we use here compareSVSInfo?

src/VecSim/algorithms/svs/svs_tiered.h
```diff
@@ -451,9 +452,10 @@ class TieredSVSIndex : public VecSimTieredIndex<DataType, float> {
         return result;
     }
 
+    backend_index_t *GetBackendIndex() const { return this->backendIndex; }
```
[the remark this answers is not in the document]
alonre24: Not sure I understand, this is for getting the **backend** index, not the flat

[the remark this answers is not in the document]
rfsaliev: This function was defined just for unit tests to avoid `svs_XXX_tests_friends.h` files.
Internally, `this->backendIndex` can be used instead.

rfsaliev (commented):
Some comments to `getLabelsSet()`

src/VecSim/vec_sim_interface.h
```diff
@@ -81,6 +81,8 @@ struct VecSimIndexInterface : public VecsimBaseObject {
 
     /**
      * @brief Return the number of unique labels in the index (which are not deleted).
+     * !!! Note: this should only be called in debug mode, as in tiered indexes it may require
```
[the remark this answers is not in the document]
alonre24: Right... so I changed the comment and added new API only for tiered index to have a more distinct separation between tiered index and non-tiered. I didn't move the API to VecSimIndexAbstract to avoid many changes that would be needed in tests and benchmarks...

src/VecSim/vec_sim_tiered_index.h
```diff
@@ -301,12 +301,23 @@ VecSimIndexDebugInfo VecSimTieredIndex<DataType, DistType>::debugInfo() const {
...
+        // Compute the union of the labels from both indices.
+        std::vector<size_t> labels_union;
+        auto flat_labels = this->frontendIndex->getLabelsSet();
+        auto main_labels = this->backendIndex->getLabelsSet();
+        labels_union.reserve(flat_labels.size() + main_labels.size());
+        std::set_union(flat_labels.begin(), flat_labels.end(), main_labels.begin(),
+                       main_labels.end(), std::back_inserter(labels_union));
+        info.commonInfo.indexLabelCount = labels_union.size();
+    }
+
```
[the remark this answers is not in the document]
alonre24: not relevant since not adding the new API. In tiered indexes, this should be guarded anyway

src/VecSim/vec_sim_index.h
```diff
@@ -305,9 +305,15 @@ struct VecSimIndexAbstract : public VecSimIndexInterface {
...
     virtual std::vector<std::vector<char>> getStoredVectorDataByLabel(labelType label) const = 0;
-
 #endif
 
+    /**
+     * Virtual functions that access the label lookup which is implemented in the derived classes
+     * Return all the labels in the index - this should be used for computing the number of distinct
+     * labels in a tiered index, and caller should hold the appropriate guards.
+     */
+    virtual vecsim_stl::set<labelType> getLabelsSet() const = 0;
```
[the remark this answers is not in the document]
alonre24: What do you mean by " with the implementation in `VecSimTieredIndex` as well." ? `VecSimTieredIndex` derives from `VecSimIndexInterface` and will have to implement it as well if I move it. Why would that be better?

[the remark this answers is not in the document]
rfsaliev: As I understand, `getLabelsSet()` is declared in `VecSimIndexAbstract` to avoid it's implementation for tiered indices.
I propose to implement it for tiered indices as well and reuse it for labels count queries. Such implementation can be common for all kind of tiered indices, i.e. in `VecSimTieredIndex`.

src/VecSim/algorithms/hnsw/hnsw_multi.h
```diff
@@ -42,8 +42,9 @@ class HNSWIndex_Multi : public HNSWIndex<DataType, DistType> {
     inline void resizeLabelLookup(size_t new_max_elements) override;
 
     // Return all the labels in the index - this should be used for computing the number of distinct
-    // labels in a tiered index, and caller should hold the index data guard.
+    // labels in a tiered index.
     inline vecsim_stl::set<labelType> getLabelsSet() const override {
+        std::shared_lock<std::shared_mutex> index_data_lock(this->indexDataGuard);
```
meiravgri, line 47: consider using the api lockSharedIndexDataGuard

src/VecSim/vec_sim_tiered_index.h
```diff
@@ -121,6 +121,38 @@ class VecSimTieredIndex : public VecSimIndexInterface {
...
+     * This is a debug-only method for tiered indexes that computes the union of labels
+     * from both frontend and backend indexes. It may require locking and is time-consuming.
+     * !!! Note: this should only be called in debug mode for tiered indexes !!!
+     *
+     * @return index label count for debug purposes.
+     */
+    virtual size_t computeUnifiedIndexLabelCount() const {
+        // Compute the union of both labels set in both tiers of the index.
+        // This is the expensive debug-only implementation
+        auto [flat_labels, svs_labels] = [this] {
```
meiravgri, line 135:
```suggestion
        auto [frontend_labels, backend_labels] = [this] {
```

src/VecSim/vec_sim_tiered_index.h
```diff
@@ -121,6 +121,38 @@ class VecSimTieredIndex : public VecSimIndexInterface {
...
+    /**
+     * @brief Return the number of unique labels in the index (which are not deleted).
+     * This is a debug-only method for tiered indexes that computes the union of labels
+     * from both frontend and backend indexes. It may require locking and is time-consuming.
+     * !!! Note: this should only be called in debug mode for tiered indexes !!!
+     *
+     * @return index label count for debug purposes.
+     */
+    virtual size_t computeUnifiedIndexLabelCount() const {
+        // Compute the union of both labels set in both tiers of the index.
```
meiravgri, line 133:
```suggestion
        // Compute the union of both labels set in both tiered of the index.
```

src/VecSim/vec_sim.cpp
```diff
@@ -190,10 +190,6 @@ extern "C" void VecSim_Normalize(void *blob, size_t dim, VecSimType type) {
 
 extern "C" size_t VecSimIndex_IndexSize(VecSimIndex *index) { return index->indexSize(); }
 
-extern "C" size_t VecSimIndex_IndexLabelCount(VecSimIndex *index) {
```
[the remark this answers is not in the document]
alonre24: no, search collects that via `VecSimIndex_DebugInfo()`

src/VecSim/vec_sim_tiered_index.h
alonre24: :)

src/VecSim/vec_sim_tiered_index.h
```diff
@@ -301,12 +333,23 @@ VecSimIndexDebugInfo VecSimTieredIndex<DataType, DistType>::debugInfo() const {
     VecSimIndexDebugInfo info;
     this->flatIndexGuard.lock_shared();
     this->mainIndexGuard.lock_shared();
+
     VecSimIndexDebugInfo frontendInfo = this->frontendIndex->debugInfo();
     VecSimIndexDebugInfo backendInfo = this->backendIndex->debugInfo();
+    {
+        // Compute the union of the labels from both indices.
+        std::vector<size_t> labels_union;
```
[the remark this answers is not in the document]
alonre24: It is the same. I made computeUnifiedIndexLabelCount unsafe so we can call it from both places

src/VecSim/algorithms/hnsw/hnsw_single.h
```diff
@@ -107,9 +107,10 @@ size_t HNSWIndex_Single<DataType, DistType>::indexLabelCount() const {
  */
 
 // Return all the labels in the index - this should be used for computing the number of distinct
-// labels in a tiered index, and caller should hold the index data guard.
+// labels in a tiered index.
 template <typename DataType, typename DistType>
 vecsim_stl::set<labelType> HNSWIndex_Single<DataType, DistType>::getLabelsSet() const {
+    std::shared_lock<std::shared_mutex> index_data_lock(this->indexDataGuard);
```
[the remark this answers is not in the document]
alonre24: I don't think so, the main index guard that is assumed to be held from the tiered index should be enough (we never access internal svs locks from our code), right @rfsaliev ?

src/VecSim/algorithms/hnsw/hnsw_multi.h
alonre24: Though we are using that api in many places, I believe it is a better practice to acquire locks like this where possible

PR #733, opened 2025-07-24
[SVS] Fix SVSTieredIndexTestBasic.overwriteVectorAsync test

rfsaliev:
1. Data-racing for SVS index threadpool size - now the SVS threadpool size is always updated under backend index lock
2. Wrong addVector() returning value in case of overwriting vector is being moved by the update job:
  If update job moved an updating vector during execution of `SVSTieredIndex::addVector()` it is possible the case when `svs_index->deleteVectors()` returns 0 and `frontendIndex->addVector()` returns 1 even if label exists in the index:
    * Initial state: frontend index contains a vector with the label `L` and value `X` (`{L: X}`); backend index is `empty`
    * User calls `addVector(L, Y)`
    * `addVector()`:
      * lock backend index
      * call `backend->deleteVector(L)` => `ret = 0`
      * unlock backend index
      * `<thread suspended>`
    * Background update job is started and finished
      * Index state: frontend is `empty`; backend contains `{L: X}`
    * `addVector()`:
      * `<thread wakeup>`
      * call `frontend->addVector(L, Y)` => `ret = 1` - because frontend is `emptied` by update job
    * User receives `ret=1` when `ret=0` is expected 

1. `SVSTieredIndex`

**Mark if applicable**

### DIFF
--- a/src/VecSim/algorithms/svs/svs_tiered.h
+++ b/src/VecSim/algorithms/svs/svs_tiered.h
@@ -724,30 +724,37 @@ public:
         assert(this->getWriteMode() != VecSim_WriteInPlace && "InPlace mode returns early");
 
         // Async mode - add vector to the frontend index and schedule an update job if needed.
+        if (!this->backendIndex->isMultiValue()) {
+            {
+                std::shared_lock lock(this->flatIndexGuard);
+                // If the label already exists in the frontend index, we should count it
+                // to prevent the case when existing vector is moved meanwhile by the update job.
+                if (this->frontendIndex->isLabelExists(label)) {
+                    ret = -1;
+                }
+            }
+            // Remove vector from the backend index if it exists in case of non-MULTI.
+            std::lock_guard lock(this->mainIndexGuard);
+            ret -= svs_index->deleteVectors(&label, 1);
+            // If main index is empty then update_threshold is trainingTriggerThreshold,
+            // overwise it is 1.
+        }
         { // Add vector to the frontend index.
             std::lock_guard lock(this->flatIndexGuard);
-            ret = this->frontendIndex->addVector(blob, label);
+            const auto ft_ret = this->frontendIndex->addVector(blob, label);
 
-            if (ret == 0) { // Vector was overriden - add 'skiping' swap to the journal.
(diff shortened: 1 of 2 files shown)


rfsaliev (commented):
TODO

src/VecSim/algorithms/svs/svs_tiered.h
```diff
@@ -618,6 +617,7 @@ class TieredSVSIndex : public VecSimTieredIndex<DataType, float> {
             std::lock_guard lock(this->mainIndexGuard);
             auto svs_index = GetSVSIndex();
             assert(labels_to_move.size() == vectors_to_move.size() / this->frontendIndex->getDim());
+            svs_index->setNumThreads(std::min(availableThreads, labels_to_move.size()));
```
alonre24, line 620: An optimization only, right? Not directly related to the fix?

src/VecSim/algorithms/svs/svs_tiered.h
```diff
@@ -725,16 +725,17 @@ class TieredSVSIndex : public VecSimTieredIndex<DataType, float> {
 
         // Async mode - add vector to the frontend index and schedule an update job if needed.
         if (!this->backendIndex->isMultiValue()) {
+            {
+                std::shared_lock lock(this->flatIndexGuard);
+                // If the label already exists in the frontend index, we should count it
+                // to prevent the case when existing vector is moved meanwhile by the update job.
+                if (this->frontendIndex->isLabelExists(label)) {
+                    ret = -1;
+                }
```
alonre24, line 734: Can we move this logic to after line 741, so we will not have to acquire the flat index lock twice?

rfsaliev: There is still possibile that the vector can be moved between lines 735-737, so we can count it twice, and `ret==-2`.
But this case is covered by `std::max()` call at line 751

alonre24: I understand. So there is no way to avoid this additional locking?

alonre24 (dismissed):

alonre24 (approved):

PR #752, opened 2025-08-26
SVS updates for SVS_LAZY and inf/nan

ethanglaser:

This PR update the SVS binaries and submodule revision, containing inf/nan fixes and SVS_LAZY update to address alpine issue

**Mark if applicable**

- [X] This PR updates SVS shared library


meiravgri:
Currently blocked due to valgrind failure on the merge queue.

>  It looks like Valgrind doesn't handle certain types of NaN values well—especially when MKL propagates input NaNs through matrix computations (e.g., matmul), which then show up in Leanvec datasets. Valgrind flags these cases, even though they’re technically valid.
Workaround in progress.

PR #925, opened 2026-03-31
refactor: SVS shared thread pool with rental model (MOD-9881)

meiravgri:
## Overview

Replaces per-index SVS thread pools with a **single shared thread pool** that can be physically resized at runtime, and introduces a rental-based execution model for `parallel_for`. This eliminates over-provisioning (previously K indexes × N threads = K×N OS threads) and enables the SVS pool to actually grow/shrink when the external pool capacity changes — something the old design could not do.

Also adds a new C API `VecSim_UpdateThreadPoolSize()` that combines write-mode toggling with pool resizing, and deprecates `SVSParams::num_threads`.

### The fundamental scheduling invariant

The shared SVS thread pool has **the same size as the RediSearch worker thread pool** (`search-workers`). Every SVS multi-threaded operation (update, GC) is dispatched via `SVSMultiThreadJob`, which submits N reserve jobs to the RediSearch worker pool — one per worker thread. Only the threads that actually check in (are not busy with other work) participate in the SVS operation. This means:

- **The sum of all concurrent `rent()` requests can never exceed the pool capacity**, because each rented SVS thread corresponds to a RediSearch worker that checked in via a reserve job. If a worker is busy, it doesn't check in, so it's never counted in `availableThreads` and never requested from the SVS pool.
- **`setParallelism(n)` is always called with `n = min(availableThreads, problemSize)`**, where `availableThreads ≤ poolSize()`. This is why the assertion `n <= pool->size()` is safe.
- **Shrink while threads are rented is safe by design** — rented `ThreadSlot`s are held via `shared_ptr`; shrinking drops the pool's reference but the renter keeps the slot (and its OS thread) alive until `~RentedThreads` releases it. Resize never blocks on in-flight operations.

### Other invariants

- **Pool is always valid** — the shared pool singleton is initialized with size 1 on first access (`VecSimSVSThreadPoolImpl::instance()`). There is no `nullptr` state; write-in-place mode uses a pool of size 1 (0 worker threads, calling thread only).
- **`parallelism_ >= 1`** — the calling thread always participates, so parallelism can never be 0.
- **Rental is safe during resize** — the deferred-resize protocol (`pending_jobs_`) prevents the pool from shrinking while scheduled jobs are in flight. Since all multi-threaded SVS operations run within scheduled jobs, rented `ThreadSlot` raw pointers are guaranteed to remain valid for the lifetime of `RentedThreads`.
- **No cross-index state leakage** — each index has its own `shared_ptr<atomic<size_t>> parallelism_`; `setParallelism()` on one index never affects another.

### Relationship with RediSearch

- RediSearch **owns the scheduling** — it decides when to dispatch SVS jobs and how many workers are available.
- VecSim **owns the threads** — the SVS thread pool is a separate set of OS threads inside VecSim, distinct from RediSearch workers.
- RediSearch's only responsibility toward the SVS pool is calling `VecSim_UpdateThreadPoolSize(N)` when the worker count changes, so VecSim can create/destroy the matching number of OS threads.
- The two pools are always the same size: RediSearch workers = N, SVS threads = N. This 1:1 sizing is what makes the "sum of rents ≤ capacity" guarantee work.

## Changes from Old Implementation

| Aspect | Old (per-index pools) | New (shared pool + rental) |
|---|---|---|
| **Pool ownership** | Each SVS index creates its own `VecSimSVSThreadPoolImpl` | Singleton `VecSimSVSThreadPoolImpl` shared across all indexes via `shared_ptr` |
| **Thread lifecycle** | Pre-allocated at max capacity; `resize()` clamped a logical counter | Physical grow/shrink: `resize()` spawns/destroys OS threads |
| **Pool could be null** | Yes (write-in-place = no pool) | Never null; write-in-place = pool with size 1 |
| **`parallel_for` threading** | Threads used directly from the per-index pool | Threads **rented** from shared pool under mutex as raw pointers, released lock-free via `atomic<bool>` |
| **Concurrent index operations** | Each index had dedicated threads (K×N total) | Indexes rent disjoint subsets from N shared threads |
| **Per-index parallelism control** | `setNumThreads(n)` / `resize(n)` — mutated the pool | `setParallelism(n)` — sets a per-index `atomic<size_t>`, pool unchanged |
| **Pool size query** | `capacity()` (max pre-allocated) | `poolSize()` (current shared pool size) |
| **Error handling** | `manage_exception_during_run()` threw unconditionally | `manage_workers_after_run()` only throws if an error occurred |

### Renamed methods

| Old | New | Scope |
|---|---|---|
| `setNumThreads(n)` | `setParallelism(n)` | `SVSIndex`, `SVSIndexBase` virtual interface |
| `getNumThreads()` | `getParallelism()` | `SVSIndex`, `SVSIndexBase` virtual interface |
| `getThreadPoolCapacity()` | `getPoolSize()` | `SVSIndex`, `SVSIndexBase` virtual interface |

### Removed methods
| Method | Reason |
|---|---|
| `VecSimSVSThreadPool::capacity()` | No pre-allocation; use `poolSize()` for shared pool size |
| `VecSimSVSThreadPool::resize()` | Wrappers don't own threads; use `setParallelism()` for per-index control, `VecSimSVSThreadPool::resize(size_t)` (static) for global pool |

### New C API

```c
// Combines write-mode toggling + physical SVS pool resize.
// new_size == 0 → WriteInPlace mode, pool resized to 1
// new_size >  0 → WriteAsync mode, pool resized to new_size
void VecSim_UpdateThreadPoolSize(size_t new_size);
```

### New internal APIs

| API | Purpose |
|---|---|
| `VecSimSVSThreadPoolImpl::instance()` | Static singleton accessor; returns `shared_ptr<VecSimSVSThreadPoolImpl>` by value |
| `VecSimSVSThreadPool(void* log_ctx)` | Constructor — uses shared pool singleton internally |
| `VecSimSVSThreadPool::resize(size_t)` | Static method — resizes the shared pool singleton |
| `VecSimSVSThreadPool::poolSize()` | Returns shared pool size (for scheduling reserve jobs) |
| `VecSimSVSThreadPoolImpl::rent(count, log_ctx)` | Rents worker threads with RAII guard (`RentedThreads`) |

### Deprecated

| API | Replacement | Notes |
|---|---|---|
| `VecSim_SetWriteMode()` | `VecSim_UpdateThreadPoolSize()` | Kept for internal VecSim usage (called inside `VecSim_UpdateThreadPoolSize`) and unit tests. Not called by RediSearch — RediSearch will use only `VecSim_UpdateThreadPoolSize` |
| `SVSParams::num_threads` | Shared pool singleton | Deprecated and **ignored**. Setting it to a non-zero value emits a deprecation warning log. Kept for backward compatibility with existing callers/tests. Pool size is controlled globally via `VecSim_UpdateThreadPoolSize()` |

### New types

- **`ThreadSlot`** — wraps `svs::threads::Thread` + `atomic<bool> occupied`; non-copyable, non-movable
- **`RentedThreads`** — RAII guard; move-only; holds raw `ThreadSlot*` pointers (no `shared_ptr` overhead); releases slots via lock-free atomic stores in destructor
- **Deferred resize lifecycle** — managed via a `bool isScheduled` flag on `SVSMultiThreadJob`; its destructor calls `endScheduledJob()` on the shared pool (no separate `ScheduledJobToken` class)

## Behavioral Changes

| Area | Old behavior | New behavior |
|---|---|---|
| `setParallelism(0)` | `resize(0)` silently clamped to 1 | Asserts (debug) / undefined (release) — reserving 0 threads is a bug |
| `setParallelism(n > poolSize)` | Grew the per-index pool beyond `search-workers` | Asserts `n <= pool->size()` — scheduling bug |
| `n > parallelism` in `parallel_for` | Threw `ThreadingException` if `n > size_` | Allowed (no assertion) — supports deferred resize scenarios where the pool may have shrunk after job scheduling |
| `parallelism_` initial value | Implicitly matched pool size at construction | Starts at **1** — safe for immediate write-in-place use |
| Fewer threads rented than requested | N/A (no rental) | Logs a warning via `logCallback`, asserts in debug, uses `rented.count()` for graceful degradation in release |
| Exception in `parallel_for` | `manage_exception_during_run()` always threw | `manage_workers_after_run()` only throws if an error actually occurred |
| `updateSVSIndexWrapper` | Always locked backend even if no vectors to move | Skips lock when `labels_to_move` is empty |
| `SVSParams::num_threads` set to non-zero | Used as per-index pool size | Ignored; logs a deprecation warning |

### Public info API fields — unchanged

The debug info field names `numThreads` and `lastReservedThreads` (and their string keys `NUM_THREADS`, `LAST_RESERVED_NUM_THREADS`) are **unchanged**. Their semantics shift slightly:
- `numThreads` → now reports `getPoolSize()` (shared pool size, not per-index capacity)
- `lastReservedThreads` → now reports `getParallelism()` (per-index parallelism for last operation)

## What Does NOT Change

- **ScalableVectorSearch** (`deps/ScalableVectorSearch/`) — no modifications
- **`SVSMultiThreadJob` pattern** — still submits N reserve jobs to the external worker pool
- **HNSW tiered indexes** — unaffected
- **SVS `Thread` state machine** — used as-is; lifecycle managed via `shared_ptr`

## Files Changed

| File | Changes |
|---|---|
| `src/VecSim/algorithms/svs/svs_utils.h` | Core refactor: `ThreadSlot`, `RentedThreads`, rental-based `VecSimSVSThreadPoolImpl`, per-index `VecSimSVSThreadPool` wrapper |
| `src/VecSim/algorithms/svs/svs.h` | Singleton accessor, API renames, constructor takes shared pool, `num_threads` deprecation warning |
| `src/VecSim/algorithms/svs/svs_tiered.h` | API renames, empty-labels guard in `updateSVSIndexWrapper`, `SVSMultiThreadJob` uses `bool isScheduled` + destructor-based `endScheduledJob()` instead of `ScheduledJobToken`, `createJobsImpl` merged into `createJobs` |
| `src/VecSim/vec_sim.h` / `vec_sim.cpp` | New `VecSim_UpdateThreadPoolSize()` C API |
| `src/VecSim/vec_sim_common.h` | `SVSParams::num_threads` marked as deprecated in comments |
| `tests/unit/test_svs_tiered.cpp` | Adapted to shared pool: `VecSim_UpdateThreadPoolSize` in setup, removed `num_threads` from params, updated assertions |
| `tests/unit/test_svs_fp16.cpp` | Same test adaptations |
| `tests/unit/test_svs.cpp` | `numThreads` info field expects default; new `NumThreadsParamIgnored` test |
| `tests/benchmark/bm_utils.h` / `bm_vecsim_svs.h` | API renames in benchmark helpers |

## Testing

- All existing SVS tiered tests updated to use the shared pool model
- `testThreadPool` rewritten to cover: default parallelism, `setParallelism` boundary assertions (`ASSERT_DEATH`), `parallel_for` with `n < parallelism`, write-in-place mode, exception handling
- `TestDebugInfoThreadCount` / `TestDebugInfoThreadCountWriteInPlace` updated: `lastReservedThreads` now starts at 1 (not `num_threads`)
- `ThreadsReservation` test uses `tieredIndexMock(num_threads)` to properly size the shared pool
- **New: `NumThreadsParamIgnored`** — verifies that setting `SVSParams::num_threads` emits a deprecation warning, does not affect the shared pool size, and that omitting it produces no warning
- Tests that previously set `num_threads = 1` in `SVSParams` now use `tieredIndexMock(1)` to resize the shared pool instead

**Mark if applicable**

---

> [!NOTE]
> **High Risk**
> High risk due to substantial changes to SVS concurrency, thread scheduling, and resize semantics that affect async update/GC execution and can introduce deadlocks or race conditions if invariants are violated.
> 
> **Overview**
> Refactors SVS threading from *per-index pools* to a **shared singleton thread pool** that can physically grow/shrink and **rent worker threads** per `parallel_for` call, including a deferred-shrink mechanism to avoid invalidating scheduled jobs.
> 
> Renames the SVS per-operation knob from `get/setNumThreads` to `get/setParallelism`, updates tiered SVS update/GC scheduling to create *scheduled* jobs that snapshot pool size, and adjusts debug info to report **shared pool size** (`numThreads`) and **last reserved parallelism**.
> 
> Adds `VecSim_UpdateThreadPoolSize()` to atomically toggle write mode (async vs in-place) and resize the shared SVS pool, marks `SVSParams::num_threads` as deprecated/ignored (with warning log), and updates benchmarks/unit tests including new coverage for resize/rental/deferred-shrink behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f5d65fd6f308539ee4113bfa8021bafd7178c848. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/VecSim/algorithms/svs/svs.h
+++ b/src/VecSim/algorithms/svs/svs.h
@@ -41,7 +41,7 @@ struct SVSIndexBase
     // Singleton accessor for the shared SVS thread pool.
     // Always valid — initialized with size 1 (write-in-place mode: 0 worker threads,
     // only the calling thread participates). Resized on VecSim_UpdateThreadPoolSize() calls.
-    static std::shared_ptr<VecSimSVSThreadPoolImpl> &getSharedThreadPool() {
+    static std::shared_ptr<VecSimSVSThreadPoolImpl> getSharedThreadPool() {
         static auto shared_pool = std::make_shared<VecSimSVSThreadPoolImpl>(1);
         return shared_pool;
     }
@@ -369,7 +369,7 @@ public:
               svs_details::getOrDefault(params.leanvec_dim, SVS_VAMANA_DEFAULT_LEANVEC_DIM)},
           epsilon{svs_details::getOrDefault(params.epsilon, SVS_VAMANA_DEFAULT_EPSILON)},
           is_two_level_lvq{isTwoLevelLVQ(params.quantBits)},
-          threadpool_{SVSIndexBase::getSharedThreadPool()}, impl_{nullptr} {
+          threadpool_{SVSIndexBase::getSharedThreadPool(), this->logCallbackCtx}, impl_{nullptr} {
         logger_ = makeLogger();
     }
 
--- a/src/VecSim/algorithms/svs/svs.h
+++ b/src/VecSim/algorithms/svs/svs.h
@@ -39,10 +39,10 @@ struct SVSIndexBase
     virtual ~SVSIndexBase() = default;
 
     // Singleton accessor for the shared SVS thread pool.
-    // Returns a reference to the static shared_ptr, which is initially null.
-    // Created lazily on first VecSim_UpdateThreadPoolSize(n > 0) call.
+    // Always valid — initialized with size 1 (write-in-place mode: 0 worker threads,
+    // only the calling thread participates). Resized on VecSim_UpdateThreadPoolSize() calls.
     static std::shared_ptr<VecSimSVSThreadPoolImpl> &getSharedThreadPool() {
-        static std::shared_ptr<VecSimSVSThreadPoolImpl> shared_pool;
+        static auto shared_pool = std::make_shared<VecSimSVSThreadPoolImpl>(1);
         return shared_pool;
     }
     virtual int addVectors(const void *vectors_data, const labelType *labels, size_t n) = 0;
@@ -369,8 +369,7 @@ public:
               svs_details::getOrDefault(params.leanvec_dim, SVS_VAMANA_DEFAULT_LEANVEC_DIM)},
           epsilon{svs_details::getOrDefault(params.epsilon, SVS_VAMANA_DEFAULT_EPSILON)},
           is_two_level_lvq{isTwoLevelLVQ(params.quantBits)},
-          threadpool_{std::max(size_t{SVS_VAMANA_DEFAULT_NUM_THREADS}, params.num_threads)},
-          impl_{nullptr} {
+          threadpool_{SVSIndexBase::getSharedThreadPool()}, impl_{nullptr} {
         logger_ = makeLogger();
     }
 
@@ -549,10 +548,10 @@ public:
         return impl_ ? impl_->has_id(label) : false;
     }
 
-    size_t getParallelism() const override { return threadpool_.size(); }
-    void setParallelism(size_t parallelism) override { threadpool_.resize(parallelism); }
+    size_t getParallelism() const override { return threadpool_.getParallelism(); }
+    void setParallelism(size_t parallelism) override { threadpool_.setParallelism(parallelism); }
 
-    size_t getPoolSize() const override { return threadpool_.capacity(); }
+    size_t getPoolSize() const override { return threadpool_.poolSize(); }
 
     bool isCompressed() const override { return storage_traits_t::is_compressed(); }
 
--- a/src/VecSim/algorithms/svs/svs.h
+++ b/src/VecSim/algorithms/svs/svs.h
@@ -371,6 +371,11 @@ public:
           is_two_level_lvq{isTwoLevelLVQ(params.quantBits)},
           threadpool_{SVSIndexBase::getSharedThreadPool(), this->logCallbackCtx}, impl_{nullptr} {
         logger_ = makeLogger();
+        if (params.num_threads != 0) {
+            this->log(VecSimCommonStrings::LOG_WARNING_STRING,
+                      "SVSParams.num_threads is deprecated and ignored. "
+                      "Thread pool size is controlled globally via VecSim_UpdateThreadPoolSize().");
+        }
     }
 
     ~SVSIndex() = default;
--- a/src/VecSim/algorithms/svs/svs.h
+++ b/src/VecSim/algorithms/svs/svs.h
@@ -38,13 +38,6 @@ struct SVSIndexBase
     SVSIndexBase() : num_marked_deleted{0} {};
     virtual ~SVSIndexBase() = default;
 
-    // Singleton accessor for the shared SVS thread pool.
-    // Always valid — initialized with size 1 (write-in-place mode: 0 worker threads,
-    // only the calling thread participates). Resized on VecSim_UpdateThreadPoolSize() calls.
-    static std::shared_ptr<VecSimSVSThreadPoolImpl> getSharedThreadPool() {
-        static auto shared_pool = std::make_shared<VecSimSVSThreadPoolImpl>(1);
-        return shared_pool;
-    }
     virtual int addVectors(const void *vectors_data, const labelType *labels, size_t n) = 0;
     virtual int deleteVectors(const labelType *labels, size_t n) = 0;
     virtual bool isLabelExists(labelType label) const = 0;
@@ -368,8 +361,8 @@ public:
           leanvec_dim{
               svs_details::getOrDefault(params.leanvec_dim, SVS_VAMANA_DEFAULT_LEANVEC_DIM)},
           epsilon{svs_details::getOrDefault(params.epsilon, SVS_VAMANA_DEFAULT_EPSILON)},
-          is_two_level_lvq{isTwoLevelLVQ(params.quantBits)},
-          threadpool_{SVSIndexBase::getSharedThreadPool(), this->logCallbackCtx}, impl_{nullptr} {
+          is_two_level_lvq{isTwoLevelLVQ(params.quantBits)}, threadpool_{this->logCallbackCtx},
+          impl_{nullptr} {
         logger_ = makeLogger();
         if (params.num_threads != 0) {
             this->log(VecSimCommonStrings::LOG_WARNING_STRING,
--- a/src/VecSim/algorithms/svs/svs.h
+++ b/src/VecSim/algorithms/svs/svs.h
@@ -41,9 +41,9 @@ struct SVSIndexBase
     virtual int deleteVectors(const labelType *labels, size_t n) = 0;
     virtual bool isLabelExists(labelType label) const = 0;
     virtual size_t indexStorageSize() const = 0;
-    virtual size_t getNumThreads() const = 0;
-    virtual void setNumThreads(size_t numThreads) = 0;
-    virtual size_t getThreadPoolCapacity() const = 0;
+    virtual size_t getParallelism() const = 0;
+    virtual void setParallelism(size_t parallelism) = 0;
+    virtual size_t getPoolSize() const = 0;
     virtual bool isCompressed() const = 0;
     size_t getNumMarkedDeleted() const { return num_marked_deleted; }
 
@@ -66,9 +66,9 @@ protected:
 };
 
 /** Thread Management Strategy:
- * - addVector(): Requires numThreads == 1
- * - addVectors(): Allows any numThreads value, but prohibits n=1 with numThreads>1
- * - Callers are responsible for setting appropriate thread counts
+ * - addVector(): Requires parallelism == 1
+ * - addVectors(): Allows any parallelism value, but prohibits n=1 with parallelism>1
+ * - Callers are responsible for setting appropriate parallelism
  **/
 template <typename MetricType, typename DataType, bool isMulti, size_t QuantBits,
           size_t ResidualBits, bool IsLeanVec>
@@ -251,12 +251,12 @@ protected:
         this->impl_ = std::move(svs_handler->impl);
     }
 
-    // Assuming numThreads was updated to reflect the number of available threads before this
+    // Assuming parallelism was updated to reflect the number of available threads before this
     // function was called.
-    // This function assumes that the caller has already set numThreads to the appropriate value
+    // This function assumes that the caller has already set parallelism to the appropriate value
     // for the operation.
-    // Important NOTE: For single vector operations (n=1), numThreads should be 1.
-    // For bulk operations (n>1), numThreads should reflect the number of available threads.
+    // Important NOTE: For single vector operations (n=1), parallelism should be 1.
+    // For bulk operations (n>1), parallelism should reflect the number of available threads.
     int addVectorsImpl(const void *vectors_data, const labelType *labels, size_t n) {
         if (n == 0) {
             return 0;
@@ -412,8 +412,8 @@ public:
                           .maxCandidatePoolSize = this->buildParams.max_candidate_pool_size,
                           .pruneTo = this->buildParams.prune_to,
                           .useSearchHistory = this->buildParams.use_full_search_history,
-                          .numThreads = this->getThreadPoolCapacity(),
-                          .lastReservedThreads = this->getNumThreads(),
+                          .numThreads = this->getPoolSize(),
+                          .lastReservedThreads = this->getParallelism(),
                           .numberOfMarkedDeletedNodes = this->num_marked_deleted,
                           .searchWindowSize = this->search_window_size,
                           .searchBufferCapacity = this->search_buffer_capacity,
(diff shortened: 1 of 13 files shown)


src/VecSim/algorithms/svs/svs.h
```diff
@@ -37,13 +37,21 @@ struct SVSIndexBase
 {
     SVSIndexBase() : num_marked_deleted{0} {};
     virtual ~SVSIndexBase() = default;
+
+    // Singleton accessor for the shared SVS thread pool.
+    // Always valid — initialized with size 1 (write-in-place mode: 0 worker threads,
+    // only the calling thread participates). Resized on VecSim_UpdateThreadPoolSize() calls.
+    static std::shared_ptr<VecSimSVSThreadPoolImpl> getSharedThreadPool() {
```
rfsaliev: It seems like the `VecSimSVSThreadPoolImpl` is expected to be always used as a singleton.
If this assumption is true, then I would suggest to use more generic idiom:
```cpp
class VecSimSVSThreadPoolImpl {
...
private:
    explicit VecSimSVSThreadPoolImpl(size_t num_threads = 1) {...}
public:
    static std::shared_ptr<VecSimSVSThreadPoolImpl> instance() {...}
...
};
```
 
1. Such idiom prevents non-singleton usages.
2. Moves TP lifetime managing responsibilities out of SVSIndex class

Moreover, I would move the full ownership to `VecSimSVSThreadPool` with possible hiding `VecSimSVSThreadPoolImpl` to the `details` namespace:
```cpp
class VecSimSVSThreadPool {
...
public:
    explicit VecSimSVSThreadPool(void *log_ctx = nullptr)
        : pool_(details::VecSimSVSThreadPoolImpl::instance()),
          parallelism_(std::make_shared<std::atomic<size_t>>(1)),
          log_ctx_(log_ctx) {
        assert(pool_ && "Pool must not be null");
    }

    static auto resize(size_t new_size) {
        return details::VecSimSVSThreadPoolImpl::instance()->resize(new_size);
    }
...
};

File: src/VecSim/algorithms/svs/svs_utils.h [hunk 2]
```diff
@@ -341,110 +344,285 @@ struct SVSGraphBuilder {
...
+            assert(false && "Failed to rent the expected number of SVS threads");
         }
+        return rented;
     }
 
-    void manage_exception_during_run(const std::string &thread_0_message = {}) {
+    // Wait for all rented workers to finish. If any worker (or the main thread) threw,
+    // restart crashed workers and throw a combined exception.
+    void manage_workers_after_run(const std::string &main_thread_error, RentedThreads &rented,
+                                  size_t rented_count) {
```
rfsaliev: Why there are 3 arguments if `rented.count() == rented_count` here?

rfsaliev (dismissed):
LGFM, but some suggestions.

src/VecSim/algorithms/svs/svs_utils.h
meiravgri: no good reason :)

rfsaliev (dismissed):
LGFM

src/VecSim/algorithms/svs/svs_utils.h
```diff
@@ -341,110 +344,293 @@ struct SVSGraphBuilder {
...
+// RAII guard for threads rented from the shared pool. On destruction, marks all
+// rented slots as unoccupied (lock-free atomic stores). Holds shared_ptr references
+// to keep slots alive even if the pool shrinks while the rental is active.
+class RentedThreads {
+public:
+    RentedThreads() = default;
+
+    // Move-only
+    RentedThreads(RentedThreads &&other) noexcept : slots_(std::move(other.slots_)) {}
+    RentedThreads &operator=(RentedThreads &&other) noexcept {
```
dor-forer: Is the move operator being used?

src/VecSim/algorithms/svs/svs_utils.h
```diff
@@ -341,110 +344,293 @@ struct SVSGraphBuilder {
...
+    ThreadSlot(ThreadSlot &&) = delete;
+    ThreadSlot &operator=(ThreadSlot &&) = delete;
+};
+
+// RAII guard for threads rented from the shared pool. On destruction, marks all
+// rented slots as unoccupied (lock-free atomic stores). Holds shared_ptr references
+// to keep slots alive even if the pool shrinks while the rental is active.
+class RentedThreads {
+public:
+    RentedThreads() = default;
```
dor-forer: Needs to be private or should also be used outside of `rent()`?

src/VecSim/algorithms/svs/svs_utils.h
```diff
@@ -341,110 +344,293 @@ struct SVSGraphBuilder {
...
+    RentedThreads(const RentedThreads &) = delete;
+    RentedThreads &operator=(const RentedThreads &) = delete;
+
+    ~RentedThreads() { release(); }
+
+    void add(std::shared_ptr<ThreadSlot> slot) { slots_.push_back(std::move(slot)); }
+
+    size_t count() const { return slots_.size(); }
+
+    svs::threads::Thread &operator[](size_t i) { return slots_[i]->thread; }
```
dor-forer: Should we add and aseertion on the i values or the outside assertion is enough?

src/VecSim/algorithms/svs/svs_utils.h
```diff
@@ -341,110 +344,293 @@ struct SVSGraphBuilder {
...
-            // Run on the main function.
-            try {
-                f(0);
-            } catch (const std::exception &error) {
-                manage_exception_during_run(error.what());
+            bool expected = false;
+            if (slot->occupied.compare_exchange_strong(expected, true, std::memory_order_acq_rel)) {
+                rented.add(slot);
             }
+        }
```
dor-forer, line 546: Can be checked one time outside and use a local var for faster loop:
```
if (rented.count() < count) {
    int rent = 0;
    for (auto &slot : slots_) {
        bool expected = false;
        if (slot->occupied.compare_exchange_strong(expected, true, std::memory_order_acq_rel)) {
            rented.add(slot);
            if (++rent >= count) {
                break;
            }
        }
    }
}
```

src/VecSim/algorithms/svs/svs_utils.h
meiravgri: no

src/VecSim/algorithms/svs/svs_utils.h
meiravgri: i moved the entire class to the VecSimSVSThreadPoolImpl

src/VecSim/algorithms/svs/svs_utils.h
meiravgri: we can

tests/unit/test_svs_tiered.cpp
```diff
@@ -90,30 +87,23 @@ class SVSTieredIndexTest : public ::testing::Test {
                                                    .updateJobWaitTime = update_job_wait_time}}};
     }
 
-    void verifyNumThreads(TieredSVSIndex<data_t> *tiered_index, size_t expected_num_threads,
-                          size_t expected_capcity) {
-        ASSERT_EQ(tiered_index->GetSVSIndex()->getNumThreads(), expected_num_threads);
-        ASSERT_EQ(tiered_index->GetSVSIndex()->getThreadPoolCapacity(), expected_capcity);
+    void verifyNumThreads(TieredSVSIndex<data_t> *tiered_index, size_t expected_capcity) {
+        ASSERT_EQ(tiered_index->GetSVSIndex()->getParallelism(), 1);
+        ASSERT_EQ(tiered_index->GetSVSIndex()->getPoolSize(), expected_capcity);
```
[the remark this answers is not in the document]
meiravgri: verifyNumThreads` helper is only called from `CreateTieredSVSIndex` in the test fixtures, and no test passes `thread_pool_size == 0` through that path. The benchmark's `CreateTieredSVSIndex` (line 110-111) already applies `std::max(mock_thread_pool.thread_pool_size, size_t{1})` to handle the `tieredIndexMock(0)` case correctly. If a future unit test ever passes 0, the assertion failure would be a useful signal that this is invalid input, rather than something we should silently clamp.

src/VecSim/algorithms/svs/svs_utils.h
```diff
@@ -436,24 +438,44 @@ class VecSimSVSThreadPoolImpl {
     // on shrink. new_size is total parallelism including the calling thread (minimum 1).
     // Occupied threads (held by renters) survive shrink via shared_ptr — their OS thread
```
dor-forer: Not so important but I think it's not a shared_ptr anymore

src/VecSim/algorithms/svs/svs_utils.h
```diff
@@ -360,49 +362,49 @@ struct ThreadSlot {
...
+    // pool cannot shrink (destroy slots) while scheduled jobs are in flight, and all
+    // multi-threaded SVS operations run within scheduled jobs.
+    class RentedThreads {
+    public:
+        RentedThreads() = default;
 
-    ~RentedThreads() { release(); }
+        // Move-only
+        RentedThreads(RentedThreads &&other) noexcept : slots_(std::move(other.slots_)) {}
+        RentedThreads(const RentedThreads &) = delete;
```
dor-forer, line 384: I would delete the 
```
RentedThreads &operator=(const RentedThreads &) = delete;
RentedThreads &operator=(RentedThreads &&) = delete;
```
Just for completeness and to make sure nobody uses it implicitly when no needed.

dor-forer (approved):

PR #928, opened 2026-04-06
Extend SVS_URL logic with new binaries

memark:
Extend SVS_URL logic with new binaries from Intel for older systems, in order to enable LTO compilation between C and Rust for these platforms in RediSearch.

The binaries are currently named "nightly", but those links are permanent. They will be renamed in a coming SVS release.

RediSearch workflow run pointing to this commit
https://github.com/RediSearch/RediSearch/actions/runs/24208329013

---

> [!NOTE]
> **Medium Risk**
> Build-time logic now runs compiled probes to select SVS prebuilt binaries and may change which artifacts (or source builds) are used on different Linux toolchains, potentially breaking CI or user builds if detection/matrix assumptions are wrong.
> 
> **Overview**
> Updates `cmake/svs.cmake` to **replace compile-and-run GLIBC feature checks** with `try_run` probes that capture the exact detected `GLIBC` version and `_GLIBCXX_RELEASE` (libstdc++ header/GCC major) at configure time.
> 
> Extends `SVS_URL` selection to support **new Intel nightly SVS shared-library binaries for Clang 21** (with GLIBC >= 2.31) and chooses the artifact based on detected libstdc++ release (GCC 11/12/13); otherwise it **disables `SVS_SHARED_LIB`** with updated gating messages. GCC-path selection is also adjusted to use the detected GLIBC version string (>=2.28, fallback >=2.26) when picking existing v0.3.0 binaries.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 88e5ff2c88627a7cb891878cf26ec753a1e83a8c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmake/svs.cmake
```diff
@@ -37,35 +37,54 @@ if(USE_SVS)
...
-            else()
-                set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/v0.3.0/svs-shared-library-0.3.0-reduced.tar.gz" CACHE STRING "SVS URL")
-	    endif()
-        elseif(GLIBC_2_26_FOUND)
+        if(GLIBC_VERSION VERSION_GREATER_EQUAL "2.28")
+            if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "14.0")
+                    set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/v0.3.0/svs-shared-library-0.3.0-reduced-gcc14.tar.gz" CACHE STRING "SVS URL")
+                else()
+                    set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/v0.3.0/svs-shared-library-0.3.0-reduced.tar.gz" CACHE STRING "SVS URL")
+            endif()
```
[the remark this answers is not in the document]
memark: Fixed.

cmake/svs.cmake
```diff
@@ -37,35 +37,54 @@ if(USE_SVS)
...
+        endif()
+        message(STATUS "Detected GLIBCXX_RELEASE (GCC major): ${GLIBCXX_VERSION}")
     endif()
 
     cmake_dependent_option(SVS_SHARED_LIB "Use SVS pre-compiled shared library" ON "USE_SVS AND GLIBC_FOUND AND SVS_LVQ_SUPPORTED" OFF)
     if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-        if (GLIBC_2_28_FOUND)
-            if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "20.0")
-                set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/v0.3.0/svs-shared-library-0.3.0-reduced-clang20.tar.gz" CACHE STRING "SVS URL")
+        if(GLIBC_VERSION VERSION_GREATER_EQUAL "2.29")
```
ethanglaser: A couple comments about this condition:
1. This build comes from an ubuntu20 dockerfile with glibc 2.31, so unless you have tested it on glibc 2.29/2.30 or believe it would work on these, it should probably be set to 2.31. If you require support for glibc2.29 let me know and I can look into revising our build env.
2. The `CMAKE_CXX_COMPILER_VERSION` check is removed but I don't expect these builds would work on lower than clang-21 (though I havent tried). Perhaps this should be added in with the message in the else condition revised to mention required clang version

Overall I would expect the condition would look like:

```suggestion
        if(GLIBC_VERSION VERSION_GREATER_EQUAL "2.31" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "21.0")
```

But these are just assumptions based on previous behavior, I haven't tested it out

memark: 1. I extracted the symbols from the .so in each archive, then looked at the highest GLIBC version mentioned. Those were 2.29 in all of the archives. But that could possibly change if you include newer symbols in your code. We don't have any targets lower than 2.31 anyway, so we might wanna change it back to that?

memark: 2. Ah, I see. I removed that check when we didn't need to differenciate between 20 and 21 anymore. But it might sense to add it back.

memark: Pushed new version.

ethanglaser (approved):

GuyAv46 (approved):

ethanglaser:
Actually one more thought from my side before merging: we are providing a single shared library build for standard gcc11 (svs-shared-library-0.3.0-reduced.tar.gz) which is meant to work with gcc11-13 - I believe the same breadth of support would be applicable here. Could you check if the clang21-gcc11 build works for gcc12 and gcc13 as well? Unless it breaks stuff (but I believe the LTO strictness would only apply to clang version) or performs worse, it would be great to consolidate to a single library

PR #972, opened 2026-05-26
MOD-15578 Track shared SVS thread pool memory & expose it through public API

dor-forer:
The shared `VecSimSVSThreadPool` singleton was previously created via raw `new` with the default allocator, so its slot vector and per-slot `ThreadSlot` objects bypassed `VecSimAllocator` and were invisible to any memory accounting downstream (`FT.INFO`, `INFO MODULES`, etc.).

This PR:

1. Routes the shared SVS thread pool's allocations through a dedicated `VecSimAllocator` (using `VecsimSTLAllocator` for the slot vector and `std::allocate_shared` for thread objects).
2. Adds a new public API `size_t VecSim_GetSharedMemory(void)` returning the total bytes of process-wide VecSim allocations not tied to any single index — today equal to the shared SVS thread pool's tracked allocation size.
3. Surfaces these bytes in `VECSIM_INFO` via one new field:
   - **`SHARED_MEMORY`** — appended at the top level of every algorithm's debug response by the C API wrapper `VecSimIndex_DebugInfoIterator`. Always present (value may be 0).

> **Note on a single field.** An earlier revision also emitted a per-contributor `SHARED_SVS_THREADPOOL_MEMORY` breakdown inside SVS sections. Since the SVS thread pool is the only process-wide contributor today, that field always reported the exact same bytes as `SHARED_MEMORY` and carried no extra information, so it was dropped (YAGNI). If a second global contributor is ever added, an itemized breakdown can be reintroduced then with real differentiation.

---

## Public API change

**Before**
```c
// (no API to query process-wide VecSim memory)
```

**After**
```c
// Returns total bytes of process-wide VecSim allocations not tied to any single
// index (today: the shared SVS thread pool singleton). Returns 0 when no such
// shared allocations exist.
size_t VecSim_GetSharedMemory(void);
```

Callers (e.g. RediSearch) can fold this into per-spec or process-wide memory metrics without depending on which algorithm contributes.

---

### Common header (every algorithm, unchanged)
```
ALGORITHM, TYPE, DIMENSION, METRIC, IS_MULTI_VALUE, IS_DISK,
INDEX_SIZE, INDEX_LABEL_COUNT, MEMORY, LAST_SEARCH_MODE
```

### FLAT — 11 → **12 fields**
```diff
  <common header × 10>
  BLOCK_SIZE
+ SHARED_MEMORY
```

### HNSW — 18 → **19 fields**
```diff
  <common header × 10>
  BLOCK_SIZE, M, EF_CONSTRUCTION, EF_RUNTIME, MAX_LEVEL, ENTRYPOINT,
  EPSILON, NUMBER_OF_MARKED_DELETED
+ SHARED_MEMORY
```

### SVS (non-tiered) — 25 → **26 fields**
```diff
  <common header × 10>
  BLOCK_SIZE, QUANT_BITS, ALPHA, GRAPH_MAX_DEGREE, CONSTRUCTION_WINDOW_SIZE,
  MAX_CANDIDATE_POOL_SIZE, PRUNE_TO, USE_SEARCH_HISTORY, NUM_THREADS,
  LAST_RESERVED_NUM_THREADS, NUMBER_OF_MARKED_DELETED, SEARCH_WINDOW_SIZE,
  SEARCH_BUFFER_CAPACITY, LEANVEC_DIMENSION, EPSILON
+ SHARED_MEMORY
```

### Tiered HNSW — 16 → **17 fields**
```diff
  <common header × 10>           (ALGORITHM = "TIERED")
  MANAGEMENT_LAYER_MEMORY, BACKGROUND_INDEXING, TIERED_BUFFER_LIMIT
  FRONTEND_INDEX = nested [<FLAT fields, 11>]      (no SHARED_MEMORY in nested)
  BACKEND_INDEX  = nested [<HNSW fields, 18>]      (no SHARED_MEMORY in nested)
  TIERED_HNSW_SWAP_JOBS_THRESHOLD
+ SHARED_MEMORY
```

### Tiered SVS — 18 → **19 fields**
```diff
  <common header × 10>           (ALGORITHM = "TIERED")
  MANAGEMENT_LAYER_MEMORY, BACKGROUND_INDEXING, TIERED_BUFFER_LIMIT
  FRONTEND_INDEX = nested [<FLAT fields, 11>]      (no SHARED_MEMORY in nested)
  BACKEND_INDEX  = nested [<SVS fields, 25>]       (no SHARED_MEMORY in nested)
  TIERED_SVS_TRAINING_THRESHOLD, TIERED_SVS_UPDATE_THRESHOLD,
  TIERED_SVS_THREADS_RESERVE_TIMEOUT
+ SHARED_MEMORY
```

### Emission rule

**`SHARED_MEMORY`** is appended exactly once, at the **outermost** iterator level, by the C API wrapper `VecSimIndex_DebugInfoIterator`. It never appears inside a nested `FRONTEND_INDEX`/`BACKEND_INDEX`.

---

### `VecSim_GetSharedMemory()`

**Before**: API did not exist. The pool's slot vector and per-slot `ThreadSlot` objects went through the default allocator and were not tracked anywhere.

**After**:
```c
size_t bytes = VecSim_GetSharedMemory();
// == VecSimSVSThreadPool::getSharedAllocationSize()
//    (slot vector + per-slot ThreadSlot objects, allocated through the
//     dedicated VecSimAllocator).
// == 0 before any SVS index has been constructed and the worker pool size set.
```

### Per-index `getAllocationSize()` (SVS only)

**Before**: Did not include any per-index portion of the parallelism slot, since the pool was untracked entirely.

**After**: Each SVS index's per-index allocator now tracks its own `parallelism` slot (a small fixed-size structure inside the index, allocated through the index's `VecSimAllocator`). The shared pool itself remains process-wide and reported via `VecSim_GetSharedMemory()`.

---

## Tests

- Added `SVSTest.debugInfoSharedMemoryMatchesApi` (typed test, runs per SVS data type) — asserts the `SHARED_MEMORY` field appears exactly once in a non-tiered SVS response and reports the same bytes as `VecSim_GetSharedMemory()`.
- Added `SVSTest.sharedMemoryTracksThreadPoolResize` (typed test) — asserts the reported shared memory actually **grows** when the pool is resized up and **shrinks** when resized down, so the value tracks real allocation rather than being a constant that happens to agree with the API.
- Updated `compareFlatIndexInfoToIterator`, `compareHNSWIndexInfoToIterator`, `compareSVSIndexInfoToIterator` to take an `expect_shared_memory` parameter (default `true`) — needed because these comparators are called both with the C API iterator (top level, has `SHARED_MEMORY`) and as nested-backend comparators inside `compareTieredIndexInfoToIterator` (no `SHARED_MEMORY`).
- Expected field counts (C++ method, no `SHARED_MEMORY`): `FLAT` 11, `HNSW` 18, `SVS` 25, `TIERED_HNSW` 16, `TIERED_SVS` 18. The top-level comparators add `+1` for the `SHARED_MEMORY` field that the C API wrapper appends.
- `getFlatFields()`, `getHNSWFields()`, `getTieredHNSWFields()`, `getSVSFields()`, `getTieredSVSFields()` updated to append `SHARED_MEMORY` at the top level.

---

## Compatibility

- **Wire-compatible** — adds one new field at a well-defined position in `VECSIM_INFO`. Existing consumers parsing by field name are unaffected; consumers indexing by position must shift their expectations accordingly (covered above).
- **Source-compatible** — `VecSim_GetSharedMemory()` is purely additive.
- **Behavioral parity** — for callers that don't read the new field/API, total tracked memory is now strictly larger by exactly the bytes the shared SVS thread pool was already consuming but not reporting.

**Mark if applicable**

---

Supersedes #967 — ownership transferred to @dor-forer. Original branch authored by @meiravgri; identical commits cherry-pushed to `dor-forer-MOD-15578-track-svs-thpool-memory` so a non-self reviewer can be assigned.


src/VecSim/vec_sim.h
```diff
@@ -183,6 +183,18 @@ VecSimIndexBasicInfo VecSimIndex_BasicInfo(VecSimIndex *index);
...
+ * @brief Return process-wide VecSim statistics that are not tied to any single index.
+ * Currently exposes the memory used by the shared SVS thread pool singleton.
+ * Safe to call without holding any index lock; does not force initialization of the
+ * shared SVS pool (returns 0 in fields whose backing singleton has not been touched).
+ *
+ * @return Total bytes currently allocated by VecSim outside any single index
+ *         (e.g. the shared SVS thread pool singleton). 0 if no such allocations
+ *         have been made.
+ */
+size_t VecSim_GetGlobalMemory(void);
```
alonre24: Problematic name, implies that it also includes the per-index memory

src/VecSim/algorithms/svs/svs.h
```diff
@@ -429,8 +429,10 @@ class SVSIndex : public VecSimIndexAbstract<svs_details::vecsim_dt<DataType>, fl
 
     VecSimDebugInfoIterator *debugInfoIterator() const override {
         VecSimIndexDebugInfo info = this->debugInfo();
-        // For readability. Update this number when needed.
-        size_t numberOfInfoFields = 23;
+        // Capacity hint for the iterator. Must equal the number of addInfoField()
+        // calls below (1 for ALGORITHM + 9 from addCommonInfoToIterator + 16 SVS-specific).
+        // Update this number when fields are added or removed.
+        size_t numberOfInfoFields = 26;
```
alonre24: Only one pair was added no? How come it increased by 3?

tests/unit/test_svs.cpp
```diff
@@ -3362,6 +3362,60 @@ TEST(SVSTest, NumThreadsParamIgnored) {
...
+
+    VecSimDebugInfoIterator_Free(infoIterator);
+    VecSimIndex_Free(index);
+
+    // Reset the shared singleton pool to size 1 so the next test is not affected.
+    // Use VecSimSVSThreadPool::resize(1) directly (matching other thread-pool tests)
+    // to avoid the write-mode side effect that VecSim_UpdateThreadPoolSize(0) carries.
+    VecSimSVSThreadPool::resize(1);
+}
+
```
alonre24, line 3404: Is there a validation that the tracked memory actually increased/decreased as expected (rather than just that both APIs return the same value) ?

src/VecSim/algorithms/svs/svs_utils.h
```diff
@@ -615,8 +654,9 @@ class VecSimSVSThreadPoolImpl {
         }
     }
 
+    std::shared_ptr<VecSimAllocator> allocator_; // pool's own allocator for memory tracking
     mutable std::mutex pool_mutex_;
-    std::vector<std::shared_ptr<ThreadSlot>> slots_;
+    std::vector<SlotPtr, SlotVecAllocator> slots_;
```
alonre24: Wouldn't using vecsim_stl::vector simplify the whole thing?

src/VecSim/algorithms/svs/svs.h
dor-forer: It's actually +1 for the new field and +2 because the old 23 was already wrong.
Before the change the real count was 25 (1 ALGORITHM + 9 from addCommonInfoToIterator + 15 SVS-specific), but the comment just said "update this when needed" and it looks like nobody had. 
So adding the shared-memory field bumped the correct count to 26, not 24.
I added a little breakdown in the comment so it doesn't drift again. Worth noting it's only a reserve() hint, not a hard size, so the stale value wasn't actually causing any bug — just an extra realloc — which is probably why it slipped by.

src/VecSim/algorithms/svs/svs_utils.h
dor-forer: Done — slots_ is now `vecsim_stl::vector<SlotPtr>` (constructed with the pool's allocator), which removed the explicit `SlotVecAllocator` alias entirely.

tests/unit/test_svs.cpp
dor-forer: Added a new test `sharedMemoryTracksThreadPoolResize` that specifically validates directional tracking: it asserts that `VecSim_GetSharedMemory()` increases when the pool grows and decreases when it shrinks, ensuring the accounting is live and not just a static value that happens to match both APIs

tests/unit/test_svs.cpp
```diff
@@ -3362,6 +3362,85 @@ TEST(SVSTest, NumThreadsParamIgnored) {
...
+    size_t dim = 4;
+    SVSParams params = {.type = TypeParam::get_index_type(), .dim = dim, .metric = VecSimMetric_L2};
+    VecSimIndex *index = this->CreateNewIndex(params);
+    ASSERT_INDEX(index);
+
+    VecSimDebugInfoIterator *infoIterator = VecSimIndex_DebugInfoIterator(index);
+
+    bool seen_shared_memory = false;
+    bool seen_shared_svs = false;
+    uint64_t global_value = 0;
```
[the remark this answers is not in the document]
dor-forer: Removed

src/VecSim/utils/vec_utils.cpp
```diff
@@ -89,10 +89,11 @@ const char *VecSimCommonStrings::TIERED_SVS_UPDATE_THRESHOLD_STRING = "TIERED_SV
 const char *VecSimCommonStrings::TIERED_SVS_THREADS_RESERVE_TIMEOUT_STRING =
     "TIERED_SVS_THREADS_RESERVE_TIMEOUT";
 
-const char *VecSimCommonStrings::GLOBAL_MEMORY_STRING = "GLOBAL_MEMORY";
 const char *VecSimCommonStrings::SHARED_SVS_THREADPOOL_MEMORY_STRING =
     "SHARED_SVS_THREADPOOL_MEMORY";
 
+const char *VecSimCommonStrings::SHARED_MEMORY_STRING = "SHARED_MEMORY";
```
alonre24, line 92: Why do we need both `SHARED_MEMORY` and `SHARED_SVS_THREADPOOL_MEMORY_STRING` ? is there any difference between them?

dor-forer: Good catch — there's no real difference today, so I collapsed them into the single generic `SHARED_MEMORY` field (a923fa73).

The original two-field design came from the ticket spec (umbrella `GLOBAL_MEMORY`/`SHARED_MEMORY` + per-contributor `SHARED_SVS_THREADPOOL_MEMORY` breakdown). But the SVS thread pool is the only process-wide contributor today, so both always reported the exact same bytes — the breakdown carried no extra information. YAGNI: dropped `SHARED_SVS_THREADPOOL_MEMORY` and kept only `SHARED_MEMORY` + `VecSim_GetSharedMemory()`.

If a second global contributor is ever added, an itemized breakdown can be reintroduced then with real differentiation.

alonre24 (approved):

PR #984, opened 2026-07-02
[MOD-16688] x86 distance kernels: multi-accumulator unrolling + SIMD for smaller dimensions

GuyAv46:
**Jira:** [MOD-16688](https://redislabs.atlassian.net/browse/MOD-16688)

## What

Two related optimizations to the x86 distance kernels, porting the lessons learned from the ARM implementations (and from the newer SQ8↔FP16 kernels, which already do both):

1. **Multiple accumulators**: all touched kernels accumulated into a single SIMD register, so every `add`/`fmadd` waited for the previous one (3-5 cycle latency chain, while the CPU can issue 2/cycle). The kernels now accumulate into 2-4 independent registers and combine them once before the final reduction — same pattern as `IP_AVX512F_SQ8_FP16.h` / `IP_AVX2_SQ8_FP16.h`.
2. **SIMD for smaller dimensions**: the `dim < 16/32` early-outs to the scalar implementation were needed because the main loops were `do-while` (always executing one full chunk) — not because the residual loads were unsafe. The main loops are now `while` loops that can run zero times, and each dispatcher threshold is lowered to the *actual* minimum implied by its residual handling (the widest full-width load a kernel performs before there is provably enough data).

### Kernels changed (IP + L2; SQ8 Cosine/L2 wrappers inherit)

The accumulator count is tuned per SIMD width — a 1-vs-2-vs-4 study (see the [follow-up comment](https://github.com/RedisAI/VectorSimilarity/pull/984#issuecomment-4886480444)) showed the benefit shrinks as the vector widens, so wider tiers keep fewer accumulators:

| Family | Tiers | Accumulators | New min dim (was) |
|---|---|---|---|
| FP32 | SSE, AVX | 4 | 8 (16) |
| FP32 | AVX512F | 2 | 8 (16) |
| FP64 | SSE, AVX, AVX512F | 2 | 4 (8) |
| FP16 | F16C | 4 | 8 (32) |
| FP16 | AVX512F | 2 | 16 (32) |
| FP16 | AVX512FP16_VL | unchanged | 32 (32) — residual does a full 512-bit load |
| SQ8↔FP32 | SSE4, AVX2, AVX2_FMA | 4 | 8 (16) |
| SQ8↔FP32 | AVX512F_BW_VL_VNNI | 2 | 8 (16) |

Rationale for the accumulator split: 4 accumulators beat 2 by 10–45% on the narrow FP32/FP16/SQ8 tiers (SSE/AVX/F16C/AVX2), but gave nothing on FP64 (too few lanes) or on any AVX-512 kernel (the two 512-bit FMA units are already saturated by 2 accumulators at these dims). Using 2 there also halves the template instantiations and simplifies the tail.

BF16/INT8/UINT8/SQ8↔SQ8 are intentionally untouched: BF16 AVX2 was benchmarked and showed **no** gain (its unpacklo/unpackhi halves already provide enough ILP), and the integer VNNI kernels have a different bottleneck profile.

## Why (measured)

Standalone benchmark over the exact kernel templates (google-benchmark, pinned to a P-core, 9+ repetitions, random interleaving, medians; Intel Core Ultra 7 255U — AVX2/F16C class). The table below reports the multi-accumulator variant vs. the original single-accumulator kernel:

| Kernel | dim=512 | dim=733/754 | dim=1024 |
|---|---|---|---|
| FP16 L2 F16C | **1.69x** | **1.79x** | **1.87x** |
| SQ8-FP32 IP AVX2_FMA | **1.64x** | **1.51x** | **1.84x** |
| FP32 L2 SSE | 1.22x | 1.31x | 1.43x |
| FP64 L2 SSE | 1.39x | 1.42x | 1.69x |
| FP32 L2/IP AVX | 1.24x / 1.18x | 1.14x / 1.12x | 1.16x / 1.27x |
| FP64 L2 AVX | 1.12x | 1.12x | 1.24x |
| SQ8-FP32 IP AVX2 (no FMA) | 1.09x | 1.13x | 1.23x |
| BF16 L2 AVX2 | 1.02x | 0.96x | 1.01x | ← left unchanged

Newly-SIMD small dimensions vs. the scalar fallback they used before:

- FP32 dims 8-15: 1.4x-2.8x (scalar 3.0-6.2 ns → SIMD 2.1-2.2 ns)
- FP16 dims 8-31: **5.7x-9.4x** (scalar software-converts each element: 13-56 ns → SIMD 2.3-5.9 ns)
- Below the new thresholds the scalar loop is at least as fast, so the cutoffs stay there.

`tests/benchmark/bm_spaces_*` before/after on the same machine confirms (medians of 5 interleaved repetitions): FP32 SSE high-dim 1.38-1.48x, FP64 SSE high-dim 1.65-1.77x, FP16 F16C high-dim 1.76-1.82x, FP32/FP64 AVX 1.08-1.20x, residual dims improved across the board, `_512_bit_chunks` at dim 16/32 ~1.0x (no regression at the smallest sizes).

**AVX-512** (Xeon Platinum 8488C, via `bm-spaces` CI): FP32/FP64 AVX-512 kernels ~1.2–1.35x over the previous single-accumulator versions. A dedicated 2-vs-4 comparison run showed 4 accumulators give no benefit on AVX-512 (once an ~11% inter-run instance offset — visible on every identical-code control family — is normalized out), which is why the AVX-512 kernels use 2. Full detail in the [follow-up comment](https://github.com/RedisAI/VectorSimilarity/pull/984#issuecomment-4886480444).

## Residual-load safety (why the new thresholds are safe)

- **AVX-512 kernels** use genuinely fault-suppressing masked loads (`_mm512_maskz_loadu_*`), safe for any dimension.
- **AVX / F16C / AVX2 kernels** use full-width-load-then-blend for the sub-chunk residual (`my_mm256_maskz_loadu_ps`, `_mm_loadu_si128` + blend). These read a full 8-element block from the *start* of the vector, so they are safe iff the vector spans at least 8 elements — hence the `dim >= 8` (FP32/FP16/SQ8) and `dim >= 4` (FP64, 4-double block) floors. Every subsequent full step stays within `dim` by construction (verified for all dims ≤ 4096 with a coverage simulation of every loop/tail structure).
- **SSE kernels** load the residual element-wise (except FP32 L2's reversed-load trick for `residual%4==3`, which reads one aligned 16-byte block — in-bounds for every dim ≥ 8 that reaches it).
- SQ8 storage-side full 16-byte loads were already safe at any dim thanks to the metadata that follows the quantized values.

## Testing

- `test_spaces` (1267 tests) passes locally, including the extended coverage added here: FP32 opt tests now run dims 8-32, FP64 4-16, FP16 8-64, SQ8-FP32 8-32, and `smallDimChooser` asserts the new per-type scalar ranges.
- The AVX-512 tier kernels compile locally but **could not be executed or benchmarked locally** (no AVX-512 machine). They follow the exact structure of the measured AVX versions and of the existing `IP_AVX512F_SQ8_FP16.h`, were exercised on the `bm-spaces` CI Xeon, and passed **basic + sanitizer** unit-test CI on the (now-closed) 2-vs-4 comparison branch #986.
- Pre-existing local failures unrelated to this change (verified to fail identically with `main`'s spaces sources): `*SerializationCurrentVersionMulti`, `HNSWSerializationV3` (environmental), and a `test_common` segfault at `SearchDifferentScores` when run after the rest of the suite.

[MOD-16688]: https://redislabs.atlassian.net/browse/MOD-16688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Touches hot-path distance kernels used in vector search; correctness depends on residual load bounds at small dimensions, though behavior is covered by expanded unit tests and documented safety rules.
> 
> **Overview**
> **x86 inner-product and L2 SIMD kernels** (FP32/FP64/FP16 and SQ8↔FP32) now use **2–4 independent SIMD accumulators** per loop instead of one, then fold once before reduction, to break FMA/mul→add dependency chains. Main loops switch from **`do-while` to `while`**, so they can run **zero times** when only the compile-time residual remains.
> 
> **Residual handling** is reworked (`residual % lane_width`, extra constexpr steps for 8/16/24-byte tails), template residual ranges widen (e.g. **0..31** on many FP32/SQ8 paths), and **`CHOOSE_IMPLEMENTATION` chunk sizes** move to match **32-element** (or **16** FP64 AVX-512 / **4** FP64 SSE) unrolling.
> 
> **Dispatchers** in `IP_space.cpp` / `L2_space.cpp` lower SIMD cutoffs to what each kernel’s loads actually require (**FP32/SQ8 from dim ≥ 8**, **FP64 from ≥ 4**, **FP16** tiered at **8 / 16 / 32** by ISA). **`test_spaces`** extends parameterized dim coverage and updates `smallDimChooser` expectations.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3b04e145aaa2829936f9d4c139aba7e30e8cdc19. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/VecSim/spaces/functions/implementation_chooser.h
+++ b/src/VecSim/spaces/functions/implementation_chooser.h
@@ -32,8 +32,9 @@
 #define C32(func, N) C16(func, 2 * (N)) C16(func, 2 * (N) + 1)
 #define C64(func, N) C32(func, 2 * (N)) C32(func, 2 * (N) + 1)
 
-// Macros for 8, 16, 32 and 64 cases. Used to collapse the switch statement.
-// Expands into 0-7, 0-15, 0-31 or 0-63 cases respectively.
+// Macros for 4, 8, 16, 32 and 64 cases. Used to collapse the switch statement.
+// Expands into 0-3, 0-7, 0-15, 0-31 or 0-63 cases respectively.
+#define CASES4(func)  C4(func, 0)
 #define CASES8(func)  C8(func, 0)
 #define CASES16(func) C16(func, 0)
 #define CASES32(func) C32(func, 0)
--- a/src/VecSim/spaces/functions/implementation_chooser.h
+++ b/src/VecSim/spaces/functions/implementation_chooser.h
@@ -46,8 +46,8 @@
 // out:     The output variable that will be set to the chosen implementation.
 // dim:     The dimension.
 // func:    The templated function that we want to choose the implementation for.
-// chunk:   The chunk size. Can be 64, 32, 16 or 8. Should be the number of elements of the expected
-//          type fitting in the expected register size.
+// chunk:   The chunk size. Can be 64, 32, 16, 8 or 4. Should be the number of elements of the
+//          expected type fitting in the expected register size.
 #define CHOOSE_IMPLEMENTATION(out, dim, chunk, func)                                               \
     do {                                                                                           \
         decltype(out) __ret_dist_func;                                                             \
--- a/tests/unit/test_spaces.cpp
+++ b/tests/unit/test_spaces.cpp
@@ -835,7 +835,7 @@ TEST_P(FP32SpacesOptimizationTest, FP32InnerProductTest) {
 }
 
 INSTANTIATE_TEST_SUITE_P(FP32OptFuncs, FP32SpacesOptimizationTest,
-                         testing::Range(8UL, 16 * 2UL + 1));
+                         testing::Range(8UL, 32 * 2UL + 1));
 
 class FP64SpacesOptimizationTest : public testing::TestWithParam<size_t> {};
 
@@ -1030,7 +1030,7 @@ TEST_P(FP64SpacesOptimizationTest, FP64InnerProductTest) {
 }
 
 INSTANTIATE_TEST_SUITE_P(FP64OptFuncs, FP64SpacesOptimizationTest,
-                         testing::Range(4UL, 8 * 2UL + 1));
+                         testing::Range(4UL, 16 * 2UL + 1));
 
 class BF16SpacesOptimizationTest : public testing::TestWithParam<size_t> {};
 
@@ -2383,7 +2383,7 @@ TEST_P(SQ8_FP32_SpacesOptimizationTest, SQ8_FP32_InnerProductTest) {
 
 // Instantiate the test suite with dimensions to test
 INSTANTIATE_TEST_SUITE_P(SQ8_FP32_Test, SQ8_FP32_SpacesOptimizationTest,
-                         testing::Range(8UL, 16 * 2UL + 1));
+                         testing::Range(8UL, 32 * 2UL + 1));
 
 TEST_P(SQ8_FP32_SpacesOptimizationTest, SQ8_FP32_CosineTest) {
     auto optimization = getCpuOptimizationFeatures();
--- a/tests/unit/test_spaces.cpp
+++ b/tests/unit/test_spaces.cpp
@@ -582,44 +582,27 @@ TEST_F(SpacesTest, GetDistFuncSQ8FP16Asymmetric) {
 
 #ifdef CPU_FEATURES_ARCH_X86_64
 TEST_F(SpacesTest, smallDimChooser) {
-    // Verify that small dimensions gets the no optimization function.
+    // Verify that dimensions below each type's SIMD threshold get the no optimization function.
+    // FP64 is optimized from dim >= 4.
+    for (size_t dim = 1; dim < 4; dim++) {
+        ASSERT_EQ(L2_FP64_GetDistFunc(dim), FP64_L2Sqr);
+        ASSERT_EQ(IP_FP64_GetDistFunc(dim), FP64_InnerProduct);
+    }
+    // FP32 and FP16 are optimized from dim >= 8 (FP16 only on machines with F16C; for
+    // 8 <= dim < 32 the chosen FP16 function depends on the available features, so we only
+    // assert the range that is naive regardless of features).
     for (size_t dim = 1; dim < 8; dim++) {
         ASSERT_EQ(L2_FP32_GetDistFunc(dim), FP32_L2Sqr);
-        ASSERT_EQ(L2_FP64_GetDistFunc(dim), FP64_L2Sqr);
-        ASSERT_EQ(L2_BF16_GetDistFunc(dim), BF16_L2Sqr_LittleEndian);
-        ASSERT_EQ(L2_FP16_GetDistFunc(dim), FP16_L2Sqr);
-        ASSERT_EQ(L2_INT8_GetDistFunc(dim), INT8_L2Sqr);
-        ASSERT_EQ(L2_UINT8_GetDistFunc(dim), UINT8_L2Sqr);
         ASSERT_EQ(IP_FP32_GetDistFunc(dim), FP32_InnerProduct);
-        ASSERT_EQ(IP_FP64_GetDistFunc(dim), FP64_InnerProduct);
-        ASSERT_EQ(IP_BF16_GetDistFunc(dim), BF16_InnerProduct_LittleEndian);
-        ASSERT_EQ(IP_FP16_GetDistFunc(dim), FP16_InnerProduct);
-        ASSERT_EQ(IP_INT8_GetDistFunc(dim), INT8_InnerProduct);
-        ASSERT_EQ(IP_UINT8_GetDistFunc(dim), UINT8_InnerProduct);
-        ASSERT_EQ(Cosine_INT8_GetDistFunc(dim), INT8_Cosine);
-        ASSERT_EQ(Cosine_UINT8_GetDistFunc(dim), UINT8_Cosine);
-    }
-    for (size_t dim = 8; dim < 16; dim++) {
-        ASSERT_EQ(L2_FP32_GetDistFunc(dim), FP32_L2Sqr);
-        ASSERT_EQ(L2_BF16_GetDistFunc(dim), BF16_L2Sqr_LittleEndian);
         ASSERT_EQ(L2_FP16_GetDistFunc(dim), FP16_L2Sqr);
-        ASSERT_EQ(L2_INT8_GetDistFunc(dim), INT8_L2Sqr);
-        ASSERT_EQ(L2_UINT8_GetDistFunc(dim), UINT8_L2Sqr);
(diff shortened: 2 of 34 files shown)


GuyAv46:
**Update (commit 1afb939):** all residual-based conditions are now resolved at compile time. Each kernel's dispatch modulus (`CHOOSE_IMPLEMENTATION` chunk) was widened to match its main-loop stride, so the leftover full blocks that were previously handled with runtime tail checks after the main loop are instead part of the fully-unrolled `if constexpr` residual section:

| Kernels | Dispatch modulus | Main loop |
|---|---|---|
| AVX FP32, AVX2 SQ8-FP32 (+FMA) | 32 | 4×8, branch-free |
| AVX FP64 | 16 | 4×4 |
| AVX512F FP32 / FP16 / VNNI SQ8-FP32 | 64 | 4×16 |
| AVX512F FP64 | 32 | 4×8 |

SSE / F16C kernels already had stride == modulus, so they were unrolled through the template from the start. `test_spaces` (1267) passes; block-coverage simulation re-verified for all dims ≤ 4096.

Added the `bm-spaces` label to get CI benchmark numbers — that's where the AVX-512 results will show up.

GuyAv46:
## Follow-up: tuned accumulator counts per SIMD width

After the initial 4-accumulator pass, I benchmarked **1 vs 2 vs 4 accumulators** to check where the 4th/3rd accumulator actually pays off:

- **Local sweep** (standalone harness, dims 128/512/960, medians of 11 interleaved reps, pinned core) for the AVX2-class tiers.
- **CI 2-vs-4 comparison** on the Xeon 8488C (throwaway branch, now closed) for the AVX-512 tiers, since the dev machine has no AVX-512.

The multi-accumulator benefit shrinks as SIMD width grows:

| Tier | 4-acc vs 2-acc | Decision |
|---|---|---|
| SSE / AVX / F16C — FP32, FP16 | +10–45% | **keep 4** |
| AVX2 SQ8_FP32 (±FMA) | +8–21% | **keep 4** |
| FP64 (AVX + SSE) | ~0% (AVX slightly *worse*) | **→ 2** |
| AVX-512 (FP32/FP64/FP16/SQ8_FP32) | parity after normalizing an ~11% inter-run instance offset | **→ 2** |

Commit 8550b86a drops those kernels to 2 accumulators and narrows each dispatch modulus to 2× the SIMD chunk (AVX-512 FP32/FP16/SQ8 64→32, AVX-512 FP64 32→16, AVX FP64 16→8, SSE FP64 8→4; added `CASES4`). Fewer template instantiations, simpler tail, same or better speed. Residual handling stays fully `if constexpr`. 1267 `test_spaces` pass locally; the AVX-512 2-acc kernels passed basic + sanitizer CI on the (now-closed) comparison branch #986.

⚠️ The AVX-512 2-vs-4 signal was within single-pass CI noise, so the definitive check is the Grafana view aggregated over multiple runs. If a future run shows AVX-512 FP32/FP16 genuinely preferring 4, that's a cheap revert for those specific kernels.

src/VecSim/spaces/functions/implementation_chooser.h
dor-forer: Can now be also 4

src/VecSim/spaces/functions/implementation_chooser.h
```diff
@@ -32,8 +32,9 @@
 #define C32(func, N) C16(func, 2 * (N)) C16(func, 2 * (N) + 1)
 #define C64(func, N) C32(func, 2 * (N)) C32(func, 2 * (N) + 1)
 
-// Macros for 8, 16, 32 and 64 cases. Used to collapse the switch statement.
-// Expands into 0-7, 0-15, 0-31 or 0-63 cases respectively.
+// Macros for 4, 8, 16, 32 and 64 cases. Used to collapse the switch statement.
+// Expands into 0-3, 0-7, 0-15, 0-31 or 0-63 cases respectively.
+#define CASES4(func)  C4(func, 0)
```
dor-forer, line 37: Add `#undef CASES4` in `implementation_chooser_cleanup.h.`

tests/unit/test_spaces.cpp
```diff
@@ -852,7 +835,7 @@ TEST_P(FP32SpacesOptimizationTest, FP32InnerProductTest) {
 }
 
 INSTANTIATE_TEST_SUITE_P(FP32OptFuncs, FP32SpacesOptimizationTest,
-                         testing::Range(16UL, 16 * 2UL + 1));
+                         testing::Range(8UL, 16 * 2UL + 1));
```
dor-forer: The AVX/AVX512F FP32 kernels' residual  doubled to 32, but this range still stops at 16*2+1, so residual instantiations 1–7 (dims 33–39) are never executed. Same for FP64 at line 1033 (AVX512F modulus 16, residuals 1–3 / dims 17–19 uncovered) and SQ8↔FP32 at line 2386 (AVX2/FMA/VNNI modulus 32, residuals 1–7 uncovered). Should we extend to Range(8UL, 40UL) / Range(4UL, 20UL) / Range(8UL, 40UL)?

src/VecSim/spaces/functions/implementation_chooser.h
GuyAv46: Done in 3b04e145 — updated the `chunk` doc to list 4 as a valid size.

src/VecSim/spaces/functions/implementation_chooser.h
GuyAv46: Done in 3b04e145 — added `#undef CASES4` to implementation_chooser_cleanup.h alongside the other CASES macros.

tests/unit/test_spaces.cpp
GuyAv46: Good catch — fixed in 3b04e145. I widened each Range to `widest_modulus * 2 + 1` (the same idiom as the FP16 line): FP32 `Range(8, 32*2+1)` = 8..64, FP64 `Range(4, 16*2+1)` = 4..32, SQ8_FP32 `Range(8, 32*2+1)` = 8..64. That fully covers the previously-missed residual instantiations (and a second period as margin), rather than the minimal 8..40/4..20. test_spaces now runs 1460 tests, all green.

dor-forer (approved):

PR #985, opened 2026-07-03
Fix build on powerpc*

pkubaj:
Two simple patches to fix powerpc* builds.

**Mark if applicable**

---

> [!NOTE]
> **Low Risk**
> Small, build-target-specific wiring with no API or behavior change on x86_64/AArch64; PPC continues to use scalar distance paths guarded by existing arch macros.
> 
> **Overview**
> Fixes **powerpc\*** compilation and runtime CPU feature detection in VecSim by treating PPC like the existing x86_64 and AArch64 paths.
> 
> **`space_includes.h`** now includes `cpuinfo_ppc.h` when `CPU_FEATURES_ARCH_PPC` is defined. **`getCpuOptimizationFeatures`** in `spaces.h` selects `cpu_features::PPCFeatures` and `GetPPCInfo` on PPC instead of falling through to the x86 types and `GetX86Info`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f03391281c9ba09a139fab946bc3361bcc69e126. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/VecSim/spaces/space_includes.h
+++ b/src/VecSim/spaces/space_includes.h
@@ -17,6 +17,9 @@
 #ifdef CPU_FEATURES_ARCH_AARCH64
 #include "cpuinfo_aarch64.h"
 #endif // CPU_FEATURES_ARCH_AARCH64
+#ifdef CPU_FEATURES_ARCH_PPC
+#include "cpuinfo_ppc.h"
+#endif // CPU_FEATURES_ARCH_PPC
 
 #if defined(__AVX512F__) || defined(__AVX__) || defined(__SSE__)
 #if defined(__GNUC__)
--- a/src/VecSim/spaces/spaces.h
+++ b/src/VecSim/spaces/spaces.h
@@ -57,6 +57,9 @@ static inline auto getCpuOptimizationFeatures(const void *arch_opt = nullptr) {
 #if defined(CPU_FEATURES_ARCH_AARCH64)
     using FeaturesType = cpu_features::Aarch64Features;
     constexpr auto getFeatures = cpu_features::GetAarch64Info;
+#elif defined(CPU_FEATURES_ARCH_PPC)
+    using FeaturesType = cpu_features::PPCFeatures;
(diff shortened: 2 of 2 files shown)


alonre24:
@pkubaj Thank you for the contribution. First, please sign the CLA. Second, did you confirm that this addition would enable SIMD optimizations on PPC?

pkubaj:
No, it doesn't — and it isn't meant to. This is purely a build fix; PPC keeps using the scalar distance functions.
The SIMD selection is entirely gated behind CPU_FEATURES_ARCH_X86_64 and CPU_FEATURES_ARCH_AARCH64. For example in L2_space.cpp every optimized path sits inside those blocks:
```
#ifdef CPU_FEATURES_ARCH_AARCH64
    if (features.sve2) return Choose_..._SVE2(dim);
    if (features.asimd) return Choose_..._NEON(dim);
#endif
#ifdef CPU_FEATURES_ARCH_X86_64
    if (features.avx512f && ...) return Choose_..._AVX512...(dim)
    if (features.avx2) return Choose_..._AVX2(dim);
#endif
    // falls through to the scalar implementation
```
  
There are no CPU_FEATURES_ARCH_PPC dispatch blocks and no VSX/AltiVec kernels, so on PPC none of the features.* fields are ever read — the functions fall straight through to the scalar default.
      
What the change actually does is narrower: without it, getCpuOptimizationFeatures() on PPC falls into the #else and instantiates cpu_features::X86Features / GetX86Info, which don't exist on a PPC build (cpu_features compiles its PPC module there, not the x86 one) — that's the compile failure. Routing PPC to PPCFeatures/GetPPCInfo makes it build; the returned struct is simply never consulted, exactly like the scalar fallback intends.
 
Actually enabling SIMD on PPC would be a separate, much larger change — adding VSX implementations plus the OPT_* macros and Choose_* dispatch — which is out of scope for a build fix.

PR #986, opened 2026-07-05
[BENCH-ONLY] AVX-512: 2 accumulators vs 4 (do not merge)

GuyAv46:
**Do not merge.** Benchmark comparison branch off #984 (`perf-x86-multi-accumulators-small-dims`).

## Purpose
On #984 the AVX-512 float/SQ8 distance kernels use **4 accumulators** (dispatch modulus 64 for FP32/FP16/SQ8_FP32, 32 for FP64). Locally we confirmed 4-acc beats 2-acc for the AVX/SSE/F16C/SQ8 tiers, but the Core Ultra 7 255U dev machine has **no AVX-512**, so the 2-vs-4 question for the AVX-512 kernels can only be answered on CI hardware (Xeon Platinum 8488C, Sapphire Rapids).

This branch is identical to #984 except the AVX-512 kernels are reverted to **2 accumulators**, narrowing each dispatch modulus to 2× the SIMD chunk:
- FP32 / FP16 / SQ8_FP32 VNNI: 64 → 32
- FP64: 32 → 16

Compare the `bm-spaces` results here (branch `perf-avx512-2acc-compare`) against #984's run (branch `perf-x86-multi-accumulators-small-dims`) for the `*_AVX512F_*` and `*_AVX512F_BW_VL_VNNI_*` families to decide whether 4 accumulators is worth it on AVX-512.

### DIFF
--- a/src/VecSim/spaces/IP/IP_AVX2_FMA_SQ8_FP32.h
+++ b/src/VecSim/spaces/IP/IP_AVX2_FMA_SQ8_FP32.h
@@ -52,10 +52,15 @@ float SQ8_FP32_InnerProductImp_FMA(const void *pVect1v, const void *pVect2v, siz
     const float *pVect2 = static_cast<const float *>(pVect2v);     // FP32 query
     const uint8_t *pEnd1 = pVect1 + dimension;
 
-    // Initialize sum accumulator for Σ(q_i * y_i)
-    __m256 sum256 = _mm256_setzero_ps();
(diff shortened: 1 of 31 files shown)


GuyAv46:
Superseded: folded the AVX-512 → 2-accumulator change into #984 (commit 8550b86a), together with the same reduction for FP64 (AVX+SSE). The 2-vs-4 CI comparison this branch produced showed 4 accumulators give no benefit on AVX-512 (after normalizing out an ~11% inter-run EC2 speed offset that every identical-code control family exhibited). Closing this throwaway branch.

PR #996, opened 2026-07-13
[MOD-16610] SVS shared pool: rent() waits for occupied slots instead of dropping work

dor-forer:

Follow-up to RediSearch/RediSearch#10419 (cap the shared SVS pool at the available CPU count), addressing the concurrency gap flagged in [its review](https://github.com/RediSearch/RediSearch/pull/10419#discussion_r3566495192).

**The problem.** Cross-index exclusion of SVS scheduled jobs was an *accident* of the pool sizes being equal: with the SVS pool sized to `WORKERS`, one scheduled job's reserve jobs saturated the entire RediSearch worker pool, so a second index's update/GC could not start, and `rent()` could provably never come up short (total rented = total reserved − #jobs ≤ WORKERS − 1 = slots). Once the SVS pool is capped at the CPU count while `WORKERS` stays larger, a scheduled job reserves only pool-size workers, spare workers pick up a second index's jobs immediately, and both jobs rent from the same slots. The loser's `rent()` came up short: warning + `assert(false)` in debug builds (shard abort), and in release builds `parallel_for` silently dropped the un-rented partitions — an incomplete update/GC that reports success. Reachable even in OSS, e.g. `WORKERS 16` on an 8-CPU host with two SVS indexes.

**The fix.** Make slot acquisition in `rent()` all-or-nothing and blocking:

- A renter that finds fewer free slots than requested sleeps on a new condition variable (`slots_released_`) until slots are released or the pool is resized. It never holds a partial set of slots while waiting, so concurrent renters cannot deadlock each other.
- Slot holders never block on waiters — slots are always released via RAII (`RentedThreads` destructor) when their `parallel_for` ends — so the wait always terminates. The waiter's job keeps holding its reserved RediSearch workers meanwhile, same as any long-running job.
- `RentedThreads::release()` wakes waiters via an empty lock/unlock of `pool_mutex_` + `notify_all`, pairing with the waiter's condition wait so a wakeup cannot be missed.
- `resize_locked()` wakes waiters when the slot count changes (new free slots after a grow; an impossible request after a shrink).
- Requests larger than the whole pool (only possible if the pool shrinks under a non-scheduled caller) keep the legacy best-effort + warn + assert path.

Concurrent *disjoint* rentals (two indexes whose parallelism sums to ≤ pool size) are unchanged and still run in parallel (`ConcurrentRentalFromTwoIndexes` still passes). The Redis main thread is never affected: it only enqueues jobs; renting happens on worker threads (in write-in-place mode the pool has 0 worker slots and `parallel_for(f, 1)` never rents).

**Testing.** Rewrote `SVSThreadPoolTest.AllThreadsOccupied`, which previously asserted the degradation as a "defensive fallback" (`resultB == 1` / `ASSERT_DEATH`), to assert the new semantics: wrapper B arriving at a fully-occupied pool executes nothing while A holds the slots, then runs **all** of its partitions after A finishes, observing A's completed work. Written TDD-style: the test aborts on the previous code at the `rent()` assert; passes with the fix. Full `test_svs` suite: 465 passed.

1. MOD-16610 (together with RediSearch/RediSearch#10419)

1. `VecSimSVSThreadPoolImpl::rent()` — waits (all-or-nothing) instead of returning short
2. `VecSimSVSThreadPoolImpl::RentedThreads` — release now wakes waiting renters
3. `VecSimSVSThreadPoolImpl::resize_locked()` — wakes waiting renters on slot-count changes
4. `tests/unit/test_svs_threadpool.cpp::AllThreadsOccupied` — asserts wait-then-complete semantics

**Mark if applicable**

### DIFF
--- a/src/VecSim/algorithms/svs/svs_utils.h
+++ b/src/VecSim/algorithms/svs/svs_utils.h
@@ -22,9 +22,11 @@
 #endif
 
 #include <atomic>
+#include <condition_variable>
 #include <cstdint>
 #include <cstdlib>
 #include <memory>
+#include <mutex>
 #include <optional>
 #include <string>
 #include <utility>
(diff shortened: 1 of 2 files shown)


dor-forer:
Closing: the plan for MOD-16610 changed. Instead of capping only the SVS pool at the CPU count (which decouples the pool sizes and makes this concurrent-rent scenario reachable), RediSearch/RediSearch#10419 now hard-caps MAX_WORKER_THREADS itself at 24. The worker pool and the SVS pool stay the same size, so the reserve-job saturation invariant that makes rent() total-demand-safe holds structurally and no VecSim change is needed. Keeping the branch around in case the pools are ever decoupled again — this fix (and its test) would become necessary at that point.

PR #1000, opened 2026-07-22
[MOD-14952] Support normalization in QuantPreprocessor

dor-forer:
**Describe the changes in the pull request**

PR [#999](https://github.com/RedisAI/VectorSimilarity/pull/999) has merged. This PR targets `main` and builds on its stored-to-query distance dispatch.

Cherry-picks ARM's [`027d5510b2bf7add6a45be9b0c07cda38c15b5ca`](https://github.com/ARM-software/VectorSimilarity-for-Arm/commit/027d5510b2bf7add6a45be9b0c07cda38c15b5ca) from [ARM-software/VectorSimilarity-for-Arm#2](https://github.com/ARM-software/VectorSimilarity-for-Arm/pull/2).

This PR adds an optional normalization-aware mode to `QuantPreprocessor`. It stores the mean-vector inner product as additional SQ8 metadata for normalized storage and queries, extends SQ8 metadata sizing, and shares a float-based quantization implementation to avoid duplicated logic.

Existing `QuantPreprocessor` configurations retain their current behavior. FP32 and FP16 tests cover normalization for L2 and inner-product metrics.

It also follows up on review feedback from #999 with an HNSW integration test that installs distinct stored-to-stored and stored-to-query kernels. Top-K, range, and batch-iterator searches must all return the query-kernel score, proving HNSW selects the cached `StoredToQuery` dispatch.

**Which issues this PR fixes**

1. MOD-14952

**Main objects this PR modified**

1. `QuantPreprocessor`
2. `vecsim_types::sq8`
3. Quantization component unit tests
4. HNSW stored-to-query dispatch integration coverage

**Mark if applicable**

- [x] This PR introduces API changes
- [ ] This PR introduces serialization changes

**Validation**

- `make format`
- `make check-format`
- Release build of `test_components` and `test_hnsw`
- `test_components`: 32/32 passed
- `test_hnsw` excluding the two external serialization fixtures: 169/169 passed
- GitHub Actions, sanitizer, CodeQL, and Codecov checks passed
- `git diff --check main...HEAD`

---

> [!NOTE]
> **Medium Risk**
> Introduces a new preprocessor constructor and larger SQ8 metadata when normalization is on; default `WithNorm=false` limits blast radius, but wrong mean or layout mismatches would break distance reconstruction.
> 
> **Overview**
> Adds an optional **normalization-aware** mode to `QuantPreprocessor` via a `WithNorm` template flag (default **off**, so existing indexes keep the same blob layout and behavior).
> 
[...]

PR #999, opened 2026-07-22
[MOD-14953] Add calcDistanceForQuery to IndexCalculatorInterface

dor-forer:
**Describe the changes in the pull request**

Cherry-picks ARM's [`ed1df68a4f1585747068a0560720fd76fb92a6b9`](https://github.com/ARM-software/VectorSimilarity-for-Arm/commit/ed1df68a4f1585747068a0560720fd76fb92a6b9) from [ARM-software/VectorSimilarity-for-Arm#1](https://github.com/ARM-software/VectorSimilarity-for-Arm/pull/1).

This PR adds `calcDistanceForQuery` so stored candidates and query blobs can use asymmetric representations and operand ordering.

It also introduces a construction-time `DistanceDispatch` selected through `getDistanceDispatch(DistanceMode)`. HNSW caches one dispatch for stored-to-stored distances and another for stored-to-query distances, avoiding calculator vtable lookups in the hot path.

The dispatch has separate optimized representations:

- Stateless calculators cache the raw distance kernel directly, avoiding a context adapter or second indirect function call.
- Stateful calculators cache a contextual callback and opaque context pointer, allowing later calculators to access index-level state without restoring virtual dispatch in HNSW.

HNSW search and batch-iterator paths use the stored-to-query dispatch. Index construction and maintenance continue using the stored-to-stored dispatch. Component tests cover symmetric defaults, asymmetric query functions, stateless dispatch, and contextual dispatch.

**Which issues this PR fixes**

1. MOD-14953

**Main objects this PR modified**

1. `IndexCalculatorInterface`, `DistanceCalculatorInterface`, and `DistanceDispatch`
2. `VecSimIndexAbstract`
3. `HNSWIndex` and `HNSW_BatchIterator`
4. Component and tiered-HNSW unit tests

**Mark if applicable**

- [x] This PR introduces API changes
- [ ] This PR introduces serialization changes

**Validation**

- `make format`
- `make check-format`
- Clean release build of `test_components` and `test_hnsw`
- `test_components`: 28/28 passed
- `test_hnsw` excluding the two external serialization fixtures: 168/168 passed
- ASAN focused regression test: `HNSWTest/0.hnsw_blob_sanity_test` passed

PR #1002, opened 2026-07-23
[MOD-14955] Add DistanceCalculatorWithNorm

dor-forer:

This PR is stacked on [#1000](https://github.com/RedisAI/VectorSimilarity/pull/1000). Review only the incremental diff against `dor-forer-MOD-14952-support-quant-normalization`.

Cherry-picks the two commits from [ARM-software/VectorSimilarity-for-Arm#3](https://github.com/ARM-software/VectorSimilarity-for-Arm/pull/3): [`493ee7a8`](https://github.com/ARM-software/VectorSimilarity-for-Arm/commit/493ee7a8596c19c10ad2b467364fbfefd61732f6) and [`c27401d6`](https://github.com/ARM-software/VectorSimilarity-for-Arm/commit/c27401d6d5a29e2f98d8c80fbe77f8ce9e57156e).

This PR adds `DistanceCalculatorWithNorm`, which invokes the selected SQ8 kernel and applies the mean-normalization correction terms needed to recover the distance between the original vectors. The follow-up commit makes metadata reads safe for unaligned storage.

The calculator integrates with the construction-time dispatch introduced in #999:

- HNSW caches a non-virtual callback and a minimal immutable context containing only the stored/query kernels and `mean_sum_squares`.
- The candidate loop does not call `IndexCalculatorInterface::calcDistance` or `calcDistanceForQuery`, so it does not restore the per-distance vtable lookup removed by #937.
- Existing calculators retain the stateless raw-kernel path. Stored-to-stored normalized L2 also uses that path because its mean terms cancel.

Tests cover FP32/FP16, L2/IP, symmetric/asymmetric, zero-mean, randomized-vector, and stateless/stateful dispatch behavior.

1. MOD-14955

1. `DistanceCalculatorWithNorm`
2. Normalized-distance component unit tests

**Mark if applicable**

**Validation**

- `make format`
- `make check-format`
- Release build of `test_components` and `test_hnsw`
- `test_components`: 40/40 passed
- `test_hnsw` excluding the two external serialization fixtures: 169/169 passed
- `git diff --check dor-forer-MOD-14952-support-quant-normalization...HEAD`
- Focused stateless-dispatch microbenchmark: 36 alternating-order rounds, 20M calls per path per round; cached raw kernel averaged 2.75 ns/call and `DistanceDispatch` averaged 2.69 ns/call (no measurable regression within run-to-run noise)

---

> [!NOTE]
> **Medium Risk**
> Changes core VecSim distance math and blob layouts for normalized SQ8 indices; mistakes would skew ANN search, but behavior is heavily covered by new unit tests including numerical edge cases.
> 
> **Overview**
> Adds **`DistanceCalculatorWithNorm`**, which runs the existing SQ8 distance kernels on mean-centered quantized storage and applies **IP-only** correction terms (`x_mean_ip`, `y_mean_ip`, `mean_sum_squares`) so reported distances match the **original** (unshifted) vectors. **Normalized L2** skips runtime correction: queries are **centered at preprocess time** (`y - mean`), so stored/query kernels already compute the true L2 distance; **L2** paths expose **stateless** `DistanceDispatch` (raw kernels), while **IP** uses **stateful** dispatch for the correction callbacks.
> 
> **`QuantPreprocessor` (WithNorm)** is aligned with that split: L2 query blobs store centered values and metadata sums over centered data; IP keeps raw query values and records **`y_mean_ip`** from the original query. **`sq8` metadata layout** only adds the extra **`mean_ip`** slot when **WithNorm + IP** (not for normalized L2).
> 
> Unit tests cover FP32/FP16, L2/IP, symmetric/asymmetric paths, zero mean, large-offset L2 stability, random vectors, odd dimensions (unaligned metadata reads via `load_unaligned`), and dispatch stateless vs stateful behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 385741f98f53fc00f61110f2c8c759bf03c0eedf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/VecSim/spaces/computer/calculator.h
+++ b/src/VecSim/spaces/computer/calculator.h
@@ -150,8 +150,7 @@ public:
  * DistType is float
  */
 template <typename DataType, typename DistType, VecSimMetric Metric>
-class DistanceCalculatorWithNorm
-    : public DistanceCalculatorInterface<DistType, spaces::dist_func_t<DistType>> {
+class DistanceCalculatorWithNorm : public IndexCalculatorInterface<DistType> {
     static_assert(Metric == VecSimMetric_L2 || Metric == VecSimMetric_IP,
                   "DistanceCalculatorWithNorm only supports L2 and IP metrics");
 
@@ -212,13 +211,8 @@ public:
     DistanceCalculatorWithNorm(std::shared_ptr<VecSimAllocator> allocator,
                                spaces::dist_func_t<DistType> asym_func,
                                spaces::dist_func_t<DistType> sym_func, float mean_sum_squares)
-        : DistanceCalculatorInterface<DistType, spaces::dist_func_t<DistType>>(allocator, sym_func,
-                                                                               asym_func),
-          context_{
-              .stored_func = sym_func,
-              .query_func = asym_func,
-              .mean_sum_squares = mean_sum_squares,
-          } {}
+        : IndexCalculatorInterface<DistType>(allocator),
+          context_{sym_func, asym_func, mean_sum_squares} {}
 
     // Symmetric: both v1 and v2 are stored SQ8-of-x' blobs.
     DistType calcDistance(const void *v1, const void *v2, size_t dim) const override {
@@ -236,7 +230,7 @@ public:
         if (mode == DistanceMode::StoredToStored) {
             if constexpr (Metric == VecSimMetric_L2) {
                 // The mean terms cancel for stored-to-stored L2, so retain the stateless fast path.
-                return DistanceDispatch<DistType>::stateless(this->dist_func);
+                return DistanceDispatch<DistType>::stateless(context_.stored_func);
             }
             return DistanceDispatch<DistType>::stateful(&context_, calcStoredWithContext);
         }
--- a/src/VecSim/spaces/computer/calculator.h
+++ b/src/VecSim/spaces/computer/calculator.h
@@ -12,6 +12,7 @@
 #include "VecSim/memory/vecsim_base.h"
 #include "VecSim/spaces/spaces.h"
 #include "VecSim/types/sq8.h"
+#include "VecSim/utils/alignment.h"
 
 enum class DistanceMode {
     StoredToStored,
@@ -173,12 +174,12 @@ private:
             // base = 1 - IP(x', y'). We want 1 - IP(x, y).
             // IP(x, y) = IP(x', y') + x_mean_ip + y_mean_ip - mean_sum_squares
             // => 1 - IP(x, y) = base - x_mean_ip - y_mean_ip + mean_sum_squares
-            const float *meta1 =
-                reinterpret_cast<const float *>(static_cast<const uint8_t *>(v1) + dim);
-            const float *meta2 =
-                reinterpret_cast<const float *>(static_cast<const uint8_t *>(v2) + dim);
-            float x_mean_ip = meta1[sq8::template mean_ip_index<Metric>()];
-            float y_mean_ip = meta2[sq8::template mean_ip_index<Metric>()];
+            float x_mean_ip =
+                load_unaligned<float>(static_cast<const uint8_t *>(v1) + dim +
+                                      sq8::template mean_ip_index<Metric>() * sizeof(float));
+            float y_mean_ip =
+                load_unaligned<float>(static_cast<const uint8_t *>(v2) + dim +
+                                      sq8::template mean_ip_index<Metric>() * sizeof(float));
             return base - x_mean_ip - y_mean_ip + context->mean_sum_squares;
         } else { // L2: ||x - y||² = ||x' - y'||² (mean terms cancel exactly)
             return base;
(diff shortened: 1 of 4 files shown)


src/VecSim/vec_sim_index.h
```diff
@@ -163,28 +163,28 @@ struct VecSimIndexAbstract : public VecSimIndexInterface {
      *
      * Uses the cached dist func to avoid the indexCalculator vtable on the hot path.
      *
-     * @note Precondition: @c cachedDistFunc must be non-null. Subclasses that construct
-     *       this index with a null @c indexCalculator (e.g. SVS, which uses its own
-     *       internal distance kernels) must not call this method.
-     *
      * @return the distance between the vectors.
      */
     DistType calcDistance(const void *vector_data1, const void *vector_data2) const {
```
ofiryanai, line 175: This method calls `indexCalculator->calcDistance`, restoring the per-distance `IndexCalculatorInterface` vtable fallback removed by PR #937. `DistanceCalculatorWithNorm` always returns nullptr for the query kernel, so every visited candidate now takes a virtual call followed by the raw kernel call; existing calculators also pay the new branch. Cache a non-virtual state-aware callback or specialize the normalized dispatch outside the candidate loop, and benchmark against the #937 path before merging

ofiryanai (changes requested):

src/VecSim/spaces/computer/calculator.h
```diff
@@ -120,3 +122,121 @@ class DistanceCalculatorCommon
...
+            // IP(x, y) = IP(x', y) + y_mean_ip
+            // => 1 - IP(x, y) = base - y_mean_ip
+            return base - y_mean_ip;
+        } else { // L2
+            // base = ||x' - y||². We want ||x - y||².
+            // ||x - y||² = ||x' - y||² + 2*(x_mean_ip - y_mean_ip) - mean_sum_squares
+            float x_mean_ip =
+                load_unaligned<float>(static_cast<const uint8_t *>(candidate) + dim +
+                                      sq8::template mean_ip_index<Metric>() * sizeof(float));
+            return base + 2.0f * (x_mean_ip - y_mean_ip) - context->mean_sum_squares;
```
ofiryanai: This L2 correction subtracts several `O(||mean||²)` FP32 values after the base kernel has already rounded its result, causing catastrophic cancellation for nearby vectors with a nontrivial mean. For dim=128, mean[i]=1000, x[i]=1001, and y[i]=1001.1, the current FP32 evaluation produces approximately 40 instead of 1.28 and may even become negative. Avoid forming `||x′−y||²` and correcting it afterward; preprocess/use the centered query y−mean for normalized L2, or fuse the calculation so all terms remain centered. Add a large-offset/small-distance regression test.

src/VecSim/spaces/computer/calculator.h
```diff
@@ -120,3 +122,121 @@ class DistanceCalculatorCommon
...
+        spaces::dist_func_t<DistType> query_func;
+        float mean_sum_squares;
+    };
+
+    const WithNormDistanceContext context_;
+
+    static DistType calcStoredWithContext(const void *opaque_context, const void *v1,
+                                          const void *v2, size_t dim) {
+        const auto *context = static_cast<const WithNormDistanceContext *>(opaque_context);
+        DistType base = context->stored_func(v1, v2, dim);
```
ofiryanai, line 171: `load_unaligned` protects the new correction metadata, but `stored_func`/`query_func` still read existing SQ8 float metadata through `reinterpret_cast<const float *>(blob + dim)`. That address is unaligned whenever `dim % 4 != 0`, so normalized distances still have alignment UB despite this PR’s unaligned-load fix. Convert the underlying SQ8 kernel metadata reads to `load_unaligned` (or guarantee an aligned layout) and add an odd-dimension sanitizer test; all current calculator tests use dim=8 or 16.

ofiryanai (changes requested):

src/VecSim/spaces/computer/calculator.h
dor-forer: Good catch — this is not fixed in the current head. The contextual dispatch change addressed the per-distance virtual-call regression, but the normalized L2 query path still applies this cancellation-prone correction. I will change normalized L2 query preprocessing to use the centered query (`y - mean`) and add the large-offset/small-distance regression test before resolving this thread.

src/VecSim/spaces/computer/calculator.h
dor-forer: This underlying SQ8 metadata-alignment issue is covered by #1001 (MOD-15303), which converts the scalar and SIMD SQ8-FP32/SQ8-SQ8 metadata reads to `load_unaligned<float>` and adds odd-dimension sanitizer regression coverage. We will merge that fix separately and rebase this PR onto it rather than duplicating the broad kernel changes here. I will also add a small odd-dimension `DistanceCalculatorWithNorm` integration test in this PR after rebasing.

src/VecSim/spaces/computer/preprocessors.h
```diff
@@ -370,10 +384,10 @@ class QuantPreprocessor : public PreprocessorInterface {
             }
 
             if constexpr (WithNorm) {
-                m0 += mean[i + 0] * y0;
-                m1 += mean[i + 1] * y1;
-                m2 += mean[i + 2] * y2;
-                m3 += mean[i + 3] * y3;
+                m0 += mean[i + 0] * to_fp32<DataType>(original_input[i + 0]);
```
ofiryanai: P3: Centered L2 no longer consumes x_mean_ip, y_mean_ip, or mean_sum_squares, but WithNorm still computes both dot products and stores an extra float in every L2 storage/query blob. Make correction metadata conditional on Metric == VecSimMetric_IP to avoid dead O(dim) preprocessing work and storage.

tests/unit/test_components.cpp
```diff
@@ -1805,3 +1814,529 @@ INSTANTIATE_TEST_SUITE_P(QuantPreprocessorFP16WithNormTests,
...
+        EXPECT_NEAR(asym_dist, bf, 0.05f) << "Asymmetric L2 vs brute-force";
+
+        allocator->free_allocation(x_blob);
+        allocator->free_allocation(y_blob);
+        allocator->free_allocation(y_query);
+        delete calc;
+    }
+}
+
+TEST(DistanceCalculatorWithNormTest, RandomVectors) {
```
ofiryanai, line 2279: P4: The underlying odd-dimension SQ8 kernels are covered by #1001, but the promised odd-dimension DistanceCalculatorWithNorm integration case is still missing. Add one to exercise the complete normalized metadata layout and correction path.

ofiryanai (dismissed):
Blocking correctness and alignment concerns are resolved. The remaining P3/P4 comments are non-blocking.

src/VecSim/spaces/computer/preprocessors.h
dor-forer: Fixed in e0a0a203: gated the extra mean_ip metadata slot/computation to `WithNorm && Metric == VecSimMetric_IP` (in `sq8.h` and `preprocessors.h`) instead of `WithNorm` alone. L2 no longer computes or stores `x_mean_ip`/`y_mean_ip` since the centered-L2 correction never reads them. Also had to fix a pre-existing test in `QuantPreprocessorWithNormMetricTest` (1b39f3f5) that was asserting on `mean_ip_index<Metric>()` unconditionally for L2 too — caught by Bugbot after the metadata layout changed size.

tests/unit/test_components.cpp
dor-forer: Added `DistanceCalculatorWithNormTest.OddDimension_MatchesBruteForce` (dim=13) covering IP and L2 symmetric + asymmetric paths through the unaligned storage-blob metadata offset.

ofiryanai (approved):

PR #1010, opened 2026-08-09
Allow storeNewElement to reuse an existing element id

nonirosenfeldredis:

`HNSWIndex::storeNewElement` always allocated a fresh id by advancing `curElementCount`, and
appended the vector and the graph data to their containers. There was no way to store an element
into an id that the caller already holds - e.g. an already-freed slot one wants to reuse instead
of growing the index.

This PR adds an optional `elementId` parameter:

* **`INVALID_ID` (the default)** - unchanged behaviour: a fresh id is allocated via
  `curElementCount++`, the index grows by a block first if it is full, and the vector and the
  graph data are **appended** (`RawDataContainer::addElement` / `DataBlock::addElement`).
* **A valid id** - the id is used as-is, `curElementCount` is left untouched, no capacity check or
  growth happens, and the vector and the graph data are written **in place** over that slot
  (`RawDataContainer::updateElement` / `DataBlock::updateElement`). The in-place path is required
  because both containers only support appending at `curElementCount` -
  `DataBlocksContainer::addElement` asserts `id == element_count`.

Everything downstream of the id already works by id and needed no change: `idToMetaData[id]` is
reset to `ElementMetaData(label)` (which re-raises `IN_PROCESS`), `setVectorId` re-publishes the
label mapping, and the entry-point / `maxLevel` promotion is untouched.

**Caller contract for the reuse path** (documented at the declaration): the slot must be a legal,
already-freed one - `elementId < curElementCount` (asserted) and the previous occupant's
`ElementGraphData` must have already been destroyed, otherwise its link lists are leaked by the
overwrite. Note also that the slot's `elementLocks[elementId]` is not taken here, so the caller
must guarantee the element is unreachable by concurrent scans - `storeNewElement`'s usual contract
(`indexDataGuard` held exclusive) is not sufficient on its own for the reuse path.

All three existing call sites are unchanged and keep allocating fresh ids:
`HNSWIndex::storeVector`, and the two branches of `TieredHNSWIndex::insertVectorToHNSW`.

1. MOD-9645

1. `HNSWIndex::storeNewElement` (`src/VecSim/algorithms/hnsw/hnsw.h`) - optional `elementId`
   parameter plus the in-place store path

**Testing**

Full unit suite green locally (macOS arm64, RelWithDebInfo): `100% tests passed out of 2556`.
No new test yet - the reuse path currently has no caller, so coverage should land together with
the first one.

**Mark if applicable**

> Internal (non-public) API only: `storeNewElement` gains a defaulted parameter, so every existing
> caller and the public `VecSim` API are unaffected.

### DIFF
--- a/src/VecSim/algorithms/hnsw/hnsw.h
+++ b/src/VecSim/algorithms/hnsw/hnsw.h
@@ -277,7 +277,17 @@ public:
     bool isMarkedDeleted(idType internalId) const;
     bool isInProcess(idType internalId) const;
     void unmarkInProcess(idType internalId);
-    HNSWAddVectorState storeNewElement(labelType label, const void *vector_data);
+    // `elementId` is optional:
+    // - INVALID_ID (the default): a new id is allocated by advancing `curElementCount`, and the
+    //   vector and the graph data are *appended* to their containers (growing the index by a block
+    //   first if it is full).
+    // - a valid id: the id is used as-is, `curElementCount` is left untouched and the index does
+    //   not grow. The vector and the graph data are written *in place* over that slot.
+    //   The caller must guarantee that the slot is a legal, already-freed one - that is,
+    //   `elementId < curElementCount` and the previous occupant's `ElementGraphData` has been
+    //   destroyed (see `ElementGraphData::destroy`), otherwise its link lists are leaked.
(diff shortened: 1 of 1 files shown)


nonirosenfeldredis:
Superseded by #1057 (same work, rebased onto current main).

PR #1007, opened 2026-08-05
[MOD-14956] Add SQ8 quantization support for HNSW index

dor-forer:
**Describe the changes in the pull request**

Cherry-pick of [ARM-software/VectorSimilarity-for-Arm#4](https://github.com/ARM-software/VectorSimilarity-for-Arm/pull/4) (head `125ea15d`), plus a follow-up review pass. Fourth PR in the SQ8 series, after #999, #1000 and #1002.

Adds 8-bit scalar quantization (SQ8) to the standalone HNSW index:

* `VecSimQuantType` plus `quantType` / `quantParams` on `HNSWParams`. Both fields are appended at the end of the struct and `VecSimQuant_NONE` is 0, so existing zero-initialized and designated-initializer construction is unaffected.
* `HNSWFactory` can build SQ8 indexes for FLOAT32 and FLOAT16 data types with the L2 and IP metrics, wiring `QuantPreprocessor` and `DistanceCalculatorWithNorm`, and accounts for SQ8 in `EstimateInitialSize` and `EstimateElementSize`.
* For SQ8, `quantParams` points to a `float[dim]` mean vector; a null pointer selects quantization without mean normalization.
* New `test_hnsw_sq8` unit-test target and suite (44 tests: FP32/FP16 x L2/IP).

SQ8 support for the tiered HNSW index, serialization and benchmarks is deferred to later PRs in the series.

**Commit 1: the cherry-pick**

ARM's four commits squashed into one, with no functional change to their code. Two Redis-side adjustments:

* Dropped the added `SPDX-FileCopyrightText` Arm line from the two modified files, matching how #999, #1000 and #1002 landed. It is kept on the new `tests/unit/test_hnsw_sq8.cpp`, where the `BSD-3-Clause` identifier was replaced by this repo's Redis tri-license header.
* Wrapped that header so `make check-format` passes at the 100-column limit.

**Commit 2: review follow-up**

No behavioural change intended. Interface, single-source-of-truth and idiom fixes:

* `quantParams` is now `const void *`. Every use reads it and two already cast to `const float *`. Layout-identical, so not an ABI break, and callers passing non-const still compile. Better to fix before the field ships and freezes.
* The `VecSimQuant_SQ8` comment claimed "with mean normalization", but mean normalization is optional and selected by `quantParams`. Reworded.
[...]

PR #1011, opened 2026-08-10
[MOD-17526] Make the SQ8 metadata describe the reconstruction

dor-forer:

## The issue

SQ8 storage metadata held `sum = Σx[i]` and `sum_squares = Σx[i]²` over the **input** values. Every kernel is written over the **reconstruction** `x_r[i] = min + delta * a[i]`, which is what a quantized blob actually holds. The two differ by the quantization error, about 0.4% of `||x||²`. That is larger than the distance between two similar vectors, so L2 came back wrong and often negative.

Measured at dimension 128 on two near-duplicate vectors:

```
reference L2(x_r, y_r)    =  2.927985e-04
sums over x[i]   (before) = -1.486199e-01    negative, wrong by ~500x
sums over x_r[i] (after)  = +2.927985e-04    matches the reference
```

## The fix

Make the metadata describe `x_r`. Nothing else changes: no kernel is touched, and the blob keeps the same layout, slot count and FP32 slot types.

The existing algebra is already exact once the sums describe `x_r`:

```
IP = min1*sum2 + min2*sum1 - dim*min1*min2 + delta1*delta2*Σ(a[i]*b[i])
   = IP(x_r, y_r)     iff  sum    = Σx_r[i]  = dim*min + delta*q_sum

L2 = sum_sq_x + sum_sq_y - 2*IP
   = L2(x_r, y_r)     iff  sum_sq = Σx_r[i]²
```

So `quantize()` counts the quantized bytes as exact integers, then derives both sums from them in double and stores FP32 as before:

```cpp
sum         = dim*min + delta*q_sum;
sum_squares = dim*min*min + 2.0*min*delta*q_sum + delta*delta*q_sum_squares;
```

`q_sum` and `q_sum_squares` are local counters, not stored fields. `q_sum_squares` is `uint64_t`: each byte contributes up to `255²`, so a 32-bit counter passes `UINT32_MAX` just above dimension 66051, and a wrapped counter would corrupt the stored norm rather than round it.

`SUM` has exactly one reader, the inner product algebra above, which is what makes a metadata-only fix sufficient.

## What this does not fix

L2 is now correct, but not provably non-negative. Residual FP32 cancellation in `sum_sq_x + sum_sq_y - 2*IP` is about `1e-7 * ||x||²`, and two identical blobs have a true distance of zero, so self-distance lands just below it (measured `-4.9e-4` at dimension 512). The withdrawn revision below did guarantee non-negativity; this one does not. Tests bound the result by that noise floor rather than by zero, and the gap is recorded for follow-up.

## Storage and query describe different quantities

A query is never quantized, so its metadata keeps sums over the query itself: the asymmetric distance is `Σx_r² + Σy² - 2Σ(x_r * y)`. Three test assertions compared the query sum against the storage sum. They only ever agreed because both used to be the input sum, which is the same confusion this change fixes in the product code.

## Tests

`SQ8_SQ8_L2_is_non_negative_and_matches_reconstruction` and `SQ8_FP32_L2_matches_reconstruction_against_float_query`: near-duplicate vectors from dimension 4 to 512, scalar and dispatched kernels, against a double reference over `x_r`. Near-duplicates are the case that exposes this, because the true distance is small enough for the 0.4% mismatch to dominate.

The asymmetric test is new, and its absence is why the defect survived: the existing asymmetric L2 tests run dimensions 1, 5, 7 and 15 against an absolute tolerance of `0.01`, and 0.4% of `||x||²` at dimension 5 is about `0.007`, which fits underneath.

Tolerance is `8 * FLT_EPSILON * (norm_x + norm_y)`, scaled to the norms because the error floor is FP32 cancellation, which does not scale with the distance. A flat bound is thousands of times the distance it guards at small dimensions and would let a kernel returning zero pass.

`SQ8_SQ8_L2_self_distance_is_near_zero`: near zero, not exactly zero, for the reason above.

## Considered and rejected

An earlier revision made the L2 metadata integer (`uint32` `Q_SUM` / `Q_SUM_SQUARES`), added a `reconstructed_l2_sqr` helper regrouping the quadratic term so `S1 + S2 - 2Q` stays an exact integer, and rewrote 18 kernel files to combine in double. That buys exactly-zero self-distance and protection against FP32 cancellation when two vectors share a large offset, at 24 files, +630/-277, per-metric slot semantics, and 1 to 4.8 ns per distance call.

Those properties address the large-shared-offset case, not the defect above. This revision fixes the defect at zero per-call cost without touching a kernel. Carried forward: FP32 cancellation under a large shared offset, non-negative and exactly-zero self-distance, and the `MAX_EXACT_DIM` comment describing a bound the code does not have.

## Verification

CI green at `d5b467eb`: `basic tests`, `sanitizer / Test ubuntu-latest sanitizer` (ASan, both suites), `codeql-analysis / Analyze (cpp)`, `coverage / codecov job`, `codecov/patch`, `codecov/project`, `spellcheck`, Cursor Bugbot. Locally: `check-format`.

1. MOD-17526

1. `src/VecSim/spaces/computer/preprocessors.h`: the derivation and the metadata documentation.
2. `tests/unit/unit_test_utils.h`, `tests/utils/tests_utils.h`: the two test mirrors of the quantizer.
3. `tests/unit/test_spaces.cpp`, `tests/unit/test_components.cpp`.

**Mark if applicable**

---

> [!NOTE]
> **Medium Risk**
> Changes core SQ8 quantization metadata used on every indexed vector and L2 search path, but scope is limited to preprocessor/test quantizers with no kernel or layout change; incorrect sums would break distance ordering.
> 
> **Overview**
> **Fixes wrong and often negative SQ8 L2 distances** by making storage metadata match what the distance kernels assume: `sum` and `sum_squares` are over the **reconstructed** vector `x_r[i] = min + delta * a[i]`, not the pre-quantization input. The mismatch was on the order of ~0.4% of `||x||²`, which could dominate true distance between similar vectors.
> 
> `QuantPreprocessor::quantize()` now accumulates **quantized byte** sums (with `uint64_t` for squared-byte totals on L2), then derives `sum` and `sum_squares` in double and stores FP32 metadata as before. **Blob layout and distance kernels are unchanged.**
> 
> Tests and test quantizer helpers (`ComputeSQ8Quantization`, `quantize_float_vec_to_sq8_with_metadata`) follow the same rules. FP16 preprocessor tests assert query metadata against **input** sums, not storage reconstruction sums. New `test_spaces` cases check SQ8↔SQ8 and SQ8↔FP32 L2 against a double reference on near-duplicate vectors, with tolerances tied to FP32 cancellation noise rather than a loose absolute bound.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e96910facda1b727c5728390ac9ccb8cb45a4a56. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/VecSim/spaces/computer/preprocessors.h
+++ b/src/VecSim/spaces/computer/preprocessors.h
@@ -293,13 +293,17 @@ class QuantPreprocessor : public PreprocessorInterface {
             return static_cast<OUTPUT_TYPE>(scaled + MetadataType{0.5});
         };
 
-        // Compute sum (and sum of squares for L2) while quantizing.
-        // Accumulators are FP32 to preserve metadata precision for FP16 inputs.
-        // 4 independent accumulators (sum)
-        float s0{}, s1{}, s2{}, s3{};
-
-        // 4 independent accumulators (sum of squares), only used for L2
-        float q0{}, q1{}, q2{}, q3{};
+        // Sum the quantized bytes, not the input values. Every kernel term is written in terms of
+        // the reconstruction x_r[i] = min + delta * a[i], so the stored sums have to describe x_r
+        // as well. Summing the input instead leaves a systematic mismatch of about 0.4% of ||x||^2,
+        // which is larger than a small true distance and made L2 come out negative: for two
+        // near-duplicate vectors at dim 128 the reconstruction distance is 2.93e-04 and the kernels
+        // returned -1.49e-01. The byte sums are exact integers, so the derivation below is the only
+        // place rounding enters.
+        //
+        // 4 independent accumulators each, so the unrolled loop keeps four dependency chains.
+        uint32_t s0{}, s1{}, s2{}, s3{};
+        uint32_t q0{}, q1{}, q2{}, q3{}; // only used for L2
 
         size_t i = 0;
         // round dim down to the nearest multiple of 4
@@ -315,38 +319,54 @@ class QuantPreprocessor : public PreprocessorInterface {
             // We know (input - min) => 0
             // If min == max, all values are the same and should be quantized to 0.
             // reconstruction will yield the same original value for all vectors.
-            quantized[i] = to_byte((x0 - min_val) * inv_delta);
-            quantized[i + 1] = to_byte((x1 - min_val) * inv_delta);
-            quantized[i + 2] = to_byte((x2 - min_val) * inv_delta);
-            quantized[i + 3] = to_byte((x3 - min_val) * inv_delta);
+            const OUTPUT_TYPE a0 = to_byte((x0 - min_val) * inv_delta);
+            const OUTPUT_TYPE a1 = to_byte((x1 - min_val) * inv_delta);
+            const OUTPUT_TYPE a2 = to_byte((x2 - min_val) * inv_delta);
+            const OUTPUT_TYPE a3 = to_byte((x3 - min_val) * inv_delta);
+            quantized[i] = a0;
+            quantized[i + 1] = a1;
+            quantized[i + 2] = a2;
+            quantized[i + 3] = a3;
 
-            // Accumulate sum for all metrics
-            s0 += x0;
-            s1 += x1;
-            s2 += x2;
-            s3 += x3;
+            s0 += a0;
+            s1 += a1;
+            s2 += a2;
+            s3 += a3;
 
-            // Accumulate sum of squares only for L2 metric
             if constexpr (Metric == VecSimMetric_L2) {
-                q0 += x0 * x0;
-                q1 += x1 * x1;
-                q2 += x2 * x2;
-                q3 += x3 * x3;
+                q0 += uint32_t{a0} * a0;
+                q1 += uint32_t{a1} * a1;
+                q2 += uint32_t{a2} * a2;
+                q3 += uint32_t{a3} * a3;
             }
         }
 
         // Tail: 0..3 remaining elements (still the same pass, just finishing work).
-        // Sum/sum_squares become metadata, so they are MetadataType.
-        MetadataType sum = (s0 + s1) + (s2 + s3);
-        MetadataType sum_squares = (q0 + q1) + (q2 + q3);
+        uint32_t q_sum = (s0 + s1) + (s2 + s3);
+        uint32_t q_sum_squares = (q0 + q1) + (q2 + q3);
 
         for (; i < this->dim; ++i) {
             const float x = transformed_value(input, i);
-            quantized[i] = to_byte((x - min_val) * inv_delta);
-            sum += x;
+            const OUTPUT_TYPE a = to_byte((x - min_val) * inv_delta);
+            quantized[i] = a;
+            q_sum += a;
             if constexpr (Metric == VecSimMetric_L2) {
-                sum_squares += x * x;
+                q_sum_squares += uint32_t{a} * a;
             }
+        }
+
+        // Derive the reconstruction sums from the exact byte sums, in double so the expansion does
+        // not lose the cross term, then store FP32 as before. The slot layout and types are
+        // unchanged; only what the numbers describe changes.
+        //   sum         = sum(x_r[i])   = dim*min + delta*q_sum
+        //   sum_squares = sum(x_r[i]^2) = dim*min^2 + 2*min*delta*q_sum + delta^2*q_sum_squares
+        const double d_min = min_val, d_delta = delta, d_dim = static_cast<double>(this->dim);
+        const MetadataType sum = static_cast<MetadataType>(d_dim * d_min + d_delta * q_sum);
+        MetadataType sum_squares{};
+        if constexpr (Metric == VecSimMetric_L2) {
+            sum_squares =
+                static_cast<MetadataType>(d_dim * d_min * d_min + 2.0 * d_min * d_delta * q_sum +
+                                          d_delta * d_delta * q_sum_squares);
         }
 
         // Metadata uses MetadataType. Use memcpy because the metadata offset
--- a/src/VecSim/spaces/computer/preprocessors.h
+++ b/src/VecSim/spaces/computer/preprocessors.h
@@ -152,9 +152,14 @@ private:
  *
  * Storage layout:
  * | quantized_values[dim] | min_val | delta | x_sum | (x_sum_squares for L2 only) |
- * where:
- * x_sum = Σx_i: sum of the original values,
- * x_sum_squares = Σx_i²: sum of squares of the original values.
+ * where, writing x_r[i] = min_val + delta * a[i] for the value a blob actually represents:
+ * x_sum        = Σx_r[i]:  sum of the reconstructed values,
+ * x_sum_squares = Σx_r[i]²: sum of their squares.
+ *
+ * These describe x_r, not the input x. Every formula below is written in terms of x_r, because
+ * that is what a quantized blob holds, so sums over the input would not satisfy them: the gap is
+ * the quantization error, about 0.4% of ||x||², which is larger than the distance between two
+ * similar vectors and made L2 come out negative.
  *
  * Storage metadata is always FP32 (independent of DataType) to match the asymmetric distance
  * kernels. The quantized blob size is:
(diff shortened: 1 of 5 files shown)


lerman25 (commented):
Nice - 1 blocking comment

src/VecSim/spaces/computer/preprocessors.h
```diff
@@ -269,18 +281,40 @@ class QuantPreprocessor : public PreprocessorInterface {
...
-        // 4 independent accumulators (sum)
-        float s0{}, s1{}, s2{}, s3{};
-
-        // 4 independent accumulators (sum of squares), only used for L2
-        float q0{}, q1{}, q2{}, q3{};
+        // The range is computed in double, and so is the per-element scaling below, because the
+        // FP32 arithmetic could leave the representable range for input that is entirely valid.
+        // [-FLT_MAX, +FLT_MAX] made max - min overflow to inf, then delta inf, inv_delta 0, and
+        // finally inf * 0 = NaN, whose conversion to an integer is undefined behaviour. Doubles
+        // cannot overflow for any pair of finite floats, so the whole class disappears rather than
```
lerman25: Blocking: `WithNorm` can still make `min_val`/`max_val` non-finite *before* this double range calculation. Both `find_min_max()` and `transformed_value()` compute `input[i] - mean[i]` in FP32. For example, finite FP32 input `[FLT_MAX, 0]` with finite mean `[-FLT_MAX, 0]` centers to `[+Inf, 0]`; this then gives `diff = Inf`, `delta = Inf`, and `inv_delta = 0`, so `to_byte(+Inf)` evaluates `Inf * 0` as NaN. `std::clamp` preserves NaN, and the following conversion to `uint32_t` is undefined behavior. This is reachable by the mean-centred SQ8 configuration introduced by #1007, so the finite-input safety claim is incomplete. Please perform/check centering in a representation that cannot overflow here (or reject non-finite derived values before quantization) and add a UBSan regression for this case.

dor-forer: Agreed, and fixed in 6c8d6e0d. You are right that the double range does not help when the value is already lost upstream in FP32.

`find_min_max` now clamps its endpoints to the float range. That is also a storage constraint rather than only a guard: `min_val` is stored as FP32, so a non-finite endpoint could not be stored under any arithmetic. Only the endpoints are clamped, so the per-element loop stays FP32 and pays nothing.

With `min_val` finite and `inv_delta` finite and positive, NaN is unreachable in `to_byte`: a centered `inf` element gives `inf * finite = inf`, not `0 * inf`, and lands on 255. I also moved the bound from `std::clamp` to `std::fmin`/`std::fmax`, since clamp propagates NaN, so the conversion is defined without relying on a proof that spans two functions.

I went with clamping rather than centering in double. The tradeoff: it puts overflowing elements at the ends instead of placing them proportionally. Happy to switch to double centering if you want proportionality, but it costs a conversion per element and the FP32 `min` slot still cannot hold the range.

UBSan regression added: `QuantizationHandlesNonRepresentableCenteredRange`, with your exact input.

src/VecSim/spaces/computer/preprocessors.h
```diff
@@ -269,18 +281,40 @@ class QuantPreprocessor : public PreprocessorInterface {
...
+        // clamped so that a value outside [0, 255] can never reach the conversion: the arithmetic
+        // should not produce one, and if it ever does the result must be a wrong byte rather than
+        // undefined behaviour.
+        const auto to_byte = [inv_delta, min_val_d](float x) {
+            const double scaled = (static_cast<double>(x) - min_val_d) * inv_delta;
+            // Clamped first, so the value is non-negative and adding 0.5 before truncating is
+            // std::round without the library call. This translation unit is compiled at the
+            // x86-64 baseline, where std::round is an out-of-line call to round() rather than a
+            // roundsd instruction, and this runs once per element.
+            return static_cast<uint32_t>(std::clamp(scaled, 0.0, 255.0) + 0.5);
```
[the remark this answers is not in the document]
dor-forer: Confirmed and fixed in 6c8d6e0d. Reproduced with `[0.0f, 1e-44f]`: `diff` is not zero so the old check did not fire, `(float)(diff / 255.0)` underflowed to `0`, and the minimum element scaled to `0 * inf` = NaN.

The guard now tests `delta` rather than the range it came from, since `delta` is the value that gets stored and inverted and it has the narrower type. It subsumes the old `min == max` check, and `delta = 1` is the same degenerate handling that case already got.

Correct that `std::clamp` does not replace NaN. The bound is now `std::fmin`/`std::fmax`, which return the non-NaN operand.

UBSan regression added: `QuantizationHandlesSubnormalRange`.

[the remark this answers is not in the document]
dor-forer: Following up, because two things in my earlier reply are no longer accurate.

**The underflow guard stands.** `delta` is still tested after narrowing rather than `diff` before it, which is the substance of your finding and it was correct.

**Correction 1: the bound is no longer `std::fmin`/`std::fmax`.** You were right that `std::clamp` does not replace NaN, but `fmin`/`fmax` were the wrong replacement: at this translation unit's baseline they compile to **two out-of-line libm calls per element**, the same cost `std::round` was removed for. The bound is now written by hand:

```cpp
const double bounded = scaled > 0.0 ? scaled : 0.0;
return static_cast<OUTPUT_TYPE>((bounded > 255.0 ? 255.0 : bounded) + 0.5);
```

Measured at `-O3`: 8 instructions, no libm call, against 9 for `std::clamp` plus `std::round`'s call. So it is cheaper than the *non*-defensive form it replaced.

**Correction 2: the NaN-safety framing was wrong, and I withdraw it.** `find_min_max` uses `std::minmax_element`, whose precondition is that the comparison induce a strict weak ordering. Floating-point `<` is not one once a NaN is present, because incomparability stops being transitive: `1.0` and `NaN` are incomparable, `NaN` and `2.0` are incomparable, but `1.0` and `2.0` are not. Violating that precondition is undefined behaviour *inside the algorithm*, before any range exists, so nothing downstream can offer a portable guarantee about NaN input, and UBSan reporting nothing does not establish otherwise.

The contract is therefore narrowed to **finite input**, non-finite components are documented as unsupported, and the test that asserted metadata finiteness for NaN input has been removed rather than kept. Rejecting non-finite components at the public ingestion boundary is the real fix and is filed separately.

**Tests:** the underflow case is now the `denorm_min_range_collapses` row of a table-driven matrix, and `subnormal_delta_survives` (`[0, 7e-37]`) covers the neighbouring case the FP64 reciprocal exists for, where `delta` is subnormal but representable and the endpoints must still reach 0 and 255.

**Location:** this work has been extracted into #1015, which is where it will land. Once this PR is restacked it will contain only the SQ8 metadata change, so the code these comments point at will not be here.

lerman25 (approved):
LG

src/VecSim/spaces/computer/preprocessors.h
dor-forer: Correcting the record on this thread, because my earlier reply points at a commit that is no longer here.

That reply said this was fixed in `6c8d6e0d`, which clamped the `find_min_max` endpoints. This branch was since rewritten to a much smaller change, and `6c8d6e0d` is not an ancestor of the current head. So the fix it describes is not in the code, and the thread should not be resolved on the strength of it.

**The undefined behaviour you identified is closed, by a different mechanism.** Tracing your exact input through the current head:

```
input [FLT_MAX, 0], mean [-FLT_MAX, 0]   ->  centered [+inf, 0]
find_min_max -> (0, inf)
diff = inf,  delta = inf,  inv_delta = 1/inf = 0
to_byte((inf - 0) * 0) = to_byte(inf * 0) = to_byte(NaN)
```

and `to_byte`, from #1015 which is now on `main`:

```cpp
if (!(scaled > MetadataType{0})) {
    return OUTPUT_TYPE{0};   // zero, negative, -inf, or NaN
}
```

`!(NaN > 0)` is true, so NaN takes the first branch and returns 0. Defined, with no conversion of a non-finite value. This works because the safe constant sits in the *false* branch of the comparison, which is also why that code uses ternaries rather than `std::clamp` or `std::fmin`/`std::fmax` — as you noted, `std::clamp` propagates NaN, and `fmin`/`fmax` propagate it too in this composition, besides costing two libm calls per element.

So the endpoint clamping was one way to reach a defined conversion, and the merged conversion guard is another. The regression test with your input, `QuantizationHandlesNonRepresentableCenteredRange`, is present on this branch and passing.

**What is not fixed, and I do not want to imply otherwise.** The metadata in that case is still degenerate: `delta` is infinite, so the whole vector quantizes to one value. That is defined but useless, and it is outside this PR, which changes only what the sums describe. It is tracked in MOD-17838 along with the rest of the intermediate-overflow policy: `max - min` overflow, `input - mean` overflow, `delta` underflow and reciprocal overflow, the FP32 sum accumulations, and whether non-finite input should be rejected at the public boundary instead. Item 1 there is the root question, since rejecting non-finite input at ingestion would make all of these unreachable.

Also worth flagging on this PR as it now stands: it does **not** make L2 provably non-negative. Self-distance comes out at up to about 2.6 float epsilons below zero, measured at `-9.77e-04` at dimension 512 against `||x_r||^2 ~ 3e+03`. That is down from a systematic `-3.08` at the same dimension, roughly four orders of magnitude better, but it is not a sign guarantee, and the withdrawn integer-metadata revision did provide one. If a caller takes `sqrt` of the result, that matters. A one-line `std::max(0.0f, ...)` in the L2 kernels would give the guarantee, and is only defensible now that the systematic error is gone: on `main` today it would have masked the real 0.4% error instead of float noise. Happy to add it here or as a follow-up, whichever you prefer.

PR #1012, opened 2026-08-13
[MOD-17578] Isolate a deleted element from the graph once its repairs are done

nonirosenfeldredis:

A marked-deleted element used to keep its edges until its swap job disposed of it: `removeFromGraph` was the one to take it out of its neighbours' incoming edges sets, under the exclusive main index guard.

This PR takes the element out of the graph in the **repair context** instead, as soon as the last repair job created for its deletion completes - dropping every edge in and out of it (`isolateDeletedElement`), so that by the time the swap job runs there is nothing left to disconnect. An element that never had any repair job is isolated right away, upon marking it deleted.

Each edge is removed *mutually*: both elements' links locks are taken, in ascending id order as the other multi-lock sites here do, and both sides of the edge are dropped in one step. That leaves no observable point at which an edge is recorded on neither side, so `mutuallyRemoveNeighborAtPos` can keep telling a bidirectional edge from a unidirectional one by the record alone - no extra scan on a path that every insertion and every repair takes.

One behavioural prerequisite: the pending repair jobs counter is now decreased **after** the repair has been performed, not before. Otherwise a swap job could be seen as ready - and its element isolated - while an element still pointed at it from a repair that had not run yet. The only edges left for isolation to drop are then edges between two deleted elements, since neither of them gets a repair job for the other; that is asserted rather than silently handled (measured while developing this: every edge reaching those branches had the other side marked deleted, none live).

Worth noting for reviewers: a deleted element now holds no edges while it waits for its swap job, so **it no longer generates repair jobs for subsequent deletions**, and insertion traversals can no longer reach it at all (nothing points to it). `swapJobBasic2` was retraced accordingly - one repair job instead of two, and the two "invalid repair job" steps it used to produce no longer occur - and a new `invalidRepairJobOnSwap` test covers the invalidation of a pending repair job whose node is disposed of, reached by deleting a second element while the first is still connected.

1. MOD-17578

1. `HNSWIndex::isolateDeletedElement` and the `hasLink` helper (new)
2. `TieredHNSWIndex::executeRepairJob` / `deleteLabelFromHNSW` - isolation call sites and the counter decrement moved after the repair
3. `tests/unit/test_hnsw_tiered.cpp` - `swapJobBasic2` retraced, `invalidRepairJobOnSwap` added

**Mark if applicable**

Note on the unchecked boxes: no API or serialization *format* change, but the graph content does change - a marked-deleted element is serialized with no edges. Anything comparing dumped graphs across versions will see that.

---

> [!NOTE]
> **High Risk**
> Changes concurrent tiered HNSW delete/repair/swap ordering and graph invariants under per-element locks; incorrect timing could leave dangling edges or premature swaps.
> 
> **Overview**
> **Async tiered HNSW deletion** now strips a marked-deleted node from the graph as soon as its deletion-related repair work finishes, instead of keeping edges until the swap job runs.
> 
> `HNSWIndex` adds **`isolateDeletedElement`** (mutual edge removal at every level, with `hasLink` and ordered per-node locks) and splits the old **`removeAndSwap`** path into **`removeFromGraph`** (incoming-edge cleanup + destroy; often a no-op on outgoing edges after isolation) and **`swapWithLast`** (slot reclaim). **`removeAndSwapMarkDeletedElement`** chains those two steps.
> 
> **`TieredHNSWIndex`** calls isolation when a deleted id has zero pending repairs (including immediately on delete if no repair jobs were created) and again when the last repair job for a swap job completes. The swap-job pending-repair counter is decremented **after** `repairNodeConnections` runs so a swap cannot be considered ready while a neighbor still points at the deleted id. Swap bookkeeping is split: **`fixJobsAfterSwap`** (renamed from **`executeSwapJob`**) only remaps job ids; **`invalidateRepairJobs`** runs when the physical swap disposes an id.
> 
> Tests retrace **`swapJobBasic2`** for fewer repair/invalid jobs, add **`invalidRepairJobOnSwap`**, and unit-test isolation edge shapes between two deleted nodes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1152c87ee6bf59118fe0a4a3e8f0c2e44b235718. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/VecSim/algorithms/hnsw/hnsw.h
```diff
@@ -1541,7 +1562,20 @@ void HNSWIndex<DataType, DistType>::mutuallyRemoveNeighborAtPos(ElementLevelData
...
+        // back before recording the remaining direction.
+        bool points_back = false;
+        for (size_t i = 0; i < removed_node_level.getNumLinks(); i++) {
+            if (removed_node_level.getLinkAtPos(i) == node_id) {
+                points_back = true;
+                break;
+            }
+        }
+        if (points_back) {
+            node_level.newIncomingUnidirectionalEdge(removed_node);
```
alonre24: This is a hot code path called many times for every vector insertion/repair job, and we are adding another loop over the node's `M` neighbours. This might have a significant performance impact. Let's see if we can avoid this addition, and if not, we would need to assess the penalty for large `M` use cases.

src/VecSim/algorithms/hnsw/hnsw.h
```diff
@@ -1644,20 +1678,73 @@ HNSWIndex<DataType, DistType>::~HNSWIndex() {
...
+    for (size_t level = 0; level <= element->toplevel; level++) {
+        // Take the element's edges at this level, and drop its own side of them right away - from
+        // here on it points to nothing, so a repair job that runs for it later finds no deleted
+        // neighbour and returns without touching it.
+        lockNodeLinks(internalId);
+        ElementLevelData &level_data = getElementLevelData(element, level);
+        auto neighbours = level_data.copyLinks();
+        std::vector<idType> incoming_edges(level_data.getIncomingEdges().begin(),
+                                           level_data.getIncomingEdges().end());
+        level_data.setNumLinks(0);
```
alonre24: Can we make this operation mutual? That is, in every iteration, acquire both locks (of the isolated node and the neighbor) in sorted order and update the edge removal. That might reduce the need to iterate over the node's neighbors in the remove-neighbor hot path.

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -1949,6 +1930,76 @@ TYPED_TEST(HNSWTieredIndexTest, swapJobBasic2) {
...
+
+    // Delete 0 - a repair job is created for every (node, level) pair that points to it. Note that
+    // the number of jobs depends on the levels the elements got, so the counters are compared to
+    // each other rather than to fixed values below.
+    EXPECT_EQ(tiered_index->deleteVector(0), 1);
+    ASSERT_GT(mock_thread_pool.jobQ.size(), 0);
+    ASSERT_GT(tiered_index->idToSwapJob.at(0)->pending_repair_jobs_counter.load(), 0);
+
+    // Delete 1 before those jobs run. 0 is deleted but still connected (its own repairs are
+    // pending), so a repair job for node 0 is created here and stays pending.
```
alonre24: nit: denoting the repair job as u->v is easier for me to track (rather than "repair node for v")

alonre24 (commented):
LGTM - not that scarry as I thought ;)
Isolation logic seems solid - see my comment about the hot code path potnetial regression

src/VecSim/algorithms/hnsw/hnsw.h
nonirosenfeldredis: done

tests/unit/test_hnsw_tiered.cpp
nonirosenfeldredis: done

src/VecSim/algorithms/hnsw/hnsw.h
nonirosenfeldredis: done

src/VecSim/algorithms/hnsw/hnsw.h
```diff
@@ -1644,20 +1662,89 @@ HNSWIndex<DataType, DistType>::~HNSWIndex() {
...
+            // Else the edge is already gone - a repair job of another element got to it first.
+
+            unlockNodeLinks(second);
+            unlockNodeLinks(first);
+        }
+
+        lockNodeLinks(internalId);
+        assert(level_data.getNumLinks() == 0 && level_data.getIncomingEdges().empty() &&
+               "the element should have no edge left at this level");
+        unlockNodeLinks(internalId);
```
alonre24, line 1729: This is a test-level assert (not running in production), so it means that locks are held for nothing in prod. Consider wrapping this with the build-time `#ifdef build tests` macro so the code will run only in tests

nonirosenfeldredis: done

alonre24 (approved):

PR #1014, opened 2026-08-16
[MOD-17527] Make the uint8 integer accumulators exact at any dimension

dor-forer:

The uint8 kernels accumulate byte products or squared differences, so the worst-case total is `255 * 255 * dim = 65025 * dim`. Three defects follow, all on the plain `uint8` paths that ship today.

**Defect 1, wrong results from dimension 33,026.** The totals are held in 32-bit signed integers (`int32_t result = vaddvq_u32(total_sum)` on ARM, `_mm512_reduce_add_epi32` on x86, `ret_t = int` for one-byte types in the scalar path), so once the total passed `INT32_MAX` the result wrapped negative. At dimension 33,026 with all-255 bytes the inner product returned `+2,147,451,647` where the answer is `-2,147,515,649`: a sign flip, so the most similar pair looks like the most distant. Both the SIMD and scalar paths were affected, so there was no correct path to fall back to.

**Defect 2, an illegal instruction on ARM.** `IP_NEON_UINT8.h` and `IP_NEON_DOTPROD_UINT8.h` both defined a template named `UINT8_InnerProductImp`, same signature, different bodies, and both are compiled into every ARM build. Measured on Graviton4 with gcc 12.3 at `-O2`, on `main`:

```
NEON.cpp.o:         64 weak UINT8_InnerProductImp symbols,  0 udot
NEON_DOTPROD.cpp.o: 64 weak UINT8_InnerProductImp symbols,  1284 udot
identical mangled names: 64 of 64
```

Which body survives is decided by link order, nothing else:

```
link order NEON first:     surviving body has umull(4) umull2(4) uadalp(8),  udot 0
link order DOTPROD first:  surviving body has udot(4),                       umull 0
```

In the second order the plain-NEON inner product and cosine execute `udot`, which does not exist on a core without `asimddp`. Because it depends on link order rather than on source, it can flip between builds.

`INT8_InnerProductImp` is the same defect one datatype over, in the same two directories, with `sdot` in place of `udot`. Measured on this branch after the uint8 fix and before the int8 one, all 64 remaining collisions between those two objects were the int8 twin and zero were uint8, which is both evidence the uint8 fix works and evidence it did half the job. `IP_space.cpp` selects the plain-NEON int8 chooser on `features.asimd` alone, so the consequence is identical. Both are fixed here.

**Defect 3, the cosine norm, raised in review.** `IntegralType_ComputeNorm` accumulated into a signed `int`. That is the norm the cosine preprocessor writes for every stored vector and every query, so a wrong value reaches the kernels before they run: dimension 65,537 produced a NaN norm and 66,052 produced 252.99 instead of about 65,536.49. This is independent of which kernel runs. int8 had the same shape, its largest squared term being `(-128)^2 = 16,384`, overflowing from dimension 131,072.

**The rule**

```
UINT8_MAX_EXACT_SIMD_DIM = INT32_MAX / (UINT8_MAX * UINT8_MAX) = 33,025
```

Derived from the types rather than written as a literal. At 33,025 the worst case is 2,147,450,625, the largest total a signed 32-bit accumulator holds; at 33,026 it is 2,147,515,650, which it does not. Six choosers return the scalar kernel above it, once per index, so no distance computation pays for the check.

**Why no kernel changed.** Bounding at the signed limit means `main`'s kernel arithmetic is already correct: every total the cap admits fits the `int32_t` those kernels already use. So this PR changes no kernel arithmetic, no return types and no reduces. The complete functional diff is thirteen lines:

```
2x  using ret_t = long long;                    IP.cpp, L2.cpp
3x  static float UINT8_InnerProductImp(...)     NEON, NEON_DOTPROD, SVE
6x  if (dim > spaces::UINT8_MAX_EXACT_SIMD_DIM) return ret_dist_func;
1x  long long sum = 0;                          compute_norm.h
1x  the constant                                spaces.h
```

`ret_t` becomes `long long` for every element type rather than `int` for one-byte ones, because the scalar kernel is now the fallback above the cap and has to be exact there. Keeping it signed means the wrappers keep `main`'s `1 - ip` form with no cast.

1. MOD-17527

1. `spaces.h`: `UINT8_MAX_EXACT_SIMD_DIM`.
2. `IP_space.cpp` and `L2_space.cpp`: six guards, before any architecture branch.
3. `IP.cpp` and `L2.cpp`: `ret_t`.
4. `spaces/normalize/compute_norm.h`: the norm accumulator.
5. `IP_NEON_UINT8.h`, `IP_NEON_DOTPROD_UINT8.h`, `IP_SVE_UINT8.h` and the three matching `INT8` headers: internal linkage.
6. `tests/unit/test_spaces.cpp`, `tests/unit/test_common.cpp`.

**Mark if applicable**

---

## Why six guards and not three

The `SQ8_SQ8` kernels call `UINT8_InnerProductImp` directly and never pass through a uint8 chooser, so a uint8-only guard would leave them able to wrap. They have their own dimension-aware choosers, so `IP_SQ8_SQ8`, `Cosine_SQ8_SQ8` and `L2_SQ8_SQ8` take the same guard. Their scalar fallback accumulates the quantized dot product in `float`, which loses precision past 2^24 but is well defined, so the guard trades a wrapped total for a lossy one.

## Which linkage collisions this closes, and which it does not

Five helpers get internal linkage: `UINT8_InnerProductImp` in the NEON, NEON_DOTPROD and SVE headers, and `INT8_InnerProductImp` in the NEON, NEON_DOTPROD and SVE headers.

**Fully closed: the NEON / NEON_DOTPROD pair.** Those headers are compiled into exactly one translation unit each, so `static` removes the shared symbol outright with no duplication cost. Verified on Graviton4: shared weak symbols between `NEON.cpp.o` and `NEON_DOTPROD.cpp.o` went from 64 to 0, `NEON.cpp.o` has zero `sdot` and keeps its 472 `smull`, and `NEON_DOTPROD.cpp.o` keeps all 428 `sdot`.

**Only partially closed: the SVE / SVE2 pair.** `IP_SVE_UINT8.h` and `IP_SVE_INT8.h` are each compiled into both `SVE.cpp` (`-march=armv8-a+sve`) and `SVE2.cpp` (`-march=armv9-a+sve2`), so making the helper `static` removes the helper's collision but leaves the wrappers that call it colliding, and those wrappers now differ more than before since each refers to its own unit's helper. Measured on this branch, `SVE.cpp.o` and `SVE2.cpp.o` still share 160 weak symbols, including `UINT8_InnerProductSIMD_SVE`, `UINT8_CosineSIMD_SVE`, `UINT8_L2SqrSIMD_SVE` and their int8 counterparts, 8 instantiations each.

That residue is latent rather than live, which I checked rather than assumed: across all shared SVE/SVE2 symbols both compilations use identical instruction sets, 57 distinct opcodes each with no difference in either direction, so no SVE2-only instruction can reach an SVE-only core. Closing it means giving the wrappers internal linkage too, across IP, L2 and Cosine for every data type, which is MOD-17759 rather than this PR.

## Two designs implemented and then withdrawn

The branch history shows both. First, chunked accumulation, exact at any dimension, validated on Graviton4 and Sapphire Rapids: withdrawn because it cost a shared driver, eight chunked wrappers, eight adapters, eight out-of-line helpers and a large amount of architecture-specific test machinery to serve dimensions that do not occur. Then a bound at `UINT32_MAX / 65025 = 66,051`, which required replacing `_mm512_reduce_add_epi32` with a zero-extend-then-reduce-in-64-bit sequence in every AVX-512 uint8 epilogue, because GCC implements that intrinsic as signed vector operations ending in a scalar `int + int`. An earlier revision measured unconditional widening at +20% at dim 32, +8 to 11% across 55 to 200 and +4 to 5% at 900 to 1024 on an Ice Lake-SP Xeon; that figure has not been re-measured against this head, so treat it as indicative. Either way it bought only the band 33,026 to 66,051, which no workload reaches given embeddings run 384 to 4096.

Chunking above the cap remains available as MOD-17731 if a real high-dimensional uint8 workload appears.

## Tests

One boundary, pinned from both sides. 33,025 stays on SIMD; 33,026 must come back as the scalar kernel by name. Two `static_assert`s pin the bound arithmetically rather than trusting the constant: the worst case at the bound must fit `INT32_MAX`, and must come within one element's product of it, so the test fails if the constant ever drifts away from the limit it represents.

`UINT8_every_tier_is_exact_up_to_the_dispatcher_cap` calls every compiled-in chooser directly rather than relying on dispatch, which only ever returns the best tier the host supports, so NEON is exercised on a machine that also has DOTPROD. Dimensions 32,960 / 32,993 / 33,023 / 33,024 / 33,025, all residual-bearing, so each tier is exercised at its own tail handling right up to the cap. This is the test that validates the cap's placement, since "SIMD is exact up to 33,025" is the premise the whole design rests on.

Both tier-aware tests report which tiers they reached and fail rather than skip when `VECSIM_REQUIRE_UINT8_TIER` names a tier they did not reach. A run that exercised no SIMD tier reports skipped rather than passed.

The norm defect is covered through the public API rather than by calling the helper, since the kernel tests append the norm by hand and so never exercised that path: `VecSim_Normalize` at 33,025 / 33,026 / 65,537 / 66,052, and a uint8 cosine brute force index storing and querying itself at 65,537 where the self-distance must be about zero.

## Verification

**Xeon Platinum 8375C, at this head.** 1035/1035 across the uint8, spaces, common and SQ8 suites, with `AVX512F_BW_VL_VNNI` reported at all five boundary dimensions, so the cap is exercised on SIMD rather than inferred.

**ARM, pending re-run at this head.** The earlier Graviton4 validation (1428/1428, all four tiers reported) was measured against the 66,051 boundary and the tier tests now use different dimensions, so those numbers do not carry over and are not claimed here.

**The `udot` defect and its fix**, by disassembly on Graviton4 with the real toolchain, as tabulated above. That measurement is independent of the boundary and still stands.

## Known, not fixed here

**Dimensions above 33,025 give up SIMD** and run 6 to 15 times slower on the scalar kernel, measured at dimension 131,072. That is the accepted cost of not carrying a second code path for dimensions no workload reaches.

**The `udot` fault has not been observed faulting.** That it does follows from `udot` not existing on Armv8.0-A, but it has not been executed on such a core. Graviton2 and later all implement the dot product extension, so this needs Graviton1 or equivalent. Every other link in that chain is measured.

**SQ8 above the cap is imprecise rather than exact**, because its scalar fallback accumulates in `float`. Bounding the index dimension explicitly is MOD-17732.

**The SVE/SVE2 wrapper collisions remain**, 160 shared weak symbols including the uint8 and int8 SVE wrappers. Benign today, measured rather than assumed: identical instruction sets in both compilations. MOD-17759.

---

> [!NOTE]
> **Medium Risk**
> Changes core uint8 IP/L2/cosine dispatch and normalization for large dimensions; above the cap queries use slower scalar paths but avoid wrong distances from int32 wrap and bad norms.
> 
> **Overview**
> Fixes **uint8** (and related) vector distance correctness when dimension or norms exceed 32-bit signed accumulator range, without changing SIMD kernel arithmetic.
> 
> Introduces **`UINT8_MAX_EXACT_SIMD_DIM` (33,025)** in `spaces.h` from `INT32_MAX / (255×255)`. **Six distance choosers** in `IP_space.cpp` and `L2_space.cpp` return the **scalar** IP/L2/cosine (and SQ8_SQ8) kernels when `dim` is above that cap, because SIMD paths reduce into **signed 32-bit** totals that wrap past dimension 33,026.
> 
> Scalar **`ret_t`** for integer IP/L2 in `IP.cpp` and `L2.cpp` is always **`long long`** so the fallback stays exact at large dimensions. **`IntegralType_ComputeNorm`** uses **`uint64_t`** so cosine preprocessing no longer overflows `int` on large uint8 vectors.
> 
> **`static`** on duplicate **`UINT8_InnerProductImp` / `INT8_InnerProductImp`** templates in NEON, NEON_DOTPROD, and SVE headers removes link-time ODR collisions (including wrong `udot`/`sdot` bodies on cores without dotprod).
> 
> Tests add dispatcher cap checks, near-cap SIMD vs scalar parity, **`VecSim_Normalize`** at large dims, and uint8 cosine self-query at 65,537.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bdbdf38d76607f50b367c34c846ee5b06e0f9302. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/VecSim/spaces/spaces.h
```diff
@@ -52,6 +52,27 @@ static int inline is_little_endian() {
...
+// forms and the chooser selects between them once per index.
+static constexpr size_t UINT8_NARROW_REDUCE_MAX_DIM = 66051;
+
+// The *lanes* are what bounds the widened form, and widening the final sum does not protect them.
+// NEON is the limiting ISA: it combines four accumulators with vaddq_u32 in 32 bits before any
+// widening reduce sees them, so its capacity is four lanes' worth. AVX-512 accumulates into 16
+// lanes from one accumulator and reaches roughly 1,056,816, and SVE depends on its vector length,
+// so NEON sets the shared bound. Above this the choosers hand back the scalar kernel, which is
+// exact because ret_t is 64-bit.
+static constexpr size_t UINT8_SIMD_MAX_DIM = 4 * UINT8_NARROW_REDUCE_MAX_DIM;
```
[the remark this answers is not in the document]
dor-forer: Confirmed, and this one changed the design rather than getting patched. Thank you — the numbers are worse than the description suggests.

At `UINT8_SIMD_MAX_DIM = 4 * 66,051 = 264,204`, the perfectly-even case already lands at `65,025 * 66,051 = 4,294,966,275` against `UINT32_MAX = 4,294,967,295`. That is **1,020** of headroom for any unevenness at all. The masked residual load can add up to 16 dot-products into specific lanes, worth `16 * 65,025 = 1,040,400`, exceeding that headroom by a factor of a thousand. So lanes wrap before `vaddlvq_u32` ever sees them, and the band test I added asserted exactness at exactly that dimension.

My derivation was "each lane receives dim/4 products", which holds for the main loop and is false once the residual path is counted.

A corrected bound is derivable — roughly 264,140 for this kernel — but it would have to be derived **per kernel** from its accumulator count and residual distribution, and it is a fragile invariant: any future change to a residual path silently breaks it. This is also the third time per-ISA lane reasoning has produced an error in this PR, after the AVX-512-versus-NEON asymmetry and the `vaddq_u32`-before-widening point.

So the widened band is removed rather than re-bounded. The kernels use the unsigned 32-bit reduce, exact through 66,051, and the choosers hand back the scalar kernel above it. That bound is on the horizontal **total**, which does not depend on how products land in lanes, so no per-kernel analysis is needed to justify it.

Worth recording, since it is the thing that made the wider band look attractive: on ARM the widening reduce is free instruction-for-instruction. `clang++ --target=aarch64-linux-gnu -O2` emits `addv`/`fmov w`/`ucvtf` against `uaddlv`/`fmov x`/`ucvtf`, three instructions either way. The obstacle is the lane bound you identified, not the reduce.

The rejected alternatives and their measurements are recorded in the commit message so this is not re-litigated: unconditional widening cost +20% at dimension 32, and a runtime branch cost +0.4 to 0.6 ns per call while losing 31 Cosine wrappers their inlining.

tests/unit/test_spaces.cpp
```diff
@@ -2186,6 +2186,125 @@ TEST_P(UINT8SpacesOptimizationTest, UINT8_full_range_test) {
...
+    if (has_uint8_simd) {
+        EXPECT_NE(L2_UINT8_GetDistFunc(spaces::UINT8_SIMD_MAX_DIM, &alignment, nullptr),
+                  UINT8_L2Sqr);
+        EXPECT_NE(IP_UINT8_GetDistFunc(spaces::UINT8_SIMD_MAX_DIM, &alignment, nullptr),
+                  UINT8_InnerProduct);
+        // And the widened variant is a different function from the narrow one, so the narrow/wide
+        // selection is exercised rather than assumed.
+        EXPECT_NE(
+            IP_UINT8_GetDistFunc(spaces::UINT8_NARROW_REDUCE_MAX_DIM, &alignment, nullptr),
+            IP_UINT8_GetDistFunc(spaces::UINT8_NARROW_REDUCE_MAX_DIM + 1, &alignment, nullptr));
```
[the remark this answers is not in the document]
dor-forer: Correct, the assertion was vacuous. `66,051 % 64 = 3` and `66,052 % 64 = 4`, so those are different `CHOOSE_IMPLEMENTATION` instantiations regardless of which family they come from, and the check would have passed with `_Wide` never selected.

It is moot now: the widened variant is gone, for the lane-bound reason in the thread above, so there is no narrow-versus-wide selection left to assert. The test reverts to what it was before — exactness on the SIMD path at dimensions 33,026 and 40,000, both of which the old signed reduce got wrong, and pointer identity at 66,052 to prove the scalar fallback is selected.

Had the variant stayed, the fix would have been to compare two dimensions sharing a residual, for example `66,051` against `66,051 + 64`, so the only difference between the two instantiations is the family.

tests/unit/test_spaces.cpp
```diff
@@ -2186,6 +2192,392 @@ TEST_P(UINT8SpacesOptimizationTest, UINT8_full_range_test) {
...
+    // exact multiple of it, and dimensions spanning two and three chunks.
+    for (const size_t dim : {65535UL, 65536UL, 65537UL, 65600UL, 131072UL, 131109UL, 200000UL}) {
+        // The cosine kernels read a float norm from just past the payload, so size for both.
+        std::vector<uint8_t> ones(dim + sizeof(float), 255);
+        std::vector<uint8_t> zeros(dim + sizeof(float), 0);
+        std::vector<uint8_t> ramp(dim + sizeof(float));
+        for (size_t i = 0; i < dim; i++) {
+            ramp[i] = static_cast<uint8_t>(i % 256);
+        }
+        const float norm = std::sqrt(255.0f * 255.0f * static_cast<float>(dim));
```
lerman25: I think we are missing the production preprocessing path here. These tests append the cosine norm manually, but normal UINT8 cosine usage computes it through:

`VecSimIndex_AddVector -> preprocess/preprocessForStorage -> CosinePreprocessor<uint8_t> -> integer_normalizeVector<uint8_t> -> IntegralType_ComputeNorm<uint8_t>`

Queries reach the same code through `preprocessQuery`, and `VecSim_Normalize(..., VecSimType_UINT8)` calls it directly.

`IntegralType_ComputeNorm` still accumulates into a signed `int`. With an all-255 vector, every element adds `65,025`. Dimension 33,025 produces `2,147,450,625`, which fits, but dimension 33,026 produces `2,147,515,650`, which overflows `int`. On the current head, dimension 65,537 gives a NaN norm, while dimension 66,052 gives `252.99` instead of roughly `65,536.49`.

This means the chunked cosine kernel can still receive a bad norm before it runs.

Could we widen the norm accumulator to `uint64_t` and add a test that goes through preprocessing? For example, call `VecSim_Normalize` on an all-255 UINT8 vector at dimension 65,537, verify that the appended norm is finite and correct, and then add/query the same vector through a BF cosine index and check that its self-distance is near zero.

src/VecSim/spaces/IP/IP_AVX512F_BW_VL_VNNI_UINT8.h
```diff
@@ -87,19 +103,60 @@ static inline int UINT8_InnerProductImp(const void *pVect1v, const void *pVect2v
         } while (pVect1 < pEnd1);
     }
 
-    return _mm512_reduce_add_epi32(sum);
+    // Unsigned, and exact for up to spaces::UINT8_CHUNK_ELEMENTS elements.
+    return static_cast<uint32_t>(_mm512_reduce_add_epi32(sum));
```
lerman25: I don't think this cast fully makes the AVX-512 reduction unsigned:

```cpp
static_cast<uint32_t>(_mm512_reduce_add_epi32(sum))
```

GCC 10 and 13 implement `_mm512_reduce_add_epi32` using signed additions, including a final scalar `int + int`. For all-255 input, that addition overflows from dimension 33,026 onward. The cast only happens after the overflowing addition, so the behavior is still undefined even though the final total fits in `uint32_t` up to the chosen chunk size.

I confirmed that GCC 13 emits a signed-overflow UBSan check for the final reduction. A normal optimized build happens to wrap to the expected bits, which is why the current result tests pass.

Could we fold the lanes using unsigned or widened scalar additions instead? The same issue exists in the L2 reduction. It would also be useful to run the AVX-512 cases at dimensions 33,025, 33,026, and 65,536 with signed-overflow UBSan, since a normal equality test will not reliably catch this.

lerman25 (commented):
Two correctness concerns remain in the current implementation.

tests/unit/test_spaces.cpp
dor-forer: You were right, and the coverage gap was the sharper half: the kernel tests appended the norm by hand, so `IntegralType_ComputeNorm` was never exercised at all.

Fixed by widening its accumulator to `uint64_t`. Your figures reproduce exactly: 33,025 gives 2,147,450,625 which fits, 33,026 gives 2,147,515,650 which does not, and on the old code dimension 65,537 returned NaN while 66,052 returned 252.99 against an expected 65,536.49. int8 has the same shape, though its largest squared term is `(-128)^2 = 16,384` rather than `127^2`, so its threshold was 131,072.

Two tests now go through the production path as you suggested, in `test_common.cpp`: `NormalizeUint8LargeDimension` calls `VecSim_Normalize` on an all-255 uint8 vector at 33,025 / 33,026 / 65,537 / 66,052 and checks the appended norm is finite and correct, and `Uint8CosineSelfDistanceAtLargeDimension` stores that vector in a uint8 cosine brute force index at 65,537 and queries it with itself, asserting the self-distance is about zero. Both pass on Sapphire Rapids and on Graviton4.

src/VecSim/spaces/IP/IP_AVX512F_BW_VL_VNNI_UINT8.h
dor-forer: Confirmed and fixed. You were right that the cast cannot help, because it happens after the overflowing addition. I checked GCC 13's header rather than take it on trust:

```c
#define __MM512_REDUCE_OP(op) \
  __v8si __T1 = ...;                    /* signed int vectors */
  ...
  return __T8[0] op __T8[1]              /* scalar int + int */
```

The kernels now zero-extend the 16 lanes to 64 bits and reduce there, in both the IP and L2 AVX-512 kernels:

```cpp
const __m512i widened = _mm512_add_epi64(_mm512_unpacklo_epi32(sum, zero),
                                         _mm512_unpackhi_epi32(sum, zero));
return static_cast<uint32_t>(_mm512_reduce_add_epi64(widened));
```

`_mm512_reduce_add_epi64` folds through a scalar `long long` add, but the total is at most about 4.29e9 against `INT64_MAX`, so it cannot overflow. Cost is 2 instructions per call with the SIMD loop unchanged.

One note on why this stays in the kernel rather than being handled by the new dispatcher bound. The design changed since your review: chunking is gone, replaced by a bound of `UINT32_MAX / (255 * 255) = 66,051` with a scalar fallback above it. That keeps a 33,026-dimension-wide band on SIMD where the total exceeds `INT32_MAX`, so the fold is still needed. And the `SQ8_SQ8` kernels call `UINT8_InnerProductImp` directly without passing through a uint8 chooser, so a dispatcher-level guard would never protect them.

On your suggestion to run it under signed-overflow UBSan at 33,025 / 33,026 / 65,536: not done, and I want to be straight about that. This dev box has no AVX-512, and the AVX-512 host I have available ran the correctness tests (472/472, with both tier-aware tests reporting `AVX512F_BW_VL_VNNI` at every boundary dimension) but I did not build it with `-fsanitize=signed-integer-overflow` on both the parent and the fix, which is what would demonstrate the UB rather than argue it from the header. If you want that before approving, say so and I will run it.

The dimensions you named are now covered by tests either way: 33,025 and 33,026 both stay on SIMD and are asserted against an oracle computed independently in 64-bit integers, at worst-case all-255 inputs, on every available tier.

dor-forer:
The bound moved since your review, in the opposite direction, so flagging it rather than letting it turn up in a re-read.

It was `UINT32_MAX / 65025 = 66,051`, which is what forced the widened AVX-512 fold: GCC implements `_mm512_reduce_add_epi32` as signed vector ops ending in a scalar `int + int`, so staying exact through the full unsigned range meant a zero-extend-then-reduce-in-64-bit epilogue in every AVX-512 uint8 kernel. That epilogue is paid on every call at every dimension, and it buys only the band 33,026 to 66,051, which no workload reaches given embeddings run 384 to 4096.

It is now `INT32_MAX / 65025 = 33,025`, and both AVX-512 epilogues are back to the single intrinsic, identical to `main`.

Two consequences worth your attention:

**The SQ8 dispatchers are now guarded too, and that is not optional.** `SQ8_SQ8` kernels call the shared `UINT8_InnerProductImp` directly and never pass a uint8 chooser, so once the fold came out they were the one remaining path able to wrap a signed reduce. `IP_SQ8_SQ8`, `Cosine_SQ8_SQ8` and `L2_SQ8_SQ8` take the same guard, so six dispatchers rather than three. Their scalar fallback accumulates in `float`, imprecise past 2^24 but well defined, which is the better of the two failure modes.

**ARM gives up range it did not have to.** `vaddvq_u32` and `svaddv_u32` are genuinely unsigned, so NEON and SVE are exact to 66,051 with no change at all. The cap applies to them only because the bound is uniform across architectures. The alternative is a per-architecture bound, which is correct but makes "is this dimension on SIMD" host-dependent, so two machines would disagree about which kernel serves dimension 50,000. I picked uniform for that reason; say so if you would rather ARM keep its full range.

**On the tests.** The two boundaries collapse into one, since 33,025/33,026 is now both the signed-reduction boundary and the dispatcher boundary. The oracle sweep runs 32,960 to 33,025 contiguously, covering every residual and ending exactly at the cap, and now asserts the worst case at the cap is both within `INT32_MAX` and within one element's product of it.

That second assert immediately caught a bug in the test itself, which is worth knowing about because it had been latent through your review: each dimension's cosine norm is written at `vec + dim`, which is payload for every larger dimension. With a contiguous sweep, dimension 32,960's norm had overwritten 65 bytes of dimension 33,025's all-255 vector and the total came up 2,753,904 short of `65025 * 33025`. The previous two-base sweep could not detect this because both sides of the comparison read the same corrupted array; only asserting against an absolute value exposed it. Both tier-aware tests now save and restore those bytes.

**Still outstanding on my side, not claimed as done:**

- The UBSan run you asked for. Not done.
- ARM re-validation at this head is running now. The earlier Graviton4 result (1428/1428, all four tiers) was measured against the 66,051 boundary and both tier tests now use different dimensions, so I have removed those numbers from the description rather than carry them over.
- The widened-fold-versus-plain-reduce A/B did not finish, so the description cites the earlier revision's +4 to 11% as indicative and explicitly not re-measured against this head.

x86 at this head: 1036/1036, with `AVX512F_BW_VL_VNNI` reported at 32,960 / 32,993 / 33,023 / 33,024 / 33,025, so the cap itself is exercised on SIMD rather than inferred.

tests/unit/test_common.cpp
```diff
@@ -755,6 +755,59 @@ TEST(CommonAPITest, NormalizeUint8) {
...
 }
 
+// The norm the cosine preprocessor writes goes through IntegralType_ComputeNorm, which accumulated
+// into a signed int. Each uint8 element contributes up to 255*255 = 65,025, so the total passes
+// INT32_MAX from dimension 33,026: at 65,537 the norm came back NaN, and at 66,052 it came back
+// 252.99 instead of about 65,536.49. Every stored vector and every query for a uint8 cosine index
+// takes this path, so a wrong norm reaches the distance kernels before they run, and no amount of
+// exactness inside the kernels can repair it.
+//
+// Raised in review of MOD-17527 by @lerman25. It went unnoticed because the kernel tests append the
```
lerman25: remove me don't want to be part of this

lerman25 (approved):
LGTM

PR #1018, opened 2026-08-20
[MOD-17844] Stop ARM SIMD tiers executing instructions the running CPU lacks

dor-forer:

Two defects that let an ARM SIMD tier execute instructions the running CPU does not have. Both are correctness fixes, not performance work: see the benchmark section below, which finds no throughput change outside the paths that were already correct.

1. `functions/NEON_HP.cpp` was one translation unit compiled with `-march=armv8.2-a+fp16fml`, holding both the HP-only and the FHM entry points. The whole TU therefore carried the fp16fml licence, and the compiler contracted the HP path's `vcvt_f32_f16` plus `vfmaq_f32` into `fmlal`, even though the HP source uses no FMLAL intrinsic. That path is dispatched on `features.asimdhp` alone, so on a core with `asimdhp` but without `asimdfhm` it is a SIGILL, reachable from any SQ8_FP16 IP, L2 or Cosine query. Measured on Neoverse with gcc 12: 32 AdvSIMD FMLAL instructions in the HP-only wrappers at `+fp16fml`, 0 at `+fp16`, 96 across the whole TU.

   Fixed by splitting into two tiers: `NEON_HP` at `+fp16` predicated on `asimdhp`, and a new `NEON_FHM` at `+fp16fml` predicated on `asimdhp && asimdfhm`. Tightening the predicate to require both bits instead would have deleted the HP-only fallback for exactly the CPUs that need it. The FHM tier still emits 64 FMLAL instructions, so its fast path is unchanged.

2. `SVE2.cpp` recompiles fourteen of `SVE.cpp`'s kernel headers at `-march=armv9-a+sve2` rather than `-march=armv8-a+sve`. The kernels are templates at namespace scope with no `static`, so each instantiation is a weak symbol that both objects define, holding bodies built for different architectures, and the linker keeps whichever it saw first. Nothing in the source decides which.

   Fixed by giving those definitions internal linkage: each of the fourteen headers `SVE2.cpp` shares with `SVE.cpp` opens an anonymous namespace after its own includes and closes it at end of file. The tiers may then use identical names while producing independent bodies under their respective flags, and only the `Choose_*` entry points stay externally visible. Putting the namespace inside the header, rather than around the include site, is what lets the header keep including its own dependencies, which must stay outside it. `SVE2.cpp` itself is unchanged.

   The `NEON_HP` / `NEON_FHM` pair is deliberately left alone. It shares the two SQ8_FP16 headers at different `-march`, so it is the same structural hazard, but the kernels inside already carry `_NEON_HP` and `_NEON_FHM` in their names and the pair shares 0 symbols in Release and Debug alike. Wrapping those headers would have fixed nothing measured, so that pair relies on the naming convention, with `tier_linkage` guarding it.

Defect 1 is reproduced, not just inferred. On an AWS Graviton2 (Neoverse-N1, `asimdhp` present and `asimdfhm` absent, gcc 12), a program that calls `L2_SQ8_FP16_GetDistFunc` / `IP_...` / `Cosine_...` with `arch_opt = nullptr`, so the CPU's own feature bits choose the tier, dies with `Illegal instruction (core dumped)`, exit 132, on the first call. On this branch the same binary returns from all three. The tier objects on that same box: `NEON_HP.cpp.o` carries 96 `fmlal` on main and 0 here, with the 64 in the new `NEON_FHM.cpp.o` reachable only when `asimdfhm` is set.

**Scope of defect 2, measured**

`SVE.cpp.o` and `SVE2.cpp.o` share 152 externally defined symbols on this PR's base commit. Every one of the other 27 ARM tier pairs shares 0, and so do all 105 x86 tier pairs, because those kernel names embed the ISA.

`SVE.cpp.o` precedes `SVE2.cpp.o` in the archive and first-in-link-order wins, so the SVE bodies survive and every `Choose_*_SVE2` except the three SQ8_FP16 ones has been dispatching to `armv8-a+sve` code: the SVE2 tier is selected on SVE2 hardware and runs base-SVE codegen. Confirmed by linking a probe, where the surviving `FP16_L2Sqr_SVE<false,0>` is `SVE.cpp.o`'s 51 instructions rather than `SVE2.cpp.o`'s 49.

It is a silent downgrade rather than a fault only because none of the 152 bodies currently holds an SVE2-only opcode: all 152 differ between the two objects, but diffing the instruction mnemonics of every shared body gives the empty set in both directions, so the difference is scheduling and register allocation. It becomes a SIGILL on an SVE-without-SVE2 core the first time the compiler emits one, and because link order alone picks the winner it can differ between builds of the same commit.

**Why the namespace rather than renaming each kernel**

An earlier revision of this PR renamed the kernels individually with `#define` aliases in `SVE2.cpp`. That approach is incomplete by construction, and was: it missed `SQ8_SQ8_InnerProductSIMD_SVE_IMP`, because the list was derived from the symbols `nm` reported in a release build and `-O3` inlines that helper away entirely, 0 symbols in both objects against 8 each for its SQ8_FP32 sibling. It also could not cover the eight step helpers declared plain `inline` rather than `static inline`, which collided at `-O0` and had no name to alias. The anonymous namespace covers the kernels, the `_IMP` helpers and the `inline` helpers alike, so a function added to one of these headers is safe by default.

**Regression test**

`tests/unit/check_tier_linkage.py`, run from ctest as `tier_linkage`, `nm`s `libVectorSimilaritySpaces.a` and asserts that no two tier objects define a symbol in common. It is architecture-agnostic and covers x86 as well, 15 tier objects over 105 pairs. It excludes `vecsim_types` helpers, which come from a shared type header rather than a kernel header and are scalar bit manipulation that every tier compiles to the same bytes, verified byte-identical; `float16::cvt` is a member function and cannot take internal linkage regardless.

The check is run on Debug as well as Release, which matters: at `-O0` nothing is inlined away, so it is the configuration where a hidden collision shows. The alias revision above could never pass at `-O0`; this one does.

**Benchmark: the shared SVE2 kernels do not earn their recompilation**

`bm-spaces` on `r8g.xlarge` (Neoverse-V2, real SVE2), 558 SVE vs SVE2 pairs. These numbers are only meaningful on this branch, because on main the SVE2 choosers resolve to `SVE.cpp.o`'s bodies and the comparison would measure SVE against itself.

| kernels | median SVE2/SVE | faster by >3% | slower by >3% |
| --- | --- | --- | --- |
| the three SQ8_FP16 paths, genuinely SVE2 source | 0.545, i.e. 1.83x faster | 105 of 108 | 0 |
| the eighteen shared kernels, SVE source at `+sve2` | 0.995 | 81 of 450 | 45 of 450 |

Per-type medians for the shared kernels: FP16 1.000, FP32 0.998, FP64 0.994, SQ8_FP32 0.982, SQ8_SQ8 1.000. That is noise, and it matches the codegen: identical instruction vocabulary, plus or minus 16 instructions, no SVE2-only opcodes.

The 1.83x on SQ8_FP16 is not gained by this PR. Those kernels live in `IP/L2_SVE2_SQ8_FP16.h`, are already named `_SVE2`, never collided, and already work on main.

Follow-up, not done here: the eighteen shared choosers could delegate to `Choose_*_SVE` instead of recompiling, which would delete a tier's worth of duplicate object code that buys nothing measurable. That touches roughly 250 reference sites across 13 files, mostly `test_spaces.cpp` and the benchmark registrations, so it belongs in its own change.

**Validation**

Neoverse-N1 (gcc 12), this commit: Release and Debug both build clean under `-Werror -Wall` with zero warnings, `tier_linkage` passes on both over 8 tier objects and 28 pairs, and the spaces suite is 1529/1529 on each. The Debug run is the meaningful one for defect 2, since `-O0` inlines nothing away. That box has `asimdhp` without `asimdfhm`, which is what makes it the right hardware for the defect 1 repro, but it has no SVE at all, so it does not execute the SVE or SVE2 kernels.

Real SVE2 coverage came from a `workflow_dispatch` of `arm.yml`, which provisions `r8g.xlarge`: 2704/2704, with `sve sve2 svebf16 i8mm` in the runner's `lscpu`, so the `if (optimization.sve2)` assertion paths executed rather than being skipped for a missing feature bit. That run covered an earlier revision of this branch, which reached the same end state by renaming the kernels instead of scoping them; the kernel bodies it exercised are identical to the ones here, but it has not been repeated against this commit. CI's own ARM job runs at merge-queue time.

The `tier_linkage` check also passes against an x86_64 Release archive, 15 tier objects over 105 pairs.

1. MOD-17844

1. `src/VecSim/spaces/functions/NEON_HP.{cpp,h}` and new `NEON_FHM.{cpp,h}`
2. `src/VecSim/spaces/{IP,L2}_space.cpp` dispatch sites for SQ8_FP16
3. `cmake/aarch64InstructionFlags.cmake` and `src/VecSim/spaces/CMakeLists.txt` for the new tier
4. The fourteen kernel headers under `src/VecSim/spaces/{IP,L2}/` that `SVE.cpp` and `SVE2.cpp` share
5. `tests/unit/check_tier_linkage.py`, `tests/unit/CMakeLists.txt`, `tests/unit/test_spaces.cpp` and `tests/benchmark/spaces_benchmarks/`

**Mark if applicable**

---

> [!NOTE]
> **High Risk**
> Changes ARM SIMD compilation and runtime dispatch for vector distance kernels. A mistake here can SIGILL or silently run the wrong ISA on the query path.
> 
> **Overview**
> Stops ARM distance kernels from running instructions the CPU does not have.
> 
> **NEON HP vs FHM:** `NEON_HP.cpp` was compiled with `+fp16fml` even for HP-only SQ8↔FP16 paths, so the compiler could emit `fmlal` on cores with `asimdhp` but not `asimdfhm` (SIGILL). HP now builds with `+fp16` and a new `NEON_FHM` TU builds with `+fp16fml`. Dispatch requires `asimdhp && asimdfhm` for FHM, then falls back to HP on `asimdhp` alone.
> 
> **SVE vs SVE2:** Shared kernel headers were compiled into both TUs as weak namespace-scope templates, so link order picked one `-march` for both choosers. Those headers now wrap kernels in an anonymous namespace so each tier keeps its own bodies; only `Choose_*` stays external.
> 
> Adds `tier_linkage` (`check_tier_linkage.py`) to fail if any two SIMD tier objects export the same symbol.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2fc6c1caab4f7119be0a0ad9d84ecbb16689da42. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/VecSim/spaces/functions/NEON.cpp
```diff
@@ -7,6 +7,21 @@
...
+#include "VecSim/spaces/space_includes.h"
+#include "VecSim/spaces/spaces.h"
+#include "VecSim/types/bfloat16.h"
+#include "VecSim/types/float16.h"
+#include "VecSim/types/sq8.h"
+#include <arm_neon.h>
+
+// Kernel instantiations get internal linkage, unique to this translation unit, so two tiers
+// that share a kernel header cannot emit the same weak symbol and let link order pick the
+// body. Only this tier's Choose_* entry points stay external.
```
GuyAv46: Consider renaming the problematic functions instead of adding these includes everywhere

dor-forer: You were right, and more right than I realised. Done in 36042c98, which drops the linkage commit entirely and replaces it with renaming.

Measuring which tier pairs actually collide, on this PR's base commit:

| pair | shared external symbols |
| --- | --- |
| `SVE.o` / `SVE2.o` | 152 |
| the other 27 ARM pairs | 0 |
| all 105 x86 pairs | 0 |

So the change to `NEON.cpp` you were reading, and the ones to `NEON_BF16.cpp`, `NEON_DOTPROD.cpp`, `NEON_HP.cpp`, `NEON_FHM.cpp` and `SVE_BF16.cpp`, fixed nothing. Six of the eight TUs I touched had no collision to fix. `NEON.cpp` was a reasonable place to ask.

I should also correct the description I had written: it claimed `NEON.o` and `NEON_DOTPROD.o` shared 93 symbols. They share 0, and not because main moved under the branch. #1014 and #1015 are both ancestors of this PR's base `7a5fe7f9`, so the renames they did had already removed that collision before I opened this. The number was stale when I wrote it. Scenario 3 in the description depended on it and is gone too.

On renaming: the leaf kernels already carry their ISA (`INT8_InnerProductSIMD16_NEON` vs `_NEON_DOTPROD`), so there was nothing misnamed to fix. The one real collision is that `SVE2.cpp` includes fourteen of `SVE.cpp`'s headers on purpose and recompiles them at `armv9-a+sve2`, so it is one source text producing two different bodies under one name. The rename has to happen at the include site, which is 19 `#define`s at the top of `SVE2.cpp`:

```c
#define FP32_InnerProductSIMD_SVE FP32_InnerProductSIMD_SVE2
#define FP32_L2SqrSIMD_SVE        FP32_L2SqrSIMD_SVE2
...
```

Each one covers both the definition in the header and the use in the `Choose_*` body below, since those reference the same names. 1 file, +26/-15, instead of 8 files and +190.

Your suggestion also turned out to fix something the anonymous namespace did not make visible. `SVE.cpp.o` precedes `SVE2.cpp.o` in the archive and first-in-link-order wins, so on main today every `Choose_*_SVE2` except the three SQ8_FP16 ones is dispatching to `armv8-a+sve` bodies. The SVE2 tier is selected on SVE2 hardware and runs base-SVE codegen. The linked `FP16_L2Sqr_SVE<false,0>` is `SVE.cpp.o`'s 51 instructions, not `SVE2.cpp.o`'s 49. Internal linkage would have kept the right body but left the name lying about which `-march` built it; renaming makes it say so.

And the best argument for your approach was already in this PR. `NEON_HP.cpp` and `NEON_FHM.cpp` include the same two SQ8_FP16 headers at `+fp16` and `+fp16fml`, which is structurally the identical hazard, and they share 0 symbols with no linkage tricks at all, purely because the kernels inside are named `_NEON_HP` and `_NEON_FHM`. Defect 1's own fix already used naming. `SVE2.cpp` was the only file not following the convention.

The one thing the boilerplate did buy was covering headers nobody had thought about yet, so I replaced that with a check instead: `tests/unit/check_tier_linkage.py`, run from ctest as `tier_linkage`. It `nm`s the archive and fails if any two tier objects define a symbol in common. Architecture-agnostic, passes on x86 (15 objects, 105 pairs, 0 shared), and if I strip the renames back out it fails with all 152 symbols named. So the invariant is enforced rather than remembered, and it protects the NEON tiers without putting anything in them.

Neoverse, gcc 12: full build clean, 1529/1529.

src/VecSim/spaces/functions/SVE2.cpp
```diff
@@ -8,24 +8,57 @@
...
+#define UINT8_L2SqrSIMD_SVE        UINT8_L2SqrSIMD_SVE2
+
+#define SQ8_FP32_InnerProductSIMD_SVE_IMP SQ8_FP32_InnerProductSIMD_SVE2_IMP
+#define SQ8_FP32_InnerProductSIMD_SVE     SQ8_FP32_InnerProductSIMD_SVE2
+#define SQ8_FP32_CosineSIMD_SVE           SQ8_FP32_CosineSIMD_SVE2
+#define SQ8_FP32_L2SqrSIMD_SVE            SQ8_FP32_L2SqrSIMD_SVE2
+
+#define SQ8_SQ8_InnerProductSIMD_SVE SQ8_SQ8_InnerProductSIMD_SVE2
+#define SQ8_SQ8_CosineSIMD_SVE       SQ8_SQ8_CosineSIMD_SVE2
+#define SQ8_SQ8_L2SqrSIMD_SVE        SQ8_SQ8_L2SqrSIMD_SVE2
```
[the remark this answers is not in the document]
dor-forer: Correct, and it is a live collision rather than a theoretical one. Fixed in 60c9c605.

Confirming it from the source side, as you did: there are 20 non-static namespace-scope templates in the fourteen headers `SVE.cpp` and `SVE2.cpp` share, and the first pass covered 19. `SQ8_SQ8_InnerProductSIMD_SVE_IMP` was the one left out, and `IP_SVE_SQ8_SQ8.h` calls it by name from the wrapper beside it, so both tiers instantiate it.

Worth recording why it escaped, because it is a flaw in how I derived the list rather than an oversight. I took the names from the symbols `nm` reported as shared in a release build. At `-O3` that helper is inlined into its callers and emits no out-of-line symbol at all:

| helper | `SVE.o` | `SVE2.o` |
| --- | --- | --- |
| `SQ8_SQ8_InnerProductSIMD_SVE_IMP` | 0 | 0 |
| `SQ8_FP32_InnerProductSIMD_SVE_IMP` | 8 | 8 |

So it was invisible to the binary and to the `tier_linkage` check I added, which reads the same symbol table. Reading the declarations, which is what you did, is the sound method here.

A debug build settles that it is real. At `-O0` nothing is inlined away and the helper appears 8 times in each object, and `check_tier_linkage.py` fails on that archive. After the fix, on Neoverse with gcc 12:

| | before | after |
| --- | --- | --- |
| release, 28 ARM tier pairs | 0 shared | 0 shared |
| debug, `SVE.o` / `SVE2.o` | 16 shared | 8 shared |

The 8 that remain are a separate pre-existing thing, not this helper: eight step helpers in the SVE kernel headers are declared plain `inline` rather than the `static inline` the NEON headers use, so they get vague linkage too. Their bodies are byte-identical between the two tiers, as are the three `vecsim_types` float16 helpers that `NEON_HP.o` and `NEON_FHM.o` share at `-O0`, so nothing miscompiles today. Left out of this change and recorded separately.

The wider point your finding exposes is that `tier_linkage` has a blind spot on release builds, since a kernel the optimiser inlines away cannot appear in the symbol table. Running it against a debug archive closes it, and that needs the two cleanups above first.

For completeness, I did try a source-side check to catch this class automatically and reverted it: it reported 11 collisions and all 11 were false positives, because a template in a shared header only collides if both TUs actually instantiate it. Seven were `NEON_HP`/`NEON_FHM`, which share `IP_NEON_SQ8_FP16.h` but each instantiate only their own kernels, and four were `SSE`/`SSE4`, where `SSE.cpp` includes `IP/L2_SSE4_SQ8_FP32.h` without ever instantiating them. Those four are two dead includes in `SSE.cpp`, which I have noted but not touched here.

tests/unit/CMakeLists.txt
```diff
@@ -84,3 +81,14 @@ gtest_discover_tests(test_uint8 TEST_PREFIX UINT8UNIT_)
...
 # SVS Tiered tests are slow in debug mode, so we set a longer timeout
 gtest_discover_tests(test_svs PROPERTIES TIMEOUT 3000)
+
+# Guard the per-tier SIMD objects against sharing externally visible symbols, which would let
+# link order decide which -march a kernel body was built for. See check_tier_linkage.py.
+find_program(PYTHON3_FOR_TESTS NAMES python3 python)
+if(PYTHON3_FOR_TESTS)
+	add_test(NAME tier_linkage
+		COMMAND ${PYTHON3_FOR_TESTS} ${CMAKE_CURRENT_SOURCE_DIR}/check_tier_linkage.py
+			$<TARGET_FILE:VectorSimilaritySpaces>)
```
GuyAv46, line 91: Awesome!!!

GuyAv46 (approved):

PR #1022, opened 2026-08-23
[MOD-17795] Compare merge scores exactly, so a mid-ingest label is not returned twice

dor-forer:

An `FT.SEARCH ... =>[KNN ...]` reply can carry the same document key twice, and on a full reply each duplicate costs a real document (reported: 2000 keys, 1999 distinct, over a 2000-document corpus; worst seen 18 repeats in one reply). The duplicate is produced inside VecSim, in the tiered frontend/backend result merge.

`VecSimTieredIndex::topKQueryImp` queries the flat buffer and backend separately and merges the two lists. The old single-value path passed `withSet=false`, so it relied on a shared id reaching both cursors at the same time. This PR now gives every public one-shot tiered top-K and range query explicit cross-tier ID deduplication whenever both tiers can contribute more than one result.

It did not agree. Both lists are ordered by *exact* score — `HNSWIndex::topKQuery` and `BruteForceIndex::topKQuery` drain a max-heap keyed on `(distance, label)`, and the batch iterators sort by score then id — while the comparator called any two scores within `1e-6` equal and fell back to ordering by id. Wherever those two orders disagree, the flat buffer's copy of a label and the backend's copy miss each other and both land in the reply.

A label is in both indexes for as long as an ingest job takes to get from inserting into HNSW to erasing it from the flat buffer, so a corpus dense enough to put unequal distances within `1e-6` of each other duplicates documents under any sustained write load. That matches the report: it reproduces during the background scan and also after it (`percent_indexed=1.0`, ingest queue not yet drained), on both `FLOAT32` and `FLOAT16`, with a bit-identical `__dist` on both rows and neighbouring distances ~1.2e-7 apart — inside the tolerance.

This drops the tolerance. The rewritten comparator also fixes the id comparison, which computed `(int)(res1->id - res2->id)` and so truncated a 64-bit label difference to `int`, giving the wrong sign for ids more than 2^31 apart.

The exact comparator fixes inconsistent merge ordering everywhere it is used. Public one-shot `topKQuery` and `rangeQuery` additionally use a smaller-side ID state map, so a shared ID is returned once even when its two scores differ. The batch iterator still uses its existing merge/persistent-state logic and remains a separate follow-up for arbitrary cross-tier score disparity.

1. MOD-17795

1. `cmpVecSimQueryResultByScoreThenId()` — exact score comparison and sign-correct ID comparison.
2. `merge_results_with_cross_tier_dedup()` — an allocator-aware `unordered_map` containing only the smaller input's IDs and their emitted state.
3. `merge_result_lists()` — skips deduplication state when the limit is at most one or one input is empty, because those shapes cannot emit a cross-tier duplicate.
4. Public tiered `topKQuery()` / `rangeQuery()` — perform cross-tier ID deduplication for non-trivial one-shot queries.
5. Regression tests — cover near-tied exact ordering, unequal duplicate scores in both directions, exact duplicates, large IDs, top-K filling, trivial-merge fast paths, and public top-K/range dispatch (`BY_SCORE` and `BY_ID`).

**Testing**

The new test builds the minimal shape: two labels whose L2 distances differ by less than `1e-6` but are not equal, with the farther one also sitting in the flat buffer, as an in-flight ingest job leaves it. It asserts the distance gap as a precondition so it cannot silently stop exercising the case.

Verified on `dorer-intel` (Ubuntu 24.04, gcc 13.3):
- Before the fix, the test fails with the reported signature — `id: 1` twice at a bit-identical score:
  ```
  Value of: allUniqueResults(res)
    Actual: false
  id: 1, score: 1.0000004768371582
  id: 2, score: 1
  id: 1, score: 1.0000004768371582
  ```
- After the final trivial-merge fast path: `100% tests passed, 0 tests failed out of 2752` (8 expected skips).

**Why exact comparison and explicit deduplication are both needed**

The comparator must match the lists' exact `(score, id)` ordering; a tolerance makes the merge order inconsistent and is not transitive. Exact comparison fixes MOD-17795's near-tied-neighbour case, where the duplicated label itself has a bit-identical score.

Exact comparison alone cannot deduplicate a shared ID whose frontend/backend scores differ. Public one-shot top-K and range queries therefore perform ID deduplication whenever a reply could contain the duplicate twice. `merge_result_lists` preloads only the smaller input's IDs into a reserved allocator-aware `unordered_map` and marks the first occurrence emitted. IDs absent from the smaller input cannot be cross-tier duplicates and pass without insertion. If `limit <= 1` or either input is empty, `merge_result_lists` uses the compile-time `merge_results<false>` path because that reply cannot emit a cross-tier duplicate.

Seven-repetition `dorer-intel` measurements compare this standard map with the removed custom flat hash in the same binary: the map was +0.8%, +3.1%, and +6.2% for top-K 10/100/1000, and +0.6% for a 1,026-result range query. The isolated merge is 2–6× slower, but vector search dominates end-to-end latency. The custom table can return in a separate optimization PR if those query-level costs prove material.

The trivial-merge dispatch was then measured in a 15-repetition, randomized-interleaved run on `dorer-intel` using the same release binary for both paths. In the isolated merge, K=1 improved from 128 ns to 57 ns (~55%); merging 1,000 results with an empty side improved from 3,102 ns to 2,654 ns when the first input was empty and from 3,042 ns to 1,394 ns when the second was empty. End-to-end top-K K=1 changed from 278.695 us to 278.120 us (-0.21%). K=10 through K=1,000 stayed within -0.21% to +0.22%, showing no meaningful cost from the runtime condition on ordinary queries.

The remaining `merge_results<false>` calls are the trivial one-shot cases above, internal same-iterator merges where duplicate IDs are impossible by construction, plus the existing single-value HNSW batch path. Exact comparison also fixes the single-value batch case when bit-identical cross-tier copies meet in the same merge: they now align and are consumed together. It does not make batch iteration robust to arbitrary score disparity or copies arriving in different batches; that requires persistent cross-batch state and remains a separate follow-up.

**This reverts #694, and why the implementation changed**

Before #694 the comparator used exact score comparison. #694 added the tolerance so a shared label with slightly different frontend/backend scores could align and deduplicate without an ID set. That solved one case but made the comparator disagree with the exact ordering of the input lists, allowing different near-tied labels to move the cursors out of alignment.

This PR restores exact ordering and gives deduplication a separate owner. The smaller-side state map handles unequal scores directly for public one-shot queries, rather than encoding approximate identity into the ordering comparator.

**Known debt left behind: MOD-17920**

`spaces::normalizeVector_imp()` divides by the vector norm with no zero-norm guard, so a zero vector in a cosine index normalizes to all-NaN. The integer merge comparator does not treat such a score as equal: `NaN != score` is true and `NaN > score` is false, so it returns `-1` regardless of which operand is NaN. That makes the merge order inconsistent and can select or truncate the wrong results. Separately, the score comparators passed to `std::sort` do not provide a strict weak ordering in the presence of NaN. This predates the PR and is untouched by it; filed as MOD-17920 with a proposed fix.

The explicit `!=` form avoids treating `double`'s `std::partial_ordering::unordered` result as equality and then advancing both merge cursors. It does not make NaN scores supported; MOD-17920 must either prevent them or define a total ordering for them. The comparator can be simplified after NaN scores are made unreachable.

**Mark if applicable**

---

> [!NOTE]
> **Medium Risk**
> Changes core tiered query merge behavior for all backends (HNSW/SVS); correctness fix with small added merge cost on multi-result queries, but batch-iterator cross-tier score disparity remains a known follow-up.
> 
> **Overview**
> Fixes **duplicate document keys** in tiered KNN/range replies when the same label temporarily exists in both the flat buffer and backend during ingest.
> 
> **Merge comparator** — Removes the `1e-6` score tolerance and fixes ID tie-breaking (no more `(int)(id1 - id2)` truncation). Ordering now matches the exact `(score, id)` sort used by flat/HNSW result lists, so merge cursors stay aligned for near-tied distances.
> 
> **One-shot tiered queries** — Drops the `withSet` template split (single-value “assume identical scores” fast path). `merge_result_lists` always deduplicates shared labels across tiers via `merge_results_with_cross_tier_dedup` (tracks IDs from the smaller input), except when `limit <= 1` or one side is empty. Range `BY_ID` dedup keeps the **better** score when both tiers disagree. Tiered SVS no longer overrides top-K/range to force set-based merge.
> 
> **Tests** — Regression for mid-ingest near-tied L2 scores, unit tests for cross-tier merge/dedup (large IDs, k=1, empty side), cosine batch iterator mid-ingest duplicate, and public API coverage for tiered search with different scores per tier.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit baf1ebf7d9127073b8fc489954b34323ddea29f3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## RediSearch macro-benchmark validation

A benchmark-only RediSearch integration PR ([RediSearch#11170](https://github.com/RediSearch/RediSearch/pull/11170)) pins this PR final head (`8bd95c7e`) and compares it with RediSearch `master` through the standard macro-benchmark workflow.

The #1022-relevant x86_64 results show no confirmed performance regression:
- VecSim full-text filter: -1.4% (baseline standard deviation 3.3%) — no change.
- VecSim numeric filter: -2.6% (baseline standard deviation 7.0%) — no change.
- VecSim tag filter: -3.0% (baseline standard deviation 5.0%) — no change.
- Hybrid linear numeric-vector: -2.1% — no change.
- Hybrid RRF text-vector: -3.3% — below the 8% regression threshold.
- Hybrid range cases: -4.1% to -6.3%, with highly unstable baselines; neither server- nor client-side metrics confirmed a regression.

The suite also reported a -13% x86_64 result in an unrelated full-text union-iterator benchmark, while the same aarch64 test was +0.4%; this is not attributable to the VecSim-only gitlink change.

These macro benchmarks validate steady-state query performance. They do not isolate the cross-tier deduplication cost during active ingestion, when the same label may temporarily exist in both tiers. Detailed results are available in the [benchmark comparison](https://github.com/RediSearch/RediSearch/pull/11170#issuecomment-5423590141).


alonre24 (dismissed):

src/VecSim/utils/query_result_utils.h
```diff
@@ -11,15 +11,18 @@
...
                                              const VecSimQueryResultContainer::iterator res2) {
-    return !double_eq(res1->score, res2->score) ? (res1->score > res2->score ? 1 : -1)
-                                                : (int)(res1->id - res2->id);
+    if (res1->score != res2->score) {
+        return res1->score > res2->score ? 1 : -1;
+    }
+    if (res1->id == res2->id) {
+        return 0;
+    }
+    return res1->id > res2->id ? 1 : -1;
```
ofiryanai, line 25: This comparator is also used by the HNSW batch iterator. on single value cosine tiered index, if during ingestion a label is in both tiers. scores might still differ in ~1e-7 because `executeInsertJob` feeds HNSW
the frontend's already-normalized blob and `preprocess()` normalizes it again. Previously `double_eq` treated those two copies as equal and `merge_results<false>` (`hnsw_tiered.h:1102`) collapsed them; exact comparison doesn't, so both can be returned now. and since single-value skips `filter_irrelevant_results` and records only `from_flat` ids, it also recurs across batches. Simplest fix is to let single-value use the multi path in compute_current_batch (merge_results<true>, record batch_res ids, filter both lists). Or maybe this is wrong, which should be proved using a cosine batch-iterator test showing the scores do match

tests/unit/test_hnsw_tiered.cpp
```diff
@@ -918,6 +918,56 @@ TYPED_TEST(HNSWTieredIndexTestBasic, MergeMulti) {
     runTopKSearchTest(tiered_index, query, 5, [](size_t _, double __, size_t ___) {});
 }
 
+// A label that is mid-ingest lives in the HNSW index and in the flat buffer at once, so
+// a top-K query merges two lists that both carry it. Both lists are ordered by exact score, while
+// cmpVecSimQueryResultByScoreThenId() used to call scores within 1e-6 equal and fall back to
+// ordering by id. Where those two orders disagree the shared label does not reach the two merge
+// cursors at the same time, and a merge that dedups by cursor alignment alone emits it twice.
+TYPED_TEST(HNSWTieredIndexTestBasic, MergeSingleWithNearTiedScores) {
```
ofiryanai, line 926: Could we get a couple more cases here? SearchDifferentScores covers single-value
L2 nicely, but nothing exercises the new path for a multi-value tiered index (the
old withSet=true case), and nothing covers tiered SVS now that the topKQuery/
rangeQuery overrides are gone. A cosine variant would also be worth having, since
double normalization on ingest is the shape this actually shows up as in
production.

src/VecSim/utils/query_result_utils.h
```diff
@@ -89,17 +92,82 @@ std::pair<size_t, size_t> merge_results(VecSimQueryResultContainer &results,
...
+// iterators cannot use this optimization because they also need to remember labels returned by
+// earlier calls.
+inline std::pair<size_t, size_t>
+merge_results_with_cross_tier_dedup(VecSimQueryResultContainer &results,
+                                    VecSimQueryResultContainer &first,
+                                    VecSimQueryResultContainer &second, size_t limit) {
+    results.reserve(std::min(limit, first.size() + second.size()));
+    const auto &smaller = first.size() <= second.size() ? first : second;
+    // The mapped bool records whether this tracked ID has already been emitted.
+    vecsim_stl::unordered_map<labelType, bool> tracked_ids(0, smaller.getAllocator());
```
ofiryanai, line 106: Two small performance suggestions, not blocking:
1. this puts the query-local scratch memory on the vecsim allocator (the old set used the default one), which is a little slower because of atomic bump of VecSimAllocator::allocated. Might be worth not using that allocator here, though if it's the convention, feel free to keep it
2. the map stores the ids, but `smaller` already holds them contiguously, so the only genuinely new state is one "already emitted" flag per position. For a single-digit flat buffer that's a small stack array plus a linear scan — no allocation, no hashing — which should beat the map outright. Worth keeping the map above a threshold though (~100): rangeQueryImp passes limit = -1, so with a full 1024-entry buffer a scan per emitted result would be much worse

ofiryanai (changes requested):
left some comments. the perf are nits, I simply already got into the rabbit hole

src/VecSim/utils/query_result_utils.h
dor-forer: I traced the cosine ingestion path. Tiered HNSW constructs the backend with `is_normalized=true` in `TieredHNSWFactory::NewIndex`, because normalization is done by the frontend. `CreatePreprocessorsContainerParams` then maps backend cosine preprocessing to IP, so `executeInsertJob` passes the already-normalized blob to HNSW without normalizing it again.
Therefore, while a label temporarily exists in both tiers, both copies contain the same normalized bytes and should produce bit-identical scores. The exact cmp == 0 branch still collapses that case.
I agree that arbitrary different-score duplicates across batch-iterator batches remain a broader gap, but that is separate from the current cosine ingestion path. I’ll add a focused single-value cosine overlap test that verifies the scores match and the label is returned only once across batches.

tests/unit/test_hnsw_tiered.cpp
dor-forer: I think two existing tests already cover part of this:
- `HNSWTieredIndexTestBasic.MergeMulti` exercises a multi-value HNSW cross-tier merge with the same label carrying different vectors/scores.
- `SVSTieredIndexTestBasic.MergeMulti` does the same for tiered SVS. After removing the SVS query overrides, this test now goes through the inherited generic public query path.
SearchDifferentScores additionally covers the single-value public top-K and range paths.
The focused cosine overlap/batch case is still missing, so I agree that adding that regression test would be useful.

src/VecSim/utils/query_result_utils.h
dor-forer: Both ideas make sense as possible follow-up optimizations. I’d prefer to keep the VecSim allocator here so the temporary allocation follows the existing allocator and memory-accounting conventions.
A small stack-backed linear scan may help for very small inputs, but it introduces another path and becomes expensive for larger range results. The current end-to-end comparison against the custom tracker was approximately:
- top-K 10: +0.8%
- top-K 100: +3.1%
- top-K 1000: +6.2%
- large range: +0.6%
Given that overhead, I’d keep the simpler map implementation in this correctness PR. We can benchmark the small-buffer approach separately if we want to pursue it as an optimization.

ofiryanai (approved):

PR #1025, opened 2026-08-26
Let production code read a label's stored vectors

nonirosenfeldredis:

Makes `getDataByLabel` usable outside `BUILD_TESTS`, so production code can ask what a
label is holding rather than only how far another vector is from it.

The motivating caller is RediSearch (MOD-17688). When a document is updated it gets a new
doc-id, and today every vector it owns is deleted and re-added even when the vector itself
did not change. Comparing the value about to be written against the stored one lets the
existing entry be moved to the new doc-id instead. A distance cannot serve that comparison:
`VecSimIndex_GetDistanceFrom_Unsafe` returns 0 for distinct vectors under IP, and under
cosine it requires a pre-normalized query, leaving only a tolerance — which calls two
nearby-but-different vectors equal. Byte equality of the stored form is the sound test, and
that needs the accessor.

Two fixes come with it, both consequences of the accessor now having non-test callers:

- **Asking about an absent label was unsafe.** The single-value implementations used
  `labelToIdLookup.at()`, which throws; the multi-value ones dereferenced `find()` without
  comparing to `end()`, which is undefined behaviour. Both now leave the output empty, so
  the output size answers whether the label is held.
- **A tiered index reported nothing for a label still in the flat buffer.**
  `TieredHNSWIndex::getDataByLabel` consulted the backend alone, so everything written since
  the last ingest job looked absent — worst for recently written vectors, which are the ones
  most likely to be written again. The lookup now asks the frontend first, under the guards
  `relabelVector` takes and in the same order. Nothing about it is HNSW-specific, so it moved
  to `VecSimTieredIndex`, where the SVS tiered index gets it too.

The base declaration becomes a defaulted virtual rather than pure, so SVS — whose
implementation is a not-implemented stub — needs no production definition, and an empty
output means "cannot tell" exactly as an absent label does.

1. MOD-17688

1. `VecSimIndexAbstract::getDataByLabel` — out of `BUILD_TESTS`, defaulted rather than pure
2. `BruteForceIndex_Single/_Multi`, `HNSWIndex_Single/_Multi` — same, and absence-safe
3. `VecSimTieredIndex::getDataByLabel` — new, asks the tier holding the label
4. `TieredHNSWIndex::getDataByLabel` — removed, the base covers it

**Mark if applicable**

**Testing**

`ctest` on this branch: 2759/2759 passed, including the 97 tiered HNSW tests. No test called
the tiered wrapper's `getDataByLabel` before this change — the seven existing uses all call a
tier directly — so the changed tiered behaviour is not covered yet; the RediSearch side that
motivates it is what exercises it today.

---

> [!NOTE]
> **Medium Risk**
> Changes tiered read locking/semantics and a new production API used for document update paths; behavior is intentional but affects concurrency and equality decisions for quantized/SVS indexes.
> 
> **Overview**
> **Promotes `getDataByLabel` from test-only to a documented production API** on `VecSimIndexAbstract`, so callers (notably RediSearch MOD-17688) can compare stored vector bytes on update instead of relying on distance.
> 
> **Index implementations are hardened for real use:** missing labels leave the output empty instead of throwing or UB; **quantized** indexes (SQ8/HNSW) return nothing rather than memcpy-ing compressed layout as floats, keyed off `isQuantized` not blob size. HNSW reads take **`indexDataGuard`** under the same rules as ingest/delete.
> 
> **Tiered lookup is centralized in `VecSimTieredIndex`:** read the flat buffer first, then the backend when multi-value or the label was not in the buffer—fixing “absent” for vectors still buffered. **SVS** backends skip reads entirely (partial tier answers would lie) until MOD-17706; `SVSIndex::getDataByLabel` logs and returns empty instead of asserting. **`TieredHNSWIndex`** drops its duplicate wrapper.
> 
> Unit tests cover absent labels, SQ8 at dim 4 and 40, tiered HNSW spanning both tiers, and tiered SVS reporting nothing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aaeafd0cfbe54834bfdb9ad9df130f636b0f2beb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/VecSim/algorithms/brute_force/brute_force_single.h
```diff
@@ -43,19 +42,22 @@ class BruteForceIndex_Single : public BruteForceIndex<DataType, DistType> {
     // We call this when we KNOW that the label exists in the index.
     idType getIdOfLabel(labelType label) const { return labelToIdLookup.find(label)->second; }
 
-#ifdef BUILD_TESTS
     void getDataByLabel(labelType label,
                         std::vector<std::vector<DataType>> &vectors_output) const override {
-
-        auto id = labelToIdLookup.at(label);
+        auto it = labelToIdLookup.find(label);
+        if (it == labelToIdLookup.end()) {
```
nonirosenfeldredis, line 61: fixed existing bug

src/VecSim/vec_sim_tiered_index.h
```diff
@@ -115,6 +115,29 @@ class VecSimTieredIndex : public VecSimIndexInterface {
...
+        if (this->frontendIndex->isLabelExists(label)) {
+            this->frontendIndex->getDataByLabel(label, vectors_output);
+            this->flatIndexGuard.unlock_shared();
+            return;
+        }
+        this->flatIndexGuard.unlock_shared();
+
+        this->mainIndexGuard.lock_shared();
+        this->backendIndex->getDataByLabel(label, vectors_output);
+        this->mainIndexGuard.unlock_shared();
```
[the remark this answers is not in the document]
nonirosenfeldredis: Confirmed and fixed in e41b64d8 — the guard now sits inside `HNSWIndex_Single/_Multi::getDataByLabel`.

The three facts checked out: the accessor read `labelLookup.find()` and `getDataByInternalId()` with no lock at all (tolerable while it was test-only and single-threaded), `insertVectorToHNSW` holds `mainIndexGuard.lock_shared()` plus `lockIndexDataGuard()` and then calls `storeNewElement`, which rehashes the map and can resize the data blocks — so a shared main lock does not exclude it.

I put the `shared_lock(indexDataGuard)` in the accessor rather than in the tiered wrapper, for two reasons:

1. `computeUnifiedIndexLabelsSetUnsafe` already establishes that division — it holds the outer locks and lets `getLabelsSet` take the inner one.
2. The tiered lookup lives on `VecSimTieredIndex`, where `backendIndex` is a `VecSimIndexAbstract *`; `lockSharedIndexDataGuard` is HNSW-specific and unreachable from there. Guarding inside the accessor also covers every other caller rather than just this path.

The wrapper keeps its flat/main guards for tier selection and the frontend read, and the split is documented on the method so it is not left implicit.

Adjacent and pre-existing, so not touched here: `getHNSWElementNeighbors` on the tiered index takes only `mainIndexGuard.lock_shared()` before reading the backend graph, which is the same exposure. It is debug-only (`FT.DEBUG DUMP_HNSW`).

src/VecSim/vec_sim_tiered_index.h
```diff
@@ -115,6 +115,29 @@ class VecSimTieredIndex : public VecSimIndexInterface {
...
+     * is exactly when a document is most likely to be updated again. The guards are the ones
+     * `relabelVector` takes, in the same order.
+     */
+    void getDataByLabel(labelType label, std::vector<std::vector<DataType>> &vectors_output) const {
+        this->flatIndexGuard.lock_shared();
+        if (this->frontendIndex->isLabelExists(label)) {
+            this->frontendIndex->getDataByLabel(label, vectors_output);
+            this->flatIndexGuard.unlock_shared();
+            return;
+        }
```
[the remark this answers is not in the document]
nonirosenfeldredis: Correct, and fixed in d012e58f. The early return was right only for a single-value index; in a multi-value one the tiers hold different subsets of a label while an ingest is pending, so the buffer alone omitted everything already ingested.

The condition now mirrors `getDistanceFrom_Unsafe`, which faced the same choice: short-circuit on a buffer hit only when `!backendIndex->isMultiValue()`, otherwise read the backend too.

Two properties of a tiered read that the fix cannot remove, now documented on the method instead:

- The order is buffer-then-backend, which for a split label is not insertion order.
- An ingest job inserts into the backend before removing from the buffer, so a vector caught inside that window is reported by both tiers and appears twice.

Both only ever make an equality-testing caller answer "different", which is the safe direction. `flatIndexGuard` is held across both reads — it cannot prevent the duplicate, but it does stop the buffer's copy disappearing between the two reads, which is what would turn a duplicate into an omission.

Added `HNSWTieredIndexTestBasic.getDataByLabelSpansBothTiers`, covering the split, the fully-ingested state, and an absent label. I checked it fails against the previous implementation with the subset it returned. That also closes the coverage gap noted in the PR description — no existing test called the tiered wrapper's `getDataByLabel`.

src/VecSim/algorithms/hnsw/hnsw_single.h
```diff
@@ -44,18 +44,27 @@ class HNSWIndex_Single : public HNSWIndex<DataType, DistType> {
...
+        auto it = labelLookup.find(label);
+        if (it == labelLookup.end()) {
+            return;
+        }
 
         auto vec = std::vector<DataType>(this->dim);
         // Only copy the vector data (dim * sizeof(DataType)), not any additional metadata like the
         // norm
-        memcpy(vec.data(), this->getDataByInternalId(id), this->dim * sizeof(DataType));
+        memcpy(vec.data(), this->getDataByInternalId(it->second), this->dim * sizeof(DataType));
```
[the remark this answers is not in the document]
nonirosenfeldredis: Confirmed, and worse than "always treats vectors as changed" — it was an out-of-bounds read. Fixed in 09683d83.

Walking the numbers: `VecSimParams_GetStoredDataSize` gives `sizeof(type) * dim` for a plain index, but `hnsw_factory.cpp` substitutes `GetSQ8StoredDataSize(dim, with_norm)`, which is `dim` quantized bytes plus 3–5 `MetadataType` floats. For float32 that is smaller than `dim * sizeof(DataType)` for any dim above ~5. Vectors are packed back-to-back at `storedDataSize` inside a block (`vec_sim_index.h` says so explicitly), so the copy ran into the neighbouring element — and past the block for the last one — then reinterpreted quantized bytes and metadata as floats.

The fix refuses rather than dequantizes: `getStoredDataSize() < dim * sizeof(DataType)` appends nothing, which the contract already defines as "cannot tell". That distinguishes the two reasons a stored element can differ in size from its elements — extra metadata *appended* (int8/uint8 under cosine, where the elements still come first and the copy is correct) versus a *compressed* representation (SQ8, unreadable as values). Applied to all four accessors, not just HNSW single.

You were also right that the docstring was claiming something untrue — it said the elements were "whatever that index can reconstruct". Nothing reconstructs; that sentence is gone and replaced with the refusal.

Covered by `HNSWSQ8Test.getDataByLabelReportsNothingForQuantizedStorage`, which asserts the size premise first so it cannot silently stop testing the refusal if SQ8's layout changes.

One related thing this turned up, fixed in the same commit: `SVSIndex::getDataByLabel` lived in a test-only block asserting "Not implemented", relying on the base default outside it. A tiered SVS index reaches it through `VecSimTieredIndex::getDataByLabel` once a vector is ingested, so that assert was reachable from a test build. It now has one unconditional definition that appends nothing, and with every concrete class defining the method the base is pure virtual again.

src/VecSim/algorithms/hnsw/hnsw_single.h
```diff
@@ -44,18 +44,34 @@ class HNSWIndex_Single : public HNSWIndex<DataType, DistType> {
...
-
-        auto id = labelLookup.at(label);
+        // `labelLookup` and the data blocks are mutated under `indexDataGuard` by callers that
+        // hold no more than a shared main lock -- tiered ingest and `markDelete` both do -- so
+        // reading them needs this guard, not the caller's. Same reason `getLabelsSet` takes it.
+        // A quantized index stores fewer bytes than the elements occupy (SQ8 keeps one byte per
+        // dimension plus metadata), so copying `dim * sizeof(DataType)` would read past the
+        // element and reinterpret the compression as values. Nothing here dequantizes, so the
+        // honest answer is none: per the contract an empty output reads as "cannot tell".
+        if (this->getStoredDataSize() < this->dim * sizeof(DataType)) {
```
dor-forer: `getStoredDataSize() < dim * sizeof(DataType)` does not reliably identify quantized storage. For small dimensions, SQ8 metadata can make the stored representation larger than the raw elements, causing compressed bytes to be returned as `DataType` values. Can we use `isQuantized` here and add a small-dimension SQ8 test (for example, FP32/L2 with dim = 4)?

src/VecSim/algorithms/svs/svs.h
```diff
@@ -799,14 +799,23 @@ class SVSIndex : public VecSimIndexAbstract<svs_details::vecsim_dt<DataType>, fl
...
+     * the answer a caller needs; a comparison against what this index stores is not available.
+     *
+     * Defined unconditionally, and not left to a default: a tiered SVS index reaches this
+     * through `VecSimTieredIndex::getDataByLabel` as soon as a vector has been ingested, so a
+     * definition that exists only in test builds is a definition that aborts exactly there.
+     */
     void getDataByLabel(
         labelType label,
         std::vector<std::vector<svs_details::vecsim_dt<DataType>>> &vectors_output) const override {
-        assert(false && "Not implemented");
```
dor-forer, line 805: just wondering and trying to make things clear for me - what is the expected behavior for a tiered multi-value SVS index when a label is split across the frontend and backend? `SVSIndex::getDataByLabel()` appends nothing, the tiered returns only the frontend vectors, and the caller cannot tell that the result is partial. Is a partial response acceptable for the intended caller? If not, should we disable this for SVS?

src/VecSim/vec_sim_tiered_index.h
```diff
@@ -115,6 +115,42 @@ class VecSimTieredIndex : public VecSimIndexInterface {
...
+     * `flatIndexGuard` is held across both reads, in the order `relabelVector` and
+     * `acquireSharedLocks` take: it cannot prevent a duplicate, but it does stop the buffer's
+     * copy being removed between them. The backend's own data guard is deliberately not taken
+     * here -- its `getDataByLabel` takes it, because a shared main lock does not exclude an
+     * ingest mutating under `indexDataGuard`. Same division as
+     * `computeUnifiedIndexLabelsSetUnsafe`, which holds the outer locks and lets `getLabelsSet`
+     * take the inner one.
+     */
+    void getDataByLabel(labelType label, std::vector<std::vector<DataType>> &vectors_output) const {
+        this->flatIndexGuard.lock_shared();
```
dor-forer: Can we use `std::shared_lock<std::shared_mutex>` for both `flatIndexGuard` and `mainIndexGuard` here?

src/VecSim/algorithms/brute_force/brute_force_single.h
```diff
@@ -43,19 +43,29 @@ class BruteForceIndex_Single : public BruteForceIndex<DataType, DistType> {
...
+            return;
+        }
 
         auto vec = std::vector<DataType>(this->dim);
         // Only copy the vector data (dim * sizeof(DataType)), not any additional metadata like the
         // norm
-        memcpy(vec.data(), this->getDataByInternalId(id), this->dim * sizeof(DataType));
+        memcpy(vec.data(), this->getDataByInternalId(it->second), this->dim * sizeof(DataType));
         vectors_output.push_back(vec);
```
dor-forer: Mayne now that it is production code and can happen more frequently we should consider not copy the vectors and use:
`vectors_output.push_back(std::move(vec));` instead (applies to all of the `getDataByLabel` functions)?

src/VecSim/vec_sim_tiered_index.h
```diff
@@ -115,6 +115,42 @@ class VecSimTieredIndex : public VecSimIndexInterface {
...
+     *
+     * `flatIndexGuard` is held across both reads, in the order `relabelVector` and
+     * `acquireSharedLocks` take: it cannot prevent a duplicate, but it does stop the buffer's
+     * copy being removed between them. The backend's own data guard is deliberately not taken
+     * here -- its `getDataByLabel` takes it, because a shared main lock does not exclude an
+     * ingest mutating under `indexDataGuard`. Same division as
+     * `computeUnifiedIndexLabelsSetUnsafe`, which holds the outer locks and lets `getLabelsSet`
+     * take the inner one.
+     */
+    void getDataByLabel(labelType label, std::vector<std::vector<DataType>> &vectors_output) const {
```
dor-forer, line 149: Not sure if I missed it but according to implementation we assume that vectors_output is empty when we receive it to the function.
Maybe we should add it to the comment or add a debug assertion `vectors_output.empty()` at the beginning of the function.
Sorry for the nit picking, it's just the the code is now in prod so we better be safe.

src/VecSim/algorithms/hnsw/hnsw_single.h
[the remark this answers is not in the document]
nonirosenfeldredis: Right on both counts. Fixed in 6f0ef780: the guard now tests `this->isQuantized` instead of comparing sizes, and the test moves to dim 4.

Working the arithmetic: SQ8 stores `dim * 1` bytes plus `storage_metadata_count` FP32 slots — 4 for L2, 3 for IP, 5 for IP with norm. Against FP32 raw elements the crossover is around dim 5, so at dim 4 the quantized element is *larger* (20 bytes against 16 for L2), the inequality is false, and 16 bytes of quantized data and metadata get copied out as four floats. Exactly the outcome the guard was added to prevent.

`isQuantized` is the right test and I should have used it from the start — it is set by the factory alongside `storedDataSize`, and `HNSWIndex_Single::getDistanceFrom_Unsafe` already branches on it for the same reason. Applied to all four accessors.

Worth recording why this survived review the first time: the test I wrote used dim 40, where the size comparison happens to hold. It asserted `getStoredDataSize() < dim * sizeof(DataType)` as its premise, so it encoded the wrong assumption rather than testing it. The test now covers both sides of the crossover — dim 4 asserting the element is *larger*, dim 40 asserting it is smaller — and I checked the dim 4 case fails against the previous guard before trusting it.

Full suite: 2784/2784.

src/VecSim/algorithms/hnsw/hnsw_single.h
nonirosenfeldredis: done

src/VecSim/vec_sim_tiered_index.h
nonirosenfeldredis: done

src/VecSim/algorithms/brute_force/brute_force_single.h
nonirosenfeldredis: indeed, done

src/VecSim/algorithms/svs/svs.h
nonirosenfeldredis: agreed, since both the svs methods: getRelabelData & relabelVector not implemented in SVS repo. lets return an error up front and not wait for the backend svs to respond.

dor-forer: I see that the current change skips the SVS backend but still returns vectors from the frontend, so it still returns the the partial result that looks successful. Should we ust block the SVS `getDataByLabel` at all, and allow until it is implemented?

ofiryanai: I suppose for quantized backend that's also relevant to check the backend and returning right away instead of returning results only if the vector happened to be on the frontend. Disk HNSW might behave like that (if we return from in memory rather than go to disk for this API, it's probably not decided yet)

nonirosenfeldredis: I gated it in the tiered class : checking for svs . when I added quantizied check - claude replied that i

> ts not reachable to get here . Quantized: cannot be tested, and the check cannot fire. WHY? tiered_factory.cpp rejects a quantized backend outright — NewIndex returns nullptr and EstimateInitialSize throws "Quantization is not supported for tiered HNSW indexes", because the brute-force frontend isn't quantized and the stored layouts would be incompatible. I also checked the enterprise disk backend, which does store SQ8: it never sets isQuantized, so it reports false too. so no need for quantized

nonirosenfeldredis: @dor-forer you can look now

ofiryanai (approved):

PR #1027, opened 2026-08-27
[MOD-18471] Support fine-grained SVS index

ethanglaser:
Adds support for using newly-added separate fine-grained SVS index with minor corresponding test revisions

Includes SVS 0.5.0 rc1 binaries, which will eventually be swapped out for the official release binaries after validation.

Redisearch CI failures at `test_vecsim_svs.py:711` and `test_vecsim_svs.py:729` come from changes in internal logic not being reflected in tests yet.
1. Each deletion now submit a single label consolidation job, that can change size of the reverse edges lists. This breaks the `test_vecsim_svs.py:711`.
2. Some of the submitted consolidation jobs can be unfinished before GC launch. This breaks the `test_vecsim_svs.py:729`.

---

> [!NOTE]
> **High Risk**
> Large concurrent-mutation and tiered-async rewrite across SVS/HNSW locking, ingestion, delete/relabel/GC paths, plus a new prebuilt binary dependency.
> 
> **Overview**
> **Upgrades SVS to 0.5.0rc2 prebuilts** and switches CMake download URLs to the nightly concurrent builds; **GCC shared-lib support now requires GLIBC ≥ 2.28** (older 2.26 tarball path removed), and `replace_external_id` detection targets `svs/concurrent/dynamic_index.h`.
> 
> **SVS backend** moves from `svs::index::vamana` to **`svs::concurrent`** mutable Vamana indexes, with **segmented blocked** storage/graph allocators, sequence-locked graphs, and reverse-edge accounting in size estimates. The implementation is held in a **`shared_ptr` guarded by mutexes**, supports lazy init via `ready()` / `addVectorsIfInitialized`, per-label **`consolidate`**, and maps relabel/GC to the concurrent APIs (`compact()` instead of separate consolidate in GC).
> 
> **Tiered SVS ingestion is redesigned**: batch “update with swap journal” is replaced by **per-vector insert jobs**, a **training batch** (`ids_to_init_`) that runs **`initSVSIndex`**, async **consolidate jobs** after deletes, and richer **relabel** coordination (delayed inserts, takeover of pending consolidations). Tiered batch iteration no longer holds the main index lock for SVS queries.
> 
> **Shared tiered infrastructure**: `TieredInsertJob`, invalid-job handling, `removeIngestedVectorFromFlat`, and default **`indexSize` / `getDistanceFrom`** live in `VecSimTieredIndex`; **`ScopedLocks`** lets HNSW lock flat + HNSW data guard for size while tiered SVS skips backend query locks. HNSW tiered code delegates insert-job bookkeeping to the base class.
> 
> Tests/bindings: memory estimates include reverse edges; flow tests call **`run_gc()`** before memory checks; Python **`runGC()`** calls the index directly so non-tiered SVS compacts too.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ca35f7c72b46f382fbb86fa66c46733f59f07a4c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/VecSim/algorithms/svs/svs_tiered.h
```diff
@@ -802,108 +942,162 @@ class TieredSVSIndex : public VecSimTieredIndex<DataType, float> {
...
             // compression/training.
-            if (this->backendIndex->indexSize() == 0) {
+            if ((!svs_index->ready()) && (!this->backendInitSubmited.load(std::memory_order_acquire))) {
                 // If backend index size is 0, first collect vectors in frontend index
                 // lock in scope to ensure that these will be released before
-                // updateSVSIndexWrapper() is called.
+                // initSVSIndexWrapper() is called.
                 {
                     std::lock_guard lock(this->flatIndexGuard);
+                    if (this->frontendIndex->isLabelExists(label)) {
```
ethanglaser: This condition should have the same guard as line 1002, otherwise every vector after the first of a multi-value doc wipes siblings out of the frontend buffer
```suggestion
                    if (!this->frontendIndex->isMultiValue() && this->frontendIndex->isLabelExists(label)) {
```
This update should lead to the redisearch `TestIndexMultiValueJsonReload` tests go green

src/VecSim/algorithms/svs/svs_tiered.h
```diff
@@ -670,109 +758,161 @@ class TieredSVSIndex : public VecSimTieredIndex<DataType, float> {
...
+
+            for (idType id : ids_to_init_) {
+                ids_to_move.push_back(id);
+                labels_to_move.push_back(flat_index->getVectorLabel(id));
+                auto data = flat_index->getDataByInternalId(id);
                 vectors_to_move.insert(vectors_to_move.end(), data, data + dim);
             }
-            // reset journal to the current frontend index state
-            swaps_journal.clear();
-            deleted_labels_journal.clear();
```
ethanglaser, line 694: `deleted_labels_journal`, `swaps_journal`, `applySwapsToLabelsArray` usage has been removed in current state - if this functionality is replaced and not going to be re-integrated into the code, then delete instantiation/definition of these

But it may be necessary to restore usage here to address review comments above

src/VecSim/algorithms/svs/svs_tiered.h
```diff
@@ -670,109 +758,161 @@ class TieredSVSIndex : public VecSimTieredIndex<DataType, float> {
...
+        }
+
+        // Copy the vector blob out of the flat buffer while holding flatIndexGuard, so we
+        // can release the flat lock before indexing into the SVS backend
+        size_t data_size = this->frontendIndex->getStoredDataSize();
+        auto blob_copy = this->getAllocator()->allocate_unique(data_size);
+        memcpy(blob_copy.get(), this->frontendIndex->getDataByInternalId(job->id), data_size);
+        this->flatIndexGuard.unlock_shared();
+
+        svs_index->addVector(blob_copy.get(), job->label);
```
ethanglaser, line 743: Once a job passes the `job->isValid` check at line 776 and releases the guard at 791, nothing orders its backend insert against a concurrent `deleteVector`. `deleteVector` (line 1133) invalidates pending insert jobs under `flatIndexGuard`, but that's too late for a job already past 776: the delete runs to completion, its `backendIndex->deleteVector(label)` returns 0 because the vector isn't in the backend yet, and then line 793 puts it there. Result is a backend entry for a deleted doc that is not marked deleted, so GC never reclaims it and label accounting can't see it.

This was previously addressed in 2 ways, both of which are removed in this PR: `mainIndexGuard` held exclusively across the backend add and the backend delete (significant usage reduction in this PR), and the deletions journal (usage removed in this PR)

This resolution should lead to the redisearch `test_delete_during_background_indexing` tests go green

src/VecSim/algorithms/svs/svs_tiered.h
```diff
@@ -918,65 +1112,85 @@ class TieredSVSIndex : public VecSimTieredIndex<DataType, float> {
...
-        if (label_exists) {
-            std::lock_guard lock(this->mainIndexGuard);
-            ret += this->backendIndex->deleteVector(label);
+        auto deleted = this->backendIndex->deleteVector(label);
+        if (deleted > 0) {
+            if (this->getWriteMode() == VecSim_WriteInPlace) {
+                GetSVSIndex()->consolidate({label});
+            } else {
+                scheduleSVSIndexConsolidate(label);
+            }
```
ethanglaser, line 1142: Redisearch `test_gc` tests are failing due to increased memory from per-deleted label consolidation, as well as marked-delete count and pending-job count assertions - is it possible to do this on batches or have a single deferred job that accumulates labels?

razdoburdin: it will return latency spikes back.
I think the test should be changed.

src/VecSim/algorithms/svs/svs_tiered.h
razdoburdin: The concurrent machinery is ok here. The reason is the meaning of `num_marked_deleted` field. 
I changed it from "soft deleted but not consolidated" to "soft deleted but not compacted" to align with test expectation.

src/VecSim/algorithms/svs/svs_tiered.h
razdoburdin: I checked the test manually:
1. Assertion at line 705 should pass now.
2. Assertion at line 711 is invalid. Deletions can increase memory footprint now because of reverse_edges.
3. Assertion at line 729 is invalid as well. Each deletion submit consolidation job, so we may have some unfinished jobs before GC call.

src/VecSim/algorithms/svs/svs.h
```diff
@@ -758,15 +830,12 @@ class SVSIndex : public VecSimIndexAbstract<svs_details::vecsim_dt<DataType>, fl
...
-            // https://intel.github.io/ScalableVectorSearch/python/dynamic.html#svs.DynamicVamana.consolidate
-            impl_->consolidate();
+        if (auto impl = getImpl()) {
             // There is documentation for compact():
             // https://intel.github.io/ScalableVectorSearch/python/dynamic.html#svs.DynamicVamana.compact
-            impl_->compact();
+            impl->compact();
         }
-        num_marked_deleted = 0;
+        num_marked_deleted.store(0, std::memory_order_relaxed);
```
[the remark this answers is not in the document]
razdoburdin: false positive.
comment updated to avoid confusion.

src/VecSim/algorithms/svs/svs_tiered.h
```diff
@@ -639,142 +684,207 @@ class TieredSVSIndex : public VecSimTieredIndex<DataType, float> {
...
-
-            std::sort(deleted_labels_during_update.begin(), deleted_labels_during_update.end());
-            auto it = std::unique(deleted_labels_during_update.begin(),
-                                  deleted_labels_during_update.end());
-            deleted_labels_during_update.erase(it, deleted_labels_during_update.end());
-            auto svs_index = GetSVSIndex();
-            svs_index->deleteVectors(deleted_labels_during_update.data(),
-                                     deleted_labels_during_update.size());
-        }
+        } // release frontend index
```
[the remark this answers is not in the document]
razdoburdin: false positive.
Init holds exclusive flat during training, but training on 1e4 data is much faster than 40 sec.

src/VecSim/algorithms/svs/svs_tiered.h
```diff
@@ -794,176 +904,311 @@ class TieredSVSIndex : public VecSimTieredIndex<DataType, float> {
...
-            // overwise it is updateTriggerThreshold.
-            std::shared_lock lock(this->mainIndexGuard);
-            update_threshold = this->backendIndex->indexSize() == 0 ? this->trainingTriggerThreshold
-                                                                    : this->updateTriggerThreshold;
-        }
-        if (frontend_index_size >= update_threshold) {
-            scheduleSVSIndexUpdate();
+
+            // Insert job to the queue and signal the workers' updater.
+            this->submitSingleJob(new_insert_job);
```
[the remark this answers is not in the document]
razdoburdin: false positive. SVS handles such cases by itself.

cmake/svs.cmake
```diff
@@ -76,12 +76,10 @@ if(USE_SVS)
...
-                set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/v0.3.2/svs-shared-library-reduced-gcc14.tar.gz" CACHE STRING "SVS URL")
+		    set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/nightly/svs-shared-library-0.5.0rc2-reduced-gcc14.tar.gz" CACHE STRING "SVS URL")
             else()
-                set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/v0.3.2/svs-shared-library-reduced.tar.gz" CACHE STRING "SVS URL")
+                set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/nightly/svs-shared-library-0.5.0rc2-reduced.tar.gz" CACHE STRING "SVS URL")
             endif()
-        elseif(GLIBC_VERSION VERSION_GREATER_EQUAL "2.26")
-            set(SVS_URL "https://github.com/intel/ScalableVectorSearch/releases/download/v0.3.0/svs-shared-library-0.3.0-reduced-glibc2_26.tar.gz" CACHE STRING "SVS URL")
         else()
             message(STATUS "GLIBC >= 2.26 is required for GCC SVS binaries - disabling SVS_SHARED_LIB")
```
ethanglaser:
```suggestion
            message(STATUS "GLIBC >= 2.28 is required for GCC SVS binaries - disabling SVS_SHARED_LIB")
```

src/VecSim/algorithms/svs/svs.h
```diff
@@ -259,83 +291,107 @@ class SVSIndex : public VecSimIndexAbstract<svs_details::vecsim_dt<DataType>, fl
...
+            } else if (!allow_init) {
+                this->implMutationGuard_.unlock();
+                return kNotInitialized;
+            } else {
+                auto built = initImpl(points, ids);
+                assert(built != nullptr);
+                std::lock_guard<std::shared_mutex> lock(this->pimplGuard_);
+                this->impl_ = std::move(built);
+            }
+            this->implMutationGuard_.unlock();
```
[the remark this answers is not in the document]
razdoburdin: false positive.
If the exception reaches this point uncaught, execution will terminate, so mutex lock leaks are not a concern

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -81,17 +78,10 @@ template <typename DataType, typename DistType>
 class TieredHNSWIndex : public VecSimTieredIndex<DataType, DistType> {
 private:
     /// Mappings from id/label to associated jobs, for invalidating and update ids if necessary.
-    // In MULTI, we can have more than one insert job pending per label.
     // **This map is protected with the flat buffer lock**
```
alonre24, line 81: Is this comment correct for `idToRepairJobs` or was it only true for `labelToInsertJobs` that has been removed?

src/VecSim/algorithms/svs/svs.h
```diff
@@ -113,8 +129,21 @@ class SVSIndex : public VecSimIndexAbstract<svs_details::vecsim_dt<DataType>, fl
...
-    // SVS Index implementation instance
-    std::unique_ptr<impl_type> impl_;
+    // SVS Index implementation instance.
+    // initImpl() needs at least one point to compute an entry point.
+    // so impl_ is null until the first insert
+    std::shared_ptr<impl_type> impl_;
+
+    mutable std::shared_mutex pimplGuard_;
+    // Serialize mutators of the impl_ pointer (exclusive)
+    // vs add_points() (shared) to avoid silent drops of additions
```
alonre24, line 139: Didn't understand the comment. Why do we need both locks?

src/VecSim/algorithms/svs/svs.h
```diff
@@ -259,83 +291,107 @@ class SVSIndex : public VecSimIndexAbstract<svs_details::vecsim_dt<DataType>, fl
     // for the operation.
     // Important NOTE: For single vector operations (n=1), parallelism should be 1.
     // For bulk operations (n>1), parallelism should reflect the number of available threads.
```
alonre24, line 293: is this comment still true? If not, please update accordingly and explain the purpose of the ' allow_init ' parameter

src/VecSim/algorithms/svs/svs.h
```diff
@@ -259,83 +291,107 @@ class SVSIndex : public VecSimIndexAbstract<svs_details::vecsim_dt<DataType>, fl
...
 
-        const auto deleted_num = impl_->delete_entries(std::span{&label, 1});
-
-        this->markIndexUpdate(deleted_num);
+        const int deleted_num = impl->delete_entries(std::span{&label, 1});
+        this->markIndexUpdate(impl, deleted_num);
         return deleted_num;
     }
 
     int deleteVectorsImpl(const labelType *labels, size_t n) {
```
alonre24, line 355: are we ever going to use it with more than one label?

src/VecSim/algorithms/svs/svs_tiered.h
```diff
@@ -280,21 +303,15 @@ class TieredSVSIndex : public VecSimTieredIndex<DataType, float> {
...
+
+        // IDs of all the results returned so far, so that neither tier can yield one twice.
+        // Both directions have to be covered: nothing freezes the backend for the iterator's
+        // lifetime, so an insert job can publish a label into SVS and only then drop it from the
+        // flat buffer, while the flat iterator still holds it in the snapshot it computed on its
+        // first call. Either tier can therefore surface a label first.
         vecsim_stl::unordered_set<labelType> returned_results_set;
 
-        VecSimQueryReply *compute_current_batch(size_t n_res, bool isMultiValue) {
+        VecSimQueryReply *compute_current_batch(size_t n_res) {
```
alonre24, line 314: Is this logic identical to the tiered HNSW one? Can we move this to the base tiered index level?

src/VecSim/algorithms/svs/svs_tiered.h
```diff
@@ -548,30 +537,61 @@ class TieredSVSIndex : public VecSimTieredIndex<DataType, float> {
...
+        auto *insert_job = static_cast<SVSInsertJob *>(job);
+        auto *job_index = static_cast<TieredSVSIndex<DataType> *>(insert_job->index);
+        InsertJobOutcome outcome;
+        {
+            // prevent parallel execution with index initilizing job
+            std::shared_lock<std::shared_mutex> lock(job_index->updateJobMutex);
+            outcome = job_index->executeInsertJob(insert_job);
+        }
+
+        if (outcome == InsertJobOutcome::Deferred) {
```
alonre24, line 590: Can we simplify this logic by initializing the backend index from the main thread (in `addVector`) and allowing the execute insert job _while_ vectors are moved from the frontend to the backend index (similar to how it is done in HNSW when we use SQ8)?

src/VecSim/algorithms/svs/svs_tiered.h
```diff
@@ -794,176 +926,315 @@ class TieredSVSIndex : public VecSimTieredIndex<DataType, float> {
...
+                // We are going to update the internal id that currently holds the vector associated
+                // with the given label.
+                new_flat_id =
+                    dynamic_cast<BruteForceIndex_Single<DataType, DistType> *>(this->frontendIndex)
+                        ->getIdOfLabel(label);
+                // If we are adding a new element (rather than updating an exiting one) we may need
+                // to increase index capacity.
+            }
+            // If this label already exists, this will do overwrite.
+            ret += this->frontendIndex->addVector(blob, label);
```
alonre24, line 1041: is this logic common with HNSW? Can it move to the base tiered class?

src/VecSim/algorithms/svs/svs.h
```diff
@@ -259,83 +291,107 @@ class SVSIndex : public VecSimIndexAbstract<svs_details::vecsim_dt<DataType>, fl
...
+        if (labelCountOf(*impl) == 0) {
+            std::lock_guard<std::shared_mutex> replace_lock(this->implMutationGuard_);
+            if (getImpl() != impl) {
+                // The instance was already droped
+                return;
+            }
+            if (labelCountOf(*impl) == 0) {
+                {
+                    std::lock_guard<std::shared_mutex> lock(this->pimplGuard_);
+                    this->impl_.reset();
```
alonre24, line 379: I think we should simplify and avoid `reset()` completely. Once the SVS backend has been trained, we don't need to return to an uninitialized phase again, even if all vectors are removed. Hopefully, that would simplify flows and reduce synchronization mechanisms

alonre24 (commented):
I have completed my initial review.
My general impression is that the tiered svs logic can still be simplified significantly, and follow patterns that we use for tiered HNSW with SQ8 compression. The two indexes share the exact same lifetime principles (accumulate a batch of vectors in frontend before training, initalize backend, then move vectors to backend one by one), but tiered svs logic seems much more involved. I don't see how the difference in the underline backend in the fine-grained appraoch requires such a difference.

src/VecSim/algorithms/svs/svs_tiered.h
razdoburdin: done in the follow-up PR

src/VecSim/algorithms/svs/svs_tiered.h
razdoburdin: Unfortunately is doesn't look like a good option.
Correct me if I am wrong, HNSW initialize index and than moves the corresponding vectors by a separate async jobs, right?

For SVS we can do it only by a single operation, i.e. initialization and actual data movement will be done in the main thread.

src/VecSim/algorithms/svs/svs.h
razdoburdin: not actual anymore

src/VecSim/algorithms/svs/svs.h
razdoburdin: done

src/VecSim/algorithms/svs/svs.h
razdoburdin: it is still can be used in python bindings and tests

src/VecSim/algorithms/svs/svs.h
razdoburdin: still true

src/VecSim/algorithms/hnsw/hnsw_tiered.h
razdoburdin: labelToInsertJobs only

src/VecSim/algorithms/svs/svs_tiered.h
razdoburdin: done

ethanglaser:
Replaced by https://github.com/RedisAI/VectorSimilarity/pull/1058

PR #1028, opened 2026-09-01
[MOD-17526] Compute SQ8-FP32 L2 via direct residual accumulation, not cancellation

dor-forer:
## Summary
`SQ8_FP32_L2Sqr` and its SIMD variants (SSE4/AVX2/AVX2-FMA/AVX512/NEON/SVE/SVE2) computed L2² via `||x||²+||y||²-2*IP(x,y)`. That identity is exact in real arithmetic but catastrophically cancels in FP32 when x and y share a large common offset relative to their spread: for x=[100000,100008], y=[100000,100000] the true L2² is 64 and the old kernel could return 0, a negative number, or unbounded garbage. Because these kernels serve HNSW graph construction as well as query evaluation, wrong distances select wrong neighbours and the damage is not recoverable by re-querying.

Replaced with direct residual accumulation across every ISA variant, keeping the existing loop/residual structure.

## Correctness
- Per-element `diff = delta*q + (min_val - y)`. The operand order is load-bearing: `min_val` and `y` are close in magnitude so their subtraction is exact (Sterbenz), and the small `delta*q` correction is added afterward. Computing `(min_val + delta*q) - y` instead rounds the dequantized value at the large offset's precision and silently reinstates the cancellation — the scalar path originally did exactly that and was fixed during review.
- `-ffast-math` / `-Ofast` would reassociate this back into the bug. The repo's `-O3` builds are safe; there's a comment on the kernel saying so.
- Query/storage blob layout, SQ8-SQ8, SQ8-FP16, and IP/Cosine kernels are untouched. Removing the now-unused `y_sum`/`y_sum_sq` from the L2 query blob is deferred: `query_metadata_count` is keyed by metric, not datatype, so it would change the FP16 L2 layout while FP16 still uses the identity.

## Tests
New regression tests, all checked against an independent double-precision reference rather than another kernel sharing the same formula (pre-fix, scalar and SIMD were wrong *together*, so kernel-vs-kernel proved nothing):
- **Translation invariance** — `L2(x,y) == L2(x+C,y+C)` for large C, at dims 128/129/145/512/513/527/768 so the cancellation case also runs through the residual/masked-lane tails.
- **Ticket repro shape** — the literal `[100000,100008]` vs `[100000,100000]` case.
- **Scalar associativity** — pins the ordering bug above; verified to fail against the old order and pass against the fix.
- **Regular vectors** — ordinary small-magnitude embeddings, confirming the common case didn't regress.

## Performance
Kernels got faster, not slower, measured with the repo's own `bm_spaces_sq8_fp32` (median of 5-9 reps, cv ≤ 1%):

| kernel | dim | before | after |
|---|---|---|---|
| NEON (one 16-byte load per step) | 1024 | 502 ns | 370 ns (−26%) |
| NEON | 64 | 34.5 ns | 25.8 ns (−25%) |
| AVX2+FMA (tree reduction) | 16 | 4.42 ns | 3.10 ns (−30%) |
| AVX2 (tree reduction) | 16 | 5.25 ns | 3.91 ns (−26%) |
| SSE4 (masked residual load) | 513-527 | 125-127 ns | 116-119 ns |
| SVE2 (no `udiv`, VL-aligned loads) | 1024 | 160 ns | 141 ns (−12%) |
| SVE | 1024 | 164 ns | 153 ns (−6.7%) |

The SVE loop restructure is **not** a uniform win and the regression is stated rather than hidden: on the plain SVE tier, dims that leave a tail are ~8% slower (dim 513: 78.3 → 84.6 ns) while dims that divide evenly by the vector length gain. SVE2 improves at both. Kept because SVE2 is the tier the chooser prefers when both are present, and because real embedding dims (128/256/384/512/768/1024/1536) divide evenly at 128- and 256-bit VL. The kernel comment records why the shape differs and points at the ticket; the numbers and the rejected tail variant are in the commit messages.

Left alone deliberately, per measurement: AVX512 stays at 2 accumulators and keeps `_mm512_reduce_add_ps` (already a tree); the scalar path stays at 1 accumulator since it only ever sees dims 1-7.

## Test plan
- [x] `SQ8` suite: 595/595 on `dorer-intel` (Xeon Platinum 8375C — AVX512/VNNI, AVX2+FMA, AVX2, SSE4)
- [x] `SQ8` suite under **ASAN**: 595/595 on the same host
- [x] `SQ8` suite: 593/593 on Neoverse-N1 (NEON)
- [x] **`SQ8` suite: 593/593 on Graviton4 / Neoverse-V2 — SVE and SVE2 kernels executed for real.** An earlier ARM host was Neoverse-N1, whose feature flags carry no `sve`, so the runtime-gated SVE/SVE2 checks had been skipping silently and the suite passed without ever entering those kernels.
- [x] SVE loop restructure also verified off-hardware: exhaustive coverage check over `chunk ∈ {4,8,16,32,64} × dim ∈ [1,4096]`, every element touched exactly once, no overread; plus disassembly confirming `udiv` count 3 → 0 (and 0 across all 32 instantiations in `SVE.cpp.o` / `SVE2.cpp.o`)

Running on real SVE hardware immediately earned its keep: it reproduced a bug Cursor Bugbot flagged in the restructured tail (`svmla_f32_z` zeroing live accumulator lanes), failing on exactly the dims where `dim % chunk != 0`. Fixed in fc7995d4.

---

> [!NOTE]
> **High Risk**
> Changes core HNSW distance math for SQ8–FP32 L2 across all ISA kernels; wrong distances corrupt index structure, though extensive regression tests and unchanged blob layouts limit rollout risk.
> 
> **Overview**
> Fixes **MOD-17526**: SQ8–FP32 L2² no longer uses `||x||² + ||y||² − 2·IP`, which could return 0, negative, or garbage when vectors share a large common offset (e.g. true L2² = 64 for the ticket repro). **Scalar and every SIMD path** (SSE4, AVX2, AVX2+FMA, AVX512, NEON, SVE/SVE2) now sum **per-component squared residuals** `Σ (delta·q + (min − y))²`, with **`min − y` before fusing dequant** so FP32 rounding stays at residual scale; comments warn against `-ffast-math` reassociation.
> 
> SIMD kernels stop delegating to inner-product helpers, use multi-accumulator loops with masked tails where needed, and AVX2 tiers adopt **`my_mm256_reduce_add_ps_tree`** for cheaper horizontal sums at small dims. **Tests** add a double-precision reference, translation-invariance under large offsets, ticket repro, scalar operand-order regression, and sanity checks across dispatched ISA variants.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f8148c9ce5124948b1f98b90e03b3ea035a5ec88. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/VecSim/spaces/L2/L2.cpp
+++ b/src/VecSim/spaces/L2/L2.cpp
@@ -27,9 +27,8 @@ using sq8 = vecsim_types::sq8;
  * which catastrophically cancels in FP32 when x and y share a large common offset relative to
  * their spread.
  *
- * The operand order in the loop below is load-bearing, not stylistic: it relies on FP addition
- * NOT being reassociated. `-ffast-math` / `-Ofast` permit exactly that reassociation and would
- * reinstate the bug this kernel fixes. The repo's -O3 builds are safe; keep it that way.
+ * The operand order below relies on FP addition NOT being reassociated, so `-ffast-math` /
+ * `-Ofast` would reinstate the bug. The repo's -O3 builds are safe.
  *
  * pVect1 is storage (SQ8): [uint8_t values (dim)] [min_val] [delta] [x_sum] [x_sum_squares]
  * pVect2 is query (FP32): [float values (dim)] [y_sum] [y_sum_squares]
@@ -45,12 +44,9 @@ float SQ8_FP32_L2Sqr(const void *pVect1v, const void *pVect2v, size_t dimension)
 
     float res = 0;
     for (size_t i = 0; i < dimension; i++) {
-        // diff = dequant(x_i) - y_i = delta * q_i + (min_val - y_i). Order matters: min_val and
-        // y_i are both large and close in magnitude (Sterbenz's lemma makes their subtraction
-        // exact), so computing that first and adding the small delta*q_i correction preserves the
-        // residual. Computing (min_val + delta*q_i) - y_i first rounds the dequantized value to
-        // FP32 at the large offset's precision, discarding the residual before the subtraction
-        // ever happens -- silently reintroducing the cancellation this kernel exists to avoid.
+        // Order matters: min_val - y_i is exact (Sterbenz) since both are large and close, so
+        // adding the small delta*q_i correction afterward preserves the residual. Computing
+        // (min_val + delta*q_i) - y_i instead rounds it away at the large offset's precision.
         float diff = delta * static_cast<float>(pVect1[i]) + (min_val - pVect2[i]);
         res += diff * diff;
     }
--- a/src/VecSim/spaces/L2/L2.cpp
+++ b/src/VecSim/spaces/L2/L2.cpp
@@ -19,29 +19,33 @@ using float16 = vecsim_types::float16;
 using sq8 = vecsim_types::sq8;
 
 /*
- * Optimized asymmetric SQ8-FP32 L2 squared distance using algebraic identity:
- *   ||x - y||² = Σx_i² - 2*IP(x, y) + Σy_i²
- *              = x_sum_squares - 2 * IP(x, y) + y_sum_squares
- *   where IP(x, y) = min * y_sum + delta * Σ(q_i * y_i)
+ * Asymmetric SQ8-FP32 L2 squared distance computed via direct residual accumulation:
+ *   ||x - y||² = Σ(dequant(x_i) - y_i)²
+ *   where dequant(x_i) = min_val + delta * q_i
+ *
+ * This avoids the algebraic-identity/cancellation approach (||x||² + ||y||² - 2*IP(x, y)),
+ * which catastrophically cancels in FP32 when x and y share a large common offset relative to
+ * their spread.
  *
  * pVect1 is storage (SQ8): [uint8_t values (dim)] [min_val] [delta] [x_sum] [x_sum_squares]
  * pVect2 is query (FP32): [float values (dim)] [y_sum] [y_sum_squares]
  */
 float SQ8_FP32_L2Sqr(const void *pVect1v, const void *pVect2v, size_t dimension) {
-    // Get the raw inner product using the common implementation
-    const float ip = SQ8_FP32_InnerProduct_Impl(pVect1v, pVect2v, dimension);
-
     // Storage metadata follows a byte payload and is not necessarily float-aligned.
     const auto *pVect1 = static_cast<const uint8_t *>(pVect1v);
-    const float x_sum_sq =
-        load_unaligned<float>(pVect1 + dimension + sq8::SUM_SQUARES * sizeof(float));
-
-    // Get precomputed sum of squares from query blob (pVect2 is FP32)
     const auto *pVect2 = static_cast<const float *>(pVect2v);
-    const float y_sum_sq = pVect2[dimension + sq8::SUM_SQUARES_QUERY];
 
-    // L2² = ||x||² + ||y||² - 2*IP(x, y)
-    return x_sum_sq + y_sum_sq - 2.0f * ip;
+    const auto *params1 = pVect1 + dimension;
+    const float min_val = load_unaligned<float>(params1 + sq8::MIN_VAL * sizeof(float));
+    const float delta = load_unaligned<float>(params1 + sq8::DELTA * sizeof(float));
+
+    float res = 0;
+    for (size_t i = 0; i < dimension; i++) {
+        // diff = dequant(x_i) - y_i = min_val + delta * q_i - y_i
+        float diff = min_val + delta * static_cast<float>(pVect1[i]) - pVect2[i];
+        res += diff * diff;
+    }
+    return res;
 }
 
 /*
--- a/src/VecSim/spaces/L2/L2.cpp
+++ b/src/VecSim/spaces/L2/L2.cpp
@@ -27,6 +27,10 @@ using sq8 = vecsim_types::sq8;
  * which catastrophically cancels in FP32 when x and y share a large common offset relative to
  * their spread.
  *
+ * The operand order in the loop below is load-bearing, not stylistic: it relies on FP addition
+ * NOT being reassociated. `-ffast-math` / `-Ofast` permit exactly that reassociation and would
+ * reinstate the bug this kernel fixes. The repo's -O3 builds are safe; keep it that way.
+ *
  * pVect1 is storage (SQ8): [uint8_t values (dim)] [min_val] [delta] [x_sum] [x_sum_squares]
  * pVect2 is query (FP32): [float values (dim)] [y_sum] [y_sum_squares]
  */
--- a/src/VecSim/spaces/L2/L2.cpp
+++ b/src/VecSim/spaces/L2/L2.cpp
@@ -41,8 +41,13 @@ float SQ8_FP32_L2Sqr(const void *pVect1v, const void *pVect2v, size_t dimension)
 
     float res = 0;
     for (size_t i = 0; i < dimension; i++) {
-        // diff = dequant(x_i) - y_i = min_val + delta * q_i - y_i
-        float diff = min_val + delta * static_cast<float>(pVect1[i]) - pVect2[i];
+        // diff = dequant(x_i) - y_i = delta * q_i + (min_val - y_i). Order matters: min_val and
+        // y_i are both large and close in magnitude (Sterbenz's lemma makes their subtraction
+        // exact), so computing that first and adding the small delta*q_i correction preserves the
+        // residual. Computing (min_val + delta*q_i) - y_i first rounds the dequantized value to
+        // FP32 at the large offset's precision, discarding the residual before the subtraction
+        // ever happens -- silently reintroducing the cancellation this kernel exists to avoid.
+        float diff = delta * static_cast<float>(pVect1[i]) + (min_val - pVect2[i]);
         res += diff * diff;
     }
     return res;
--- a/src/VecSim/spaces/L2/L2.cpp
+++ b/src/VecSim/spaces/L2/L2.cpp
@@ -19,16 +19,16 @@ using float16 = vecsim_types::float16;
 using sq8 = vecsim_types::sq8;
 
 /*
- * Asymmetric SQ8-FP32 L2 squared distance computed via direct residual accumulation:
+ * Asymmetric SQ8-FP32 L2 squared distance, accumulated directly per component:
  *   ||x - y||² = Σ(dequant(x_i) - y_i)²
  *   where dequant(x_i) = min_val + delta * q_i
  *
- * This avoids the algebraic-identity/cancellation approach (||x||² + ||y||² - 2*IP(x, y)),
- * which catastrophically cancels in FP32 when x and y share a large common offset relative to
- * their spread.
+ * Accumulating residuals keeps each component's rounding at the scale of |x_i - y_i|. Expanding
+ * to ||x||² + ||y||² - 2*IP(x, y) would round at the scale of the norms, which loses the
+ * distance when x and y share a large common offset relative to their spread.
  *
- * The operand order below relies on FP addition NOT being reassociated, so `-ffast-math` /
- * `-Ofast` would reinstate the bug. The repo's -O3 builds are safe.
+ * The operand order in the loop is load-bearing and relies on FP addition not being
+ * reassociated; this file must not be built with -ffast-math / -Ofast.
  *
  * pVect1 is storage (SQ8): [uint8_t values (dim)] [min_val] [delta] [x_sum] [x_sum_squares]
  * pVect2 is query (FP32): [float values (dim)] [y_sum] [y_sum_squares]
@@ -44,9 +44,11 @@ float SQ8_FP32_L2Sqr(const void *pVect1v, const void *pVect2v, size_t dimension)
 
     float res = 0;
     for (size_t i = 0; i < dimension; i++) {
-        // Order matters: min_val - y_i is exact (Sterbenz) since both are large and close, so
-        // adding the small delta*q_i correction afterward preserves the residual. Computing
-        // (min_val + delta*q_i) - y_i instead rounds it away at the large offset's precision.
+        // Order matters. When the stored range sits far from zero, min_val and y_i are within a
+        // factor of two of each other, so min_val - y_i is exact (Sterbenz) and rounding happens
+        // only at the scale of the residual. Forming (min_val + delta*q_i) first would round at
+        // the scale of min_val and wipe out small residuals. When min_val and y_i are not close,
+        // the residual is large and either order is fine.
         float diff = delta * static_cast<float>(pVect1[i]) + (min_val - pVect2[i]);
         res += diff * diff;
     }
--- a/src/VecSim/spaces/L2/L2_AVX2_FMA_SQ8_FP32.h
+++ b/src/VecSim/spaces/L2/L2_AVX2_FMA_SQ8_FP32.h
@@ -55,7 +55,6 @@ template <unsigned char residual> // 0..31
 float SQ8_FP32_L2SqrSIMD16_AVX2_FMA(const void *pVect1v, const void *pVect2v, size_t dimension) {
     const uint8_t *pVect1 = static_cast<const uint8_t *>(pVect1v); // SQ8 storage
     const float *pVect2 = static_cast<const float *>(pVect2v);     // FP32 query
-    const uint8_t *pEnd1 = pVect1 + dimension;
 
     // Get quantization parameters from stored vector (after quantized data)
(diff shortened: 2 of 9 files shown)


src/VecSim/spaces/L2/L2_SVE_SQ8_FP32.h
```diff
@@ -21,35 +20,126 @@ namespace {
...
+        svfloat32_t v1_f = svcvt_f32_u32_z(pg_tail, v1_u32);
+        svfloat32_t v2 = svld1_f32(pg_tail, pVect2 + offset);
 
-    // Get precomputed sum of squares from storage blob (pVect1v is SQ8 storage)
-    const uint8_t *pVect1 = static_cast<const uint8_t *>(pVect1v);
-    const float x_sum_sq =
-        load_unaligned<float>(pVect1 + dimension + sq8::SUM_SQUARES * sizeof(float));
+        svfloat32_t min_minus_y = svsub_f32_z(pg_tail, min_val_vec, v2);
+        svfloat32_t diff = svmla_f32_z(pg_tail, min_minus_y, delta_vec, v1_f);
+        sum3 = svmla_f32_z(pg_tail, sum3, diff, diff);
```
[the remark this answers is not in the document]
dor-forer: Confirmed, fixed in fc7995d4. This was a real bug and the diagnosis is exactly right.

`svmla_f32_z` zeroes every lane outside the predicate, and by the time the tail runs `sum3` already holds full-vector partial sums (the main loop's fourth step writes it), so those lanes were being discarded. Trigger is `dim >= 4*chunk && dim % chunk != 0`, as described.

Worth recording why it appeared: the pre-restructure kernel used `_z` safely at this spot only because the partial chunk ran *first*, against an all-zero `sum0`. Moving the tail after the main loop removed that precondition. Now using the merging form; the temporaries above stay `_z` since they are fresh values rather than accumulators.

Reproduced and verified on Graviton4 (Neoverse-V2, 128-bit VL so `chunk == 4`): before the fix the parameterized `SQ8_FP32_L2Sqr` sweep failed on precisely the dims where `dim % 4 != 0` and passed on the rest; after it, 593/593 SQ8 tests pass. That was the first execution of these kernels on hardware that actually has SVE -- the previous ARM host was Neoverse-N1, where the runtime `if (optimization.sve)` gates skipped them silently.

tests/unit/test_spaces.cpp
```diff
@@ -409,6 +410,256 @@ TEST_F(SpacesTest, SQ8_FP32_odd_dim_unaligned_metadata_test) {
...
+#ifdef OPT_AVX2_FMA
+    if (optimization.avx2 && optimization.fma3) {
+        check(Choose_SQ8_FP32_L2_implementation_AVX2_FMA(dim)(storage, query, dim), "AVX2_FMA");
+    }
+#endif
+#ifdef OPT_AVX2
+    if (optimization.avx2) {
+        check(Choose_SQ8_FP32_L2_implementation_AVX2(dim)(storage, query, dim), "AVX2");
+    }
+#endif
```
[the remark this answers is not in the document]
dor-forer: Correct, and already fixed in 07a2d9eb (this review ran against a1f7cebb, one commit earlier).

The mechanism was as described: `my_mm256_maskz_loadu_ps` is an unconditional 32-byte `_mm256_loadu_ps`, and the dim=2 repro made the test call the AVX2 tiers directly, bypassing the chooser's `dim < 8` floor -- a 16-byte overread of a 16-byte query blob, which ASAN reported.

The fix gates the x86 tiers on the same floor production uses (`L2_space.cpp`: "Optimizations assume at least 8 elements"). I gated AVX512 and SSE4 too, rather than only the AVX2 pair: their loads are fault-suppressing or per-element so they would not have faulted, but calling any of them below dim 8 exercises a configuration the chooser never produces. aarch64 is deliberately left ungated, since there is no such floor there and the NEON tail uses per-lane loads.

Verified with 595/595 SQ8 tests under ASAN on x86.

src/VecSim/spaces/L2/L2.cpp
```diff
@@ -19,29 +19,38 @@ using float16 = vecsim_types::float16;
...
- *   ||x - y||² = Σx_i² - 2*IP(x, y) + Σy_i²
- *              = x_sum_squares - 2 * IP(x, y) + y_sum_squares
- *   where IP(x, y) = min * y_sum + delta * Σ(q_i * y_i)
+ * Asymmetric SQ8-FP32 L2 squared distance computed via direct residual accumulation:
+ *   ||x - y||² = Σ(dequant(x_i) - y_i)²
+ *   where dequant(x_i) = min_val + delta * q_i
+ *
+ * This avoids the algebraic-identity/cancellation approach (||x||² + ||y||² - 2*IP(x, y)),
+ * which catastrophically cancels in FP32 when x and y share a large common offset relative to
+ * their spread.
```
lerman25: Note self-contained, why mention what doesn't work?

src/VecSim/spaces/L2/L2.cpp
```diff
@@ -19,29 +19,38 @@ using float16 = vecsim_types::float16;
...
 
-    // L2² = ||x||² + ||y||² - 2*IP(x, y)
-    return x_sum_sq + y_sum_sq - 2.0f * ip;
+    const auto *params1 = pVect1 + dimension;
+    const float min_val = load_unaligned<float>(params1 + sq8::MIN_VAL * sizeof(float));
+    const float delta = load_unaligned<float>(params1 + sq8::DELTA * sizeof(float));
+
+    float res = 0;
+    for (size_t i = 0; i < dimension; i++) {
+        // Order matters: min_val - y_i is exact (Sterbenz) since both are large and close, so
```
lerman25: Who says they are large and close?

src/VecSim/spaces/L2/L2_AVX2_FMA_SQ8_FP32.h
```diff
@@ -9,38 +9,119 @@
...
+ *   where dequant(x_i) = min_val + delta * q_i
  *
- * where:
- *   - IP(x, y) = min * y_sum + delta * Σ(q_i * y_i)  (computed via SQ8_FP32_InnerProductImp_FMA)
- *   - x_sum_squares and y_sum_squares are precomputed
+ * This avoids the algebraic-identity/cancellation approach, which catastrophically cancels in
+ * FP32 when x and y share a large common offset relative to their spread.
  *
- * This avoids dequantization in the hot loop.
+ * This version uses FMA instructions. Critically, the subtract is fused into the FMA
```
lerman25: Not self-contined

src/VecSim/spaces/L2/L2_SVE_SQ8_FP32.h
```diff
@@ -21,35 +20,108 @@ namespace {
...
+    // Multiple accumulators for ILP
+    svfloat32_t sum0 = svdup_f32(0.0f);
+    svfloat32_t sum1 = svdup_f32(0.0f);
+    svfloat32_t sum2 = svdup_f32(0.0f);
+    svfloat32_t sum3 = svdup_f32(0.0f);
+
+    // Full vectors first, predicated tail last, bounds compared rather than divided: `chunk` is
+    // runtime (svcntw), so `dimension % chunk` would emit a real udiv, and the chooser already
+    // divided once to produce `partial_chunk`/`additional_steps`. Deliberately unlike the
+    // prefix-first sibling SQ8 SVE kernels; see MOD-17526 for the measured trade-off.
```
lerman25: Not self contained, is the ticket reference needed here?

src/VecSim/spaces/L2/L2.cpp
```diff
@@ -19,29 +19,38 @@ using float16 = vecsim_types::float16;
...
+        // adding the small delta*q_i correction afterward preserves the residual. Computing
+        // (min_val + delta*q_i) - y_i instead rounds it away at the large offset's precision.
+        float diff = delta * static_cast<float>(pVect1[i]) + (min_val - pVect2[i]);
```
lerman25, line 52: Performance follow-up: can we skip the unused FP32 L2 query sums?

The direct residual calculation, `sum((delta * q_i + (min_val - y_i))^2)`, no longer reads `y_sum` or `y_sum_squares`. Those statistics were needed by the previous norm/dot-product identity, but `QuantPreprocessor::preprocessQuery` still computes them. We can remove that extra dimension-wide reduction for FP32 L2, including centered queries, while retaining the query-buffer layout and zeroing the unused slots. FP16, IP/Cosine, and stored-vector metadata should remain unchanged.

POC: [`omer-lerman-sq8-skip-query-sums`](https://github.com/RedisAI/VectorSimilarity/tree/omer-lerman-sq8-skip-query-sums), with the implementation isolated in [f89c5f69](https://github.com/RedisAI/VectorSimilarity/commit/f89c5f6977df8b16c981ed07a0c88c79389c6606).

[CI-only paired benchmarks](https://github.com/RedisAI/VectorSimilarity/actions/runs/34371155907) against this PR's `63a2309` measured nine alternating baseline/candidate rounds on the same runner, for plain and centered queries at dimensions 128, 384, 768, and 1536:

- x86-64: **61.9–93.3% lower preprocessing latency**.
- ARM64: **47.1–70.1% lower preprocessing latency**.
- Example: plain 1536-dimensional query on x86-64, **825.26 ns → 55.39 ns**, saving about **0.77 µs**.

The benchmark includes query-buffer allocation, preprocessing, and free—not HNSW traversal or distance calculations. These are preprocessing improvements, not end-to-end search speedups. The focused spaces, components, and HNSW-SQ8 tests passed on both architectures.

src/VecSim/spaces/L2/L2_AVX512F_BW_VL_VNNI_SQ8_FP32.h
```diff
@@ -8,39 +8,109 @@
...
+    sum = _mm512_fmadd_ps(diff, diff, sum);
+
+    pVec1 += 16;
+    pVec2 += 16;
+}
+
+// pVec1v = SQ8 storage, pVec2v = FP32 query
 template <unsigned char residual> // 0..31
-float SQ8_FP32_L2SqrSIMD16_AVX512F_BW_VL_VNNI(const void *pVect1v, const void *pVect2v,
+float SQ8_FP32_L2SqrSIMD16_AVX512F_BW_VL_VNNI(const void *pVec1v, const void *pVec2v,
```
lerman25, line 51: Performance follow-up: let's move the SQ8-FP32 L2 kernel into the AVX512F-only compilation tier and dispatch on `features.avx512f`.

The direct L2 implementation no longer needs BW/VL/VNNI, but `L2_SQ8_FP32_GetDistFunc` still requires all four extensions. A CPU with AVX512F but without VNNI therefore falls back to AVX2/FMA unnecessarily. The kernel must also move into the translation unit compiled with only `-mavx512f`; relaxing the runtime check while leaving it compiled with the broader instruction flags would not be safe.

POC: [`omer-lerman-sq8-avx512f-dispatch`](https://github.com/RedisAI/VectorSimilarity/tree/omer-lerman-sq8-avx512f-dispatch), implementation commit [156dfb34](https://github.com/RedisAI/VectorSimilarity/commit/156dfb34babb5c6f427f6e2c4f485c85a79fb108).

[CI-only paired benchmarks](https://github.com/RedisAI/VectorSimilarity/actions/runs/34371160673) against this PR's `63a2309`, using nine alternating baseline/candidate rounds on the same runner, measured:

- **9.2–27.8% lower kernel latency** at the tested dimensions 64–1536.
- Dimension 16: essentially unchanged.
- Dimension 8: **4.1% higher latency** (2.75 ns → 2.87 ns).

Given our priority is performance at dimensions above 8, the dimension-8 regression is an acceptable trade-off for the larger-vector gains, not a blocker. Dimensions 9–15 were not benchmarked, so this is not a claim that every dimension above 8 improves.

The benchmark masks VNNI in the dispatch feature structure on a modern Xeon: it compares the AVX2/FMA fallback with the AVX512F path, not actual older CPU hardware or end-to-end search latency. CPUs already selecting AVX512 do not gain just from relaxing the check. All 1,632 focused candidate tests passed in CI, including the AVX512F-only feature-mask dispatch test.

lerman25 (commented):
Nice, some comments

src/VecSim/spaces/L2/L2.cpp
dor-forer: Reworded in ec05d101. The six SIMD headers now state only the formula. The scalar reference keeps one sentence phrased as a property of the direct form (rounding at the scale of |x_i - y_i| rather than the norms), since this is where a future reader would look before changing the accumulation.

src/VecSim/spaces/L2/L2.cpp
dor-forer: Fair. The comment stated a data-dependent precondition as a fact. Rewritten in ec05d101 to give the condition: when the stored range sits far from zero, min_val and y_i are within a factor of two, so the subtraction is exact and rounding stays at the residual's scale. When they are not close the residual is large and either order is fine.

Measured in FP32: at offset 1000, spread 1, max error is 5.9e-8 with this order vs 3.0e-5 with (min + delta*q) - y; at offset 0 both are 5.9e-8.

src/VecSim/spaces/L2/L2_AVX2_FMA_SQ8_FP32.h
dor-forer: Replaced in ec05d101 with one line: "The subtract is fused into the FMA: diff = fma(delta, q, min - y)." Same in the AVX512 header.

src/VecSim/spaces/L2/L2_SVE_SQ8_FP32.h
dor-forer: Ticket references removed in ec05d101 (SVE, NEON, and this line). The trade-off is now inline: faster under SVE2 at every dimension, under plain SVE only when dim is a multiple of VL (about 8% slower otherwise), and SVE2 wins dispatch when both tiers are present.

src/VecSim/spaces/L2/L2.cpp
dor-forer: Agree. Checked every FP32 L2 kernel on both arches: none reads SUM_QUERY or SUM_SQUARES_QUERY, while IP/Cosine and all FP16 kernels still do, which your POC leaves untouched. Keep it out of this PR since #1028 is a correctness fix; open it as a follow-up from your branch once this lands and I will review.

src/VecSim/spaces/L2/L2_AVX512F_BW_VL_VNNI_SQ8_FP32.h
dor-forer: Agree. Confirmed the current kernel compiles with -mavx512f alone across all 32 residual instantiations, with no vpdpbusd in the object, matching the SQ8-FP16 kernels already in AVX512F.cpp. Skylake-SP class hosts (F/BW/VL, no VNNI) currently fall to AVX2-FMA for this path. Separate PR after #1028 for the same reason as above.

lerman25 (approved):
Approved. Keeping the current FP32/SIMD arithmetic; the remaining numerical concern is non-blocking.

Following up the [existing Sterbenz discussion](https://github.com/RedisAI/VectorSimilarity/pull/1028#discussion_r3978205224) with the CI reproduction and an important distinction:

For original vector and query `[1, 33554432, 66846720]`, production SQ8 preprocessing produces `min=1`, `delta=262144`, and codes `[0, 128, 255]`. The mathematical reconstruction `min + delta*q` is `[1, 33554433, 66846721]`. Its squared distance to the query is **2**, while the distance from the **original** vector to that query is **0**. The current kernel returns 0. Thus this example demonstrates a discrepancy against the reconstruction reference used by the PR tests, not worse accuracy against the original vector or demonstrated loss of search recall.

The remaining limitation is that widely separated `min_val` and `y_i` do not imply a large final residual: `delta*q_i` can cancel their difference, after FP32 subtraction has already discarded low bits. Sterbenz therefore does not establish the unconditional accuracy claim in the comment. This is an accepted precision limitation, not a request to change the arithmetic in this PR; any follow-up documentation should distinguish quantization error from arithmetic error.

The exploratory [omer-lerman-sq8-exact-l2 branch](https://github.com/RedisAI/VectorSimilarity/tree/omer-lerman-sq8-exact-l2) passed its [CI verification](https://github.com/RedisAI/VectorSimilarity/actions/runs/34487797901), but its candidate kernel was roughly 19–36.5x slower on x86 and 7–7.5x slower on ARM at the tested dimensions 16–1536. Those are hosted-CI hot-cache kernel measurements, not search latency, and this candidate replaces SIMD arithmetic with shared scalar code. No recall improvement was established, so I am not proposing that implementation or requiring double/exact arithmetic here.

dor-forer:
recheck

PR #1029, opened 2026-09-02
[MOD-14957] Add tiered SQ8 mean training

dor-forer:

Add mean training to tiered SQ8 while keeping the real HNSW backend present from construction. With a positive `QuantNormalizationSetSize`, vectors remain searchable in FLAT until the threshold is reached. Tiered then installs the learned mean in the existing SQ8 components and starts migration.

Stack **3/3**, based on #1035 (`dor-forer-MOD-14957-tiered-sq8-backend`): #1034 → #1035 → #1029. The first two PRs contain the FLAT overwrite fix and immediate-backend SQ8 support.

- The factory creates the final SQ8 layout with an initial zero mean when training is requested. The backend pointer, preprocessing layout, and cached distance-context addresses remain stable.
- Tiered owns an optional accumulation state with a double-precision running sum and a threshold. Overwrites and deletions adjust the sum using FLAT's stored vectors, including cosine normalization. The vector count and threshold decision are read under the FLAT lock.
- Finalization computes the FP32 mean and snapshots pending insertion jobs. Under the exclusive main-index lock, HNSW passes the mean to the preprocessor and distance calculator. The calculator computes the squared-mean correction for IP; L2 needs no such correction. Tiered then clears the accumulation state and releases the lock before executing or submitting jobs.
- Named scoped guards handle exclusive locking in finalization, direct writes, and swap-job execution. The guard helper preserves the cumulative lock-acquisition counter used by tests. External shared locks are released in reverse acquisition order.
- Memory estimates include the mean-enabled backend layout and the temporary running sum. Basic metadata continues to come from the backend.

A zero threshold skips training and preserves a supplied `quantParams` mean, if any. Positive thresholds override a supplied mean, take precedence over `flatBufferLimit`, and are capped at 102400 vectors. Emptying a trained graph does not restart training. SQ8 `getDataByLabel()` returns no vectors in every phase, including when a label is buffered in FLAT.

**Initialization contract and known behavior**

- Mean installation is writer-only: once, before any vector is stored, under the exclusive tiered main lock and before insertion jobs are submitted. Assertions check quantized storage, an empty graph, the mean dimension, and the concrete SQ8 components.
- The setter requires one mean-enabled SQ8 preprocessor in slot 0. Tiered performs cosine normalization in FLAT.
- In write-in-place mode, the threshold-crossing `addVector()` migrates the accumulated set before returning. This can perform up to 102400 insertions in one call.

MOD-14957: tiered SQ8 mean accumulation and initialization.

**Validation**

Validated `fb2c5b0426134c525c4af02c916112197b9c9477` after merging base `4c2f7509`, on `dorer-intel` with GCC 13.3, Debug, and SVS v0.3.2 enabled:

- `test_hnsw`: 158 focused tiered tests passed, including the SQ8 training cases in `test_hnsw_tiered.cpp`.
- `test_hnsw_sq8`: all 133 tests passed.
- `test_svs`: all 4 selected `getDataByLabel` tests passed; this binary includes standalone and tiered SVS retrieval tests.
- All three test targets built successfully with no compiler warnings.

- All 463 tracked-file checksums matched the tested source snapshot. Repository-wide `make check-format` and `git diff --check` passed.
- The conflict resolution preserves the updated base's SVS retrieval implementation; focused standalone and tiered SVS retrieval tests cover the shared header.

Historical full-suite validation at `fa472126`: 3,050 Debug CTest cases passed with 8 expected SVS skips; with `FP64_TESTS=ON`, 3,316 passed with 9 expected skips. Those full suites, RediSearch integration, and ThreadSanitizer were not rerun at this head.

**Mark if applicable**

---

> [!NOTE]
> **High Risk**
> Changes tiered insert/delete locking, deferred SQ8 quantization, and one-shot mean installation on live index components—errors could affect search correctness or concurrency during the training transition; write-in-place can migrate up to the full training batch synchronously.
> 
> **Overview**
> Adds **tiered SQ8 mean training** via new `TieredHNSWParams.QuantNormalizationSetSize`: when positive, inserts stay in the FLAT tier (searchable there) while a double-precision running sum accumulates; reaching the cap (max 102400) finalizes by calling **`HNSWIndex::setQuantizationMean()`** on the already-constructed backend, updating **`QuantPreprocessor::setMean`** and **`DistanceCalculatorWithNorm::setMeanSumSquares`**, then submitting or synchronously running pending HNSW insert jobs. **`QuantNormalizationSetSize = 0`** skips training and keeps prior behavior (optional supplied `quantParams` mean).
> 
> The tiered factory provisions the final SQ8-with-mean layout up front with a **zero mean** so backend pointers and distance context stay stable. Add/delete/overwrite during accumulation maintain the sum from FLAT-stored values (including cosine-normalized inputs); accumulation overrides **`flatBufferLimit`**. Locking moves to scoped **`acquireMainIndexGuard`** / **`acquireSharedIndexDataGuard`** helpers.
> 
> HNSW size estimators gain an explicit **`with_mean`** flag; tests cover deferred mean parity, migration overlap, and concurrent queries across the training transition.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e1e73b58b5f425e11d583b884ced50e0d269eb18. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/VecSim/algorithms/hnsw/hnsw.h
```diff
@@ -251,6 +251,7 @@ class HNSWIndex : public VecSimIndexAbstract<DataType, DistType>,
     void unlockIndexDataGuard() const;
     void lockSharedIndexDataGuard() const;
```
alonre24, line 294: Consider renaming this to exclusive lock

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -1365,7 +1557,7 @@ VecSimDebugInfoIterator *TieredHNSWIndex<DataType, DistType>::debugInfoIterator(
 
 template <typename DataType, typename DistType>
 VecSimIndexBasicInfo TieredHNSWIndex<DataType, DistType>::basicInfo() const {
-    VecSimIndexBasicInfo info = this->backendIndex->getBasicInfo();
+    VecSimIndexBasicInfo info = this->frontendIndex->getBasicInfo();
```
alonre24: Why?

src/VecSim/algorithms/hnsw/hnsw_tiered.h
```diff
@@ -263,22 +316,20 @@ class TieredHNSWIndex : public VecSimTieredIndex<DataType, DistType> {
     }
 
     void releaseSharedLocks() override {
-        this->flatIndexGuard.unlock_shared();
-        this->mainIndexGuard.unlock_shared();
         this->getHNSWIndex()->unlockSharedIndexDataGuard();
+        this->mainIndexGuard.unlock_shared();
```
alonre24, line 320: Any reason to change the order here?

alonre24 (dismissed):
LGTM! Nice and clean

I didn't review the test, asked CLaude if we have full coverage of the index combinations and edge cases and got:


  1. The 102400 cap (MAX_QUANT_NORMALIZATION_SET_SIZE) is never tested. No test sets a threshold at or above it to confirm the std::min(normalizationSetSize, MAX_QUANT_NORMALIZATION_SET_SIZE) clamp
     actually fires. Cheap to add — construct with e.g. 200000 and read the effective threshold back via a friend accessor.
  2. "Large" thresholds are never actually crossed. All finalization tests use thresholds ≤100 (mostly single/double-digit). The one test using 1000 (ConcurrentSearchDuringAccumulation) only adds 50
     vectors, so it never reaches finalization — accumulation is never exercised at real scale. calculateQuantizationMean's division and the "snapshot + bulk-submit N jobs" path in finalizeQuantizationAndSubmitJobs have only ever run with N ≤ ~100.
  3. L2 has noticeably thinner coverage than IP/Cosine. Cosine got a full mirrored suite (9 dedicated tests: accumulation, overwrite, delete, transition+query, transition+range, mean correctness, delete-and-reinsert). L2 only appears in the two 3-metric shared tests plus one dedicated write-in-place-only test — no L2-specific overwrite-during-accumulation, delete-during-accumulation, or  post-transition search-accuracy test, even though L2's mean-centering/FP32-query path is genuinely different code from IP/Cosine.
  4. Write-in-place is only exercised on the generic path, not the interesting cases. Just 5 of the ~50 SQ8-training tests use write-in-place, and none of them are the Cosine-specific tests, the  multi-label accumulation tests, or the migration-overlap dedup test — those are all async-only. A write-in-place + Cosine or write-in-place + multi-value-specific bug wouldn't be caught.
  5. Real multi-threaded concurrency at the exact finalization moment has exactly one test. Everything in test_hnsw_tiered.cpp's training suite uses the deterministic single-step mock  (thread_iteration()), not real OS worker threads. Only test_hnsw_sq8.cpp's ConcurrentQueriesDuringNormalizationTransition uses actual init_threads()-spawned workers racing the job queue      post-finalization — and it doesn't vary type/metric/write-mode.

None is critical for this PR - let's make sure we are stressing and covering in flow/integration tests

src/VecSim/algorithms/hnsw/hnsw_tiered.h
dor-forer: Yes — releaseSharedLocks() now unlocks in the reverse of the order acquireSharedLocks() locks (flat → main → hnsw locked, so hnsw → main → flat unlocked). The old code unlocked in the same order it locked, which wasn't wrong here since these are independent shared locks with no nesting dependency between them, but reverse-order release is the correct general discipline (symmetric with acquisition), so I lined it up while touching this code.

dor-forer:
recheck

alonre24 (dismissed):

dor-forer:
Rebased onto main and fixed the ARM merge-queue failure.

**Why the queue run failed.** In [run 34935460672](https://github.com/RedisAI/VectorSimilarity/actions/runs/34935460672/job/104272274777) (`arm / Test ubuntu-22.04-arm`), `SQ8TieredHNSWTest.Override` and `SQ8TieredHNSWTest.BatchIteratorBasic` for `float16` with a trained mean returned `score = inf` for the ten labels still held in the FP16 flat frontend during mean accumulation. Both tests stored `[i, i, i, i]` up to 250 and queried `[250, 250, 250, 250]`, so the farthest distances reached `4 * 250^2 = 250000`, past the FP16 maximum of 65504. The FP16 kernels accumulate in half precision on ARM, so those distances saturated to infinity, tied, and the farthest labels came back unordered. The x86 FP16 kernels widen to FP32, which is why every other job passed.

**What changed here.** `test(sq8): keep the tiered SQ8 distances inside the FP16 range` scales the values by `0.1`, the same constant `tests/unit/test_fp16.cpp` already uses for this reason, capping the largest distance at 10000. The expected score now comes from the values as they round trip through the storage type (via the existing `to_fp32`) with a relative tolerance, because `0.1 * i` is not exact in FP16.

The kernel defect itself is untouched and predates this PR: [MOD-18628](https://redislabs.atlassian.net/browse/MOD-18628).

**Merge with main.** #1034 and #1035 merged while this PR waited in the queue, so the branch stopped merging cleanly. `hnsw_tiered.h`, `tiered_factory.cpp`, `vec_sim_tiered_index.h` and `test_hnsw_sq8.cpp` took the branch side: on main those four files are byte identical to the head of #1035, which is an ancestor of this branch, so the branch side already holds everything main has for them plus this PR. `test_hnsw_tiered.cpp` took the branch side and then the tests #1045 moved over from the deleted `test_hnsw_parallel.cpp`.

**Validation.** Built the merged tree on an aarch64 Graviton machine (`ubuntu-22.04-arm` equivalent, gcc 12.3): 3056/3056 unit tests pass, including the two that failed in the queue. `check-format.sh` clean. x86 coverage is left to CI.

[MOD-18628]: https://redislabs.atlassian.net/browse/MOD-18628?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

alonre24 (approved):

PR #1033, opened 2026-09-06
Implement SVSIndex::getDataByLabel (MOD-17706)

nonirosenfeldredis:

`SVSIndex::getDataByLabel` was a stub that always reported nothing where `VecSimTieredIndex::getDataByLabel` had a bypass for SVS backend.

It now reads the stored vectors directly from the underlying SVS index (single-value via
`get_datum`, multi-value via `get_label_to_external_lookup` + `get_parent_index`), refusing only for
compressed/LeanVec-reduced storage (`storage_traits_t::is_compressed()`) -- where it reports nothing.

The read logic is shared with the existing test-only `getStoredVectorDataByLabel` via a new private
`appendStoredDataByLabel` helper, so the single/multi-value branching and the raw memcpy live in one
place.

**Tiered read logic**

`VecSimTieredIndex::getDataByLabel` always reads the flat buffer first, then decides whether the
backend also needs reading: for a single-value label the buffer alone is the whole answer once it
has contributed anything, but a multi-value label's vectors are routinely split across the tiers
while an ingest job is pending, so the backend is always consulted too in that case. 

A compressed SVS backend complicates that decision: it appends nothing from `getDataByLabel`
regardless of whether it holds part of the label -- it *cannot* report values, which is not the
same as "doesn't have any". Unlike HNSW, whose tiered backend can never be quantized, SVS's
routinely is, so a multi-value label split between a non-empty buffer contribution and such a
backend would otherwise come back as just the buffer's subset, indistinguishable from the complete
answer. Before trusting that subset, the tiered read checks backend membership via
`SVSIndexBase::isLabelExists` -- cheap because it's metadata, not the value read compression rules
out -- and if the backend is compressed and holds part of the label, discards the buffer's
contribution too and reports nothing for the whole label, the same rule `SVSIndex::getDataByLabel`
already applies within a single tier.

1. [MOD-17706](https://redislabs.atlassian.net/browse/MOD-17706)

1. `src/VecSim/algorithms/svs/svs.h`
2. `src/VecSim/vec_sim_tiered_index.h`
3. `tests/unit/test_svs.cpp`
4. `tests/unit/test_svs_tiered.cpp`

**Mark if applicable**

[MOD-17706]: https://redislabs.atlassian.net/browse/MOD-17706?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> Changes observable `getDataByLabel` behavior for SVS and tiered SVS (including RediSearch “holds vectors” paths); logic for compressed backends and multi-value labels split across tiers is easy to get wrong.
> 
> **Overview**
> **Implements `SVSIndex::getDataByLabel`** so uncompressed SVS indexes return the stored vectors for a label (single-value via `get_datum`, multi-value via label lookup + parent index). **Compressed or LeanVec-reduced indexes still append nothing**, matching HNSW—no dequantized bytes that could fool equality checks.
> 
> The read path is centralized in a new **`appendStoredDataByLabel`** helper, also used by the test-only **`getStoredVectorDataByLabel`**, with compression gated on **`storage_traits_t::is_compressed()`** instead of separate quant flags.
> 
> **Tiered `getDataByLabel` no longer skips all backend reads for SVS.** It always consults the flat buffer first, then the backend when needed (multi-value or label not fully in the buffer). For a **compressed SVS backend**, if the buffer contributed vectors but the backend also holds that label (`isLabelExists`), the tiered read **drops the buffer result and returns nothing** so a partial subset is not mistaken for the full label.
> 
> Unit tests cover uncompressed vs compressed, single vs multi, absent labels, tiered flat-only reads, tiered backend reads after ingest, and the compressed split-label case.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ee9745e66bea0c4ff146f2901c0461c8a385e625. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/VecSim/algorithms/svs/svs.h
+++ b/src/VecSim/algorithms/svs/svs.h
@@ -746,9 +746,9 @@ public:
     }
 
 private:
-    // appends the raw stored elements for `label` to `vectors_output`, one entry per stored vector -- zero if
-    // the label isn't held. Only meaningful when the caller has already excluded compressed
-    // storage; nothing here dequantizes.
+    // appends the raw stored elements for `label` to `vectors_output`, one entry per stored vector
+    // -- zero if the label isn't held. Only meaningful when the caller has already excluded
+    // compressed storage; nothing here dequantizes.
     template <typename OutputElement>
     void appendStoredDataByLabel(labelType label,
                                  std::vector<std::vector<OutputElement>> &vectors_output) const {
@@ -756,8 +756,7 @@ private:
             return;
         }
         auto append_datum = [&](auto indexed_span) {
-            std::vector<OutputElement> vec_data(this->getStoredDataSize() /
-                                                sizeof(OutputElement));
+            std::vector<OutputElement> vec_data(this->getStoredDataSize() / sizeof(OutputElement));
             const char *data_ptr = reinterpret_cast<const char *>(indexed_span.data());
             std::memcpy(vec_data.data(), data_ptr, this->getStoredDataSize());
             vectors_output.push_back(std::move(vec_data));
--- a/src/VecSim/algorithms/svs/svs.h
+++ b/src/VecSim/algorithms/svs/svs.h
@@ -16,6 +16,7 @@
 #include <memory>
 #include <cassert>
 #include <limits>
+#include <type_traits>
 #include <vector>
 
 #include "svs/index/vamana/dynamic_index.h"
@@ -756,10 +757,19 @@ private:
             return;
         }
         auto append_datum = [&](auto indexed_span) {
-            std::vector<OutputElement> vec_data(this->getStoredDataSize() / sizeof(OutputElement));
-            const char *data_ptr = reinterpret_cast<const char *>(indexed_span.data());
-            std::memcpy(vec_data.data(), data_ptr, this->getStoredDataSize());
-            vectors_output.push_back(std::move(vec_data));
+            if constexpr (std::is_same_v<OutputElement, DataType>) {
+                // The span's element type already is `OutputElement` here, so build the
+                // output vector directly from it instead of a raw byte copy.
+                vectors_output.emplace_back(indexed_span.begin(), indexed_span.end());
+            } else {
+                // FP16 and the test-only byte output: `OutputElement` differs from the
+                // span's element type but is bit-identical size, so copy the raw bytes.
+                std::vector<OutputElement> vec_data(this->getStoredDataSize() /
+                                                    sizeof(OutputElement));
+                const char *data_ptr = reinterpret_cast<const char *>(indexed_span.data());
+                std::memcpy(vec_data.data(), data_ptr, this->getStoredDataSize());
+                vectors_output.push_back(std::move(vec_data));
+            }
         };
 
         if constexpr (isMulti) {
--- a/src/VecSim/algorithms/svs/svs.h
+++ b/src/VecSim/algorithms/svs/svs.h
@@ -745,6 +745,41 @@ public:
         num_marked_deleted = 0;
     }
 
+private:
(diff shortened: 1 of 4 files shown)


src/VecSim/vec_sim_tiered_index.h
```diff
@@ -147,37 +154,26 @@ class VecSimTieredIndex : public VecSimIndexInterface {
...
-        // A quantized backend cannot report its stored vectors as values -- the stored form is
-        // compression plus metadata, and nothing here dequantizes -- so it would append nothing.
-        bool backend_can_report = true;
+        std::shared_lock<std::shared_mutex> flat_lock(this->flatIndexGuard);
+        const size_t before_flat = vectors_output.size();
+        this->frontendIndex->getDataByLabel(label, vectors_output);
+        // Whether the buffer held it, measured rather than read off emptiness, so the tier
+        // decision does not depend on an assertion that only exists in test builds.
+        if (this->backendIndex->isMultiValue() || vectors_output.size() == before_flat) {
+            std::shared_lock<std::shared_mutex> main_lock(this->mainIndexGuard);
```
dor-forer, line 160: For a compressed SVS backend with no buffer contribution, this acquires mainIndexGuard even though the result must be empty. It can wait behind a batch’s exclusive addVectors lock, then immediately return nothing from SVSIndex::getDataByLabel. While waiting, it also holds flatIndexGuard shared, potentially delaying buffer writes.
Could we resolve svs_backend before acquiring the main lock and return early when isCompressed() && vectors_output.size() == before_flat?
The multi-value case with a buffer contribution should retain the locked isLabelExists check, since we still need to determine whether the buffer represents the complete label.

src/VecSim/algorithms/svs/svs.h
```diff
@@ -745,6 +745,40 @@ class SVSIndex : public VecSimIndexAbstract<svs_details::vecsim_dt<DataType>, fl
...
+    template <typename OutputElement>
+    void appendStoredDataByLabel(labelType label,
+                                 std::vector<std::vector<OutputElement>> &vectors_output) const {
+        if (!impl_) {
+            return;
+        }
+        auto append_datum = [&](auto indexed_span) {
+            std::vector<OutputElement> vec_data(this->getStoredDataSize() / sizeof(OutputElement));
+            const char *data_ptr = reinterpret_cast<const char *>(indexed_span.data());
+            std::memcpy(vec_data.data(), data_ptr, this->getStoredDataSize());
```
dor-forer: Consider adding a same type optimization (if `OutputElement` and `DataType`),
We can create `vec_data` directly in `vectors_output`.
Something like:
```
auto append_datum = [&](auto indexed_span) {
    if constexpr (std::is_same_v<OutputElement, DataType>) {
        vectors_output.emplace_back(indexed_span.begin(), indexed_span.end());
    } else {
        // Existing implementation, preserved for FP16 and byte output.
        std::vector<OutputElement> vec_data(
            this->getStoredDataSize() / sizeof(OutputElement));
        std::memcpy(vec_data.data(), indexed_span.data(), this->getStoredDataSize());
        vectors_output.push_back(std::move(vec_data));
    }
};
```

nonirosenfeldredis: done

src/VecSim/vec_sim_tiered_index.h
nonirosenfeldredis: done

dor-forer (approved):

PR #1045, opened 2026-09-09
[MOD-17706] finalize SVSIndex::relabelVector

nonirosenfeldredis:

SVS delegates label management to its library, so `relabelVector` was reported as unsupported. The library now offers `replace_external_id`, which renames an external id (single) or a label (multi) purely in the id translation table — leaving the dataset and graph untouched, which is the same guarantee the in-tree indexes give. This bumps the submodule to `a7e3494` and routes `relabelVector` through it.

Preconditions are checked in `svs.h` rather than left to the library: it throws on a bad one, and this API answers with a code. Checking `has_id` for both labels first also means the throwing validation inside `replace_external_id` cannot fire, so no exception escapes into the C API. `markIndexUpdate` is deliberately not called — nothing was added or deleted and the label count is unchanged, so no consolidation is owed.

**Availability gate.** `SVS_SHARED_LIB` (Linux x86_64) downloads a **pre-built SVS release** rather than building the submodule, and `replace_external_id` is in none of v0.3.0 / v0.3.1 / v0.3.2 — nor is v0.3.2 an ancestor of SVS `main`. So availability depends on which SVS a build picked up, not on the platform. `cmake/svs.cmake` now detects it following the pattern already used for LVQ, with one necessary difference: LVQ tests for a *header file's existence*, but this is a *method*, so the check greps that header's contents. Where it is missing, the override is left out entirely and the interface default reports `VecSimRelabel_Unsupported` — telling a caller to fall back to delete + insert rather than read it as a no-op.

**Tiered SVS.** The tier needs its own override: the label can sit in the flat buffer, in the backend, or -- for multi -- in both, and only the backend can refuse. So the backend moves first and the buffer follows, all-or-nothing, with `new_label` required free in *both* tiers rather than only the one holding `old_label`.

Two things make it correct rather than merely serialised:

- `updateJobMutex` is held across the whole operation. An update job snapshots the buffer's labels **by value** and afterwards reconciles only id swaps and deletions, so a rename landing inside its window would be invisible to it and the vector would reach the backend under the old label.
- The three guards are taken **together** via `std::scoped_lock`, not in sequence. The orders in this class disagree: nearly everything acquires flat before main, but the in-place add path takes `mainIndexGuard` before `flatIndexGuard` while the backend is still empty. Acquiring in sequence would hold one guard while blocking on another and close a cycle with that path. Same idiom that path already uses for its own two locks.

The override is declared under the same `HAVE_SVS_REPLACE_EXTERNAL_ID` gate as `SVSIndex::relabelVector`, so a build without the API reports `Unsupported` for the whole tier. Left ungated it would move a buffered label and refuse an ingested one, which a caller cannot act on -- and it would have made the runtime probe lie, since an override answering `SameLabel` before consulting the backend looks capable on a build that is not.

**Python bindings.** Two changes, both prerequisites for testing any of this from python:

- `relabel_vector` had no binding, so no flow test could reach it. Bound on `PyVecSimIndex` (serves every index type), with `VecSimRelabelCode` exposed so callers can tell the rejections apart.
- `get_vector` was bound, but its dispatch treated "not brute force" as "is HNSW". `SVSIndex` derives from `VecSimIndexAbstract`, not `HNSWIndex`, so the `dynamic_cast` returned `nullptr` and the virtual call dereferenced it — `svs_index.get_vector(label)` was a **segfault**, which is why the SVS `getDataByLabel` merged in #1033 was unreachable from python. Since `getDataByLabel` is a pure virtual on `VecSimIndexAbstract`, the algo switch was never needed: call it virtually and BF, HNSW and SVS are all served. Tiered keeps its own branch because it wraps two `VecSimIndexAbstract` instances rather than being one.

1. MOD-17706

1. `SVSIndex::relabelVector` — via `impl_->replace_external_id`, gated on availability
2. `SVSTieredIndex::relabelVector` — cross-tier move under `updateJobMutex` + both tier guards, same gate
3. `cmake/svs.cmake` — `HAVE_SVS_REPLACE_EXTERNAL_ID` detection
4. `PyVecSimIndex` — `relabel_vector` binding, `VecSimRelabelCode` enum, `getDataByLabel` dispatch fix
5. Unit + flow tests

**Testing**

Unit tests are gated the same way as the implementation, asserting real behavior where the API exists and the unsupported code where it does not, so the contract stays covered in both configurations. Flow tests cover both APIs, **single and multi**, across every index type.

Multi is not a formality: a label owns several vectors, each backend moves them by a different route, and a move onto an occupied label must be rejected because accepting it would silently *merge* two labels' vectors rather than lose one.

The tiered test relabels an early label and the last-inserted one. Workers ingest in insertion order, so the early one is already in HNSW while the late one is still buffered with pending ingest jobs — measured at ~690 of 1000 still buffered at relabel time — so both tiers are covered. The buffered case is the delicate one: a job left holding the old label would either ingest under it or throw out of a worker thread.

Tiered SVS gets six unit tests, covering both write states, per-tier rejections, relabel after a delete on either tier, and the two concurrency hazards the locking exists for: `relabelVectorDoesNotDeadlockAgainstInPlaceAdd` runs a relabeler against the main-then-flat add path, and `relabelVectorCannotLandInsideAnUpdateJobsWindow` fires a relabel while an update job holds its snapshot. Both were mutation-verified -- reverting the locking makes them hang or ingest under the stale label rather than pass.

**Both sides of the gate are verified**, using real SVS checkouts rather than a forced macro:

| Submodule | cmake detection | Build | Relabel test present |
|---|---|---|---|
| `a7e3494` (has API) | `replace_external_id found - SVS relabeling enabled`, macro `=1` | clean | `relabelVector`, `relabelVectorRejects` — pass |
| `7786d43` (v0.3.0, no API) | `replace_external_id not found - reports unsupported`, macro `=0` | clean, 0 errors | `relabelVectorUnsupported` — passes |

So the same source tree configures and builds correctly against either SVS, and a build without the API reports `VecSimRelabel_Unsupported` rather than failing to compile.

**Mark if applicable**

**Follow-up, not blocking this PR**

SVS relabeling stays `Unsupported` on Linux x86_64 until an SVS release carries [intel/ScalableVectorSearch#383](https://github.com/intel/ScalableVectorSearch/pull/383) and `SVS_URL` in `cmake/svs.cmake` is bumped to it. That is an external dependency; the gate makes the code correct and compiling in the meantime, which is what the table above verifies.

**Also here: the plain-index parallel suites are removed**

`tests/unit/test_hnsw_parallel.cpp` and `tests/flow/test_hnsw_parallel.py` (1118 lines) drove a non-tiered HNSW index from several threads. The tiered suites cover the same concurrent paths through the job queue, so these no longer earn their runtime. Unrelated to relabel and its own commit, `492ae88`, so it can be reviewed on its own.

One consequence worth naming: `knn_parallel` and `range_parallel` in the python bindings now have no test coverage at all. `add_vector_parallel` keeps its, via `test_svs.py`.

**Notes for the reviewer**

- The tiered half of the `get_vector` dispatch fix is a **pre-existing** bug unrelated to SVS relabel — it fell out of doing the fix properly rather than special-casing SVS.
- The new `add_compile_definitions(VectorSimilarity PUBLIC ...)` call copies the existing LVQ block verbatim for consistency. Worth knowing that form also defines bare `VectorSimilarity` and `PUBLIC` macros (visible in `COMPILE_DEFS`), since `add_compile_definitions` takes definitions rather than a target — pre-existing, and I did not change it.
- Two unrelated local-environment snags, not addressed here: `make flow_test` cannot configure with CMake 4.x because pybind11 is pinned at v2.10.1, and `poetry install` needs `setuptools` in the venv on Python 3.13+.

---

> [!NOTE]
> **Medium Risk**
> Tiered relabel adds new cross-tier locking and update-job synchronization paths where a mistake could deadlock or ingest under stale labels; SVS availability is split by build artifact rather than platform.
> 
> **Overview**
> Adds **SVS (and tiered SVS) support for `relabelVector`**, renaming external labels in place when the linked SVS library exposes `replace_external_id`, instead of always reporting `VecSimRelabel_Unsupported`.
> 
> **Build:** `cmake/svs.cmake` greps `dynamic_index.h` and defines **`HAVE_SVS_REPLACE_EXTERNAL_ID`** so pre-built SVS binaries without the API still compile and surface unsupported at runtime.
> 
> **Implementation:** `SVSIndex::relabelVector` validates labels then calls `impl_->replace_external_id`. **`SVSTieredIndex`** coordinates flat buffer and backend with **`updateJobMutex`** plus **`std::scoped_lock`** on both tier guards (backend first, all-or-nothing rejections, new label free in both tiers).
> 
> **Python:** Exposes **`relabel_vector`** and **`VecSimRelabelCode`**. Fixes **`get_vector`** to use **`VecSimIndexAbstract::getDataByLabel`** (and tiered when applicable), avoiding HNSW-only casts that could crash on SVS.
> 
> **Tests:** Flow/unit coverage for relabel on flat, HNSW, tiered HNSW/SVS (single/multi, quantized variants where applicable), SVS **`get_vector`**, and runtime **`svs_relabel_supported`**. Removes the standalone **HNSW parallel** flow/unit test targets/files.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 492ae880ecdb8ac55c30971121a280ceeec56b11. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/VecSim/algorithms/svs/svs.h
+++ b/src/VecSim/algorithms/svs/svs.h
@@ -549,6 +549,29 @@ public:
         return deleteVectorsImpl(labels, n);
     }
 
+#if HAVE_SVS_REPLACE_EXTERNAL_ID
+    // Only declared when the SVS this was built against offers `replace_external_id`. The
+    // pre-built SVS releases predate it, so where it is missing this override is left out and the
+    // interface default reports `VecSimRelabel_Unsupported` - which tells a caller to fall back to
+    // delete + insert rather than read it as a no-op.
+    VecSimRelabelCode relabelVector(labelType old_label, labelType new_label) override {
+        if (old_label == new_label) {
+            return VecSimRelabel_SameLabel;
+        }
+        // `impl_` is only created on the first insertion, so an index that never held a vector
+        // trivially has nothing under `old_label`.
+        if (!impl_ || !impl_->has_id(old_label)) {
+            return VecSimRelabel_OldLabelMissing;
+        }
+        if (impl_->has_id(new_label)) {
+            return VecSimRelabel_NewLabelTaken;
+        }
+
+        impl_->replace_external_id(old_label, new_label);
+        return VecSimRelabel_OK;
+    }
+#endif // HAVE_SVS_REPLACE_EXTERNAL_ID
+
     bool isLabelExists(labelType label) const override {
         return impl_ ? impl_->has_id(label) : false;
     }
--- a/src/VecSim/algorithms/svs/svs.h
+++ b/src/VecSim/algorithms/svs/svs.h
@@ -558,12 +558,12 @@ public:
         if (old_label == new_label) {
             return VecSimRelabel_SameLabel;
         }
-        // `impl_` is only created on the first insertion, so an index that never held a vector
-        // trivially has nothing under `old_label`.
-        if (!impl_ || !impl_->has_id(old_label)) {
+        // `isLabelExists` also covers the index that never held a vector, where `impl_` has not
+        // been created yet and so trivially holds nothing.
+        if (!isLabelExists(old_label)) {
             return VecSimRelabel_OldLabelMissing;
         }
-        if (impl_->has_id(new_label)) {
+        if (isLabelExists(new_label)) {
             return VecSimRelabel_NewLabelTaken;
         }
 
--- a/src/python_bindings/bindings.cpp
+++ b/src/python_bindings/bindings.cpp
@@ -221,6 +221,11 @@ public:
     // the iterator lives, and `std::shared_mutex` is neither recursive nor upgradeable -- a
     // relabel issued while an iterator is alive would block on a lock its own thread holds.
     VecSimRelabelCode relabelVector(labelType old_label, labelType new_label) {
+        // Released for the same reason `addVector` and `knn` release it: on a tiered index this
+        // blocks on `mainIndexGuard`, which a live batch iterator holds shared until it is
+        // depleted or freed. Holding the GIL while blocking there would stop the Python thread
+        // that owns the iterator from ever running to release it.
+        py::gil_scoped_release py_gil;
         return VecSimIndex_RelabelVector(index.get(), old_label, new_label);
     }
 
--- a/src/python_bindings/bindings.cpp
+++ b/src/python_bindings/bindings.cpp
@@ -215,16 +215,7 @@ public:
 
     void runGC() { VecSimTieredIndex_GC(index.get()); }
 
-    // Deliberately unguarded, like `addVector` and `deleteVector` beside it: this binding leaves
-    // serialisation to the caller. Taking `indexGuard` here instead would deadlock, because
-    // `PyHNSWLibIndex::createBatchIterator` holds that same mutex in shared mode for as long as
-    // the iterator lives, and `std::shared_mutex` is neither recursive nor upgradeable -- a
-    // relabel issued while an iterator is alive would block on a lock its own thread holds.
     VecSimRelabelCode relabelVector(labelType old_label, labelType new_label) {
-        // Released for the same reason `addVector` and `knn` release it: on a tiered index this
-        // blocks on `mainIndexGuard`, which a live batch iterator holds shared until it is
-        // depleted or freed. Holding the GIL while blocking there would stop the Python thread
-        // that owns the iterator from ever running to release it.
         py::gil_scoped_release py_gil;
         return VecSimIndex_RelabelVector(index.get(), old_label, new_label);
     }
--- a/src/python_bindings/bindings.cpp
+++ b/src/python_bindings/bindings.cpp
@@ -106,12 +106,11 @@ private:
     template <typename DataType, typename DistType, typename NPArrayType = DataType>
     inline py::object rawVectorsAsNumpy(labelType label, size_t dim) {
         std::vector<std::vector<DataType>> vectors;
-        if (index->basicInfo().algo == VecSimAlgo_BF) {
-            dynamic_cast<BruteForceIndex<DataType, DistType> *>(this->index.get())
-                ->getDataByLabel(label, vectors);
+        if (auto *tiered =
+                dynamic_cast<VecSimTieredIndex<DataType, DistType> *>(this->index.get())) {
+            tiered->getDataByLabel(label, vectors);
         } else {
-            // index is HNSW
-            dynamic_cast<HNSWIndex<DataType, DistType> *>(this->index.get())
+            dynamic_cast<VecSimIndexAbstract<DataType, DistType> *>(this->index.get())
                 ->getDataByLabel(label, vectors);
         }
         size_t n_vectors = vectors.size();
@@ -215,6 +214,10 @@ public:
     }
 
     void runGC() { VecSimTieredIndex_GC(index.get()); }
+
+    VecSimRelabelCode relabelVector(labelType old_label, labelType new_label) {
+        return VecSimIndex_RelabelVector(index.get(), old_label, new_label);
+    }
 
     py::object getVector(labelType label) {
         VecSimIndexBasicInfo info = index->basicInfo();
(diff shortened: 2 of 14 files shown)


nonirosenfeldredis:
## Verification update

Two of the three open items are now closed.

**Unit suite re-run after the gating change** — 2788/2789. The single failure is `UtilsTests.testMockThreadPool`, which fails only inside my sandbox (gtest hardcodes `/tmp` for stream capture) and passes outside it. Unrelated to this change.

**The `HAVE_SVS_REPLACE_EXTERNAL_ID=0` path is verified**, which is the configuration Linux x86_64 CI will be in. Rather than forcing the macro, I checked the submodule back to `7786d43` (v0.3.0, which genuinely lacks `remap_external_id`) so this exercises the real path:

| Check | Result |
|---|---|
| cmake detects absence | `SVS replace_external_id not found - SVS relabeling reports unsupported`, macro `=0` |
| Compiles against API-less SVS | 0 errors |
| Which test exists | only `SVSTest.relabelVectorUnsupported`, and it passes |

So the same source tree configures and builds correctly against either SVS, and CI will report `VecSimRelabel_Unsupported` rather than failing to compile. The detection message also appears in the python bindings build, since `src/python_bindings/CMakeLists.txt` includes `cmake/svs.cmake` — worth stating because if the define had not reached that build, the SVS flow tests would have silently skipped rather than run.

Confirmed they do run rather than skip: all 6 SVS `get_vector`/`relabel_vector` flow tests report PASSED, none SKIPPED, against a module rebuilt with the gate in place.

**Still open:** the full flow suite across all four files is mid-run, and the SVS release dependency (an SVS release carrying #383 plus an `SVS_URL` bump) remains external to this PR.

src/VecSim/algorithms/svs/svs.h
```diff
@@ -549,6 +549,29 @@ class SVSIndex : public VecSimIndexAbstract<svs_details::vecsim_dt<DataType>, fl
...
+    // pre-built SVS releases predate it, so where it is missing this override is left out and the
+    // interface default reports `VecSimRelabel_Unsupported` - which tells a caller to fall back to
+    // delete + insert rather than read it as a no-op.
+    VecSimRelabelCode relabelVector(labelType old_label, labelType new_label) override {
+        if (old_label == new_label) {
+            return VecSimRelabel_SameLabel;
+        }
+        // `impl_` is only created on the first insertion, so an index that never held a vector
+        // trivially has nothing under `old_label`.
+        if (!impl_ || !impl_->has_id(old_label)) {
```
dor-forer: CAn you use the SVSIndex's `isLabelExists()`?

tests/flow/test_hnsw_tiered.py
```diff
@@ -621,3 +621,78 @@ def test_multi_range_query(test_logger):
...
+    # with a pending ingest job - between them the two tiers both get covered. The buffered case is
+    # the delicate one: a job left holding the old label would either ingest the vector under it or
+    # throw out of the worker thread, and neither would survive the assertions below.
+    buffered = index.get_curr_bf_size()
+    test_logger.info(f"relabeling with {buffered} of {num_elements} vectors still buffered")
+    moved = {7: num_elements + 500, num_elements - 1: num_elements + 501}
+    for old_label, new_label in moved.items():
+        assert index.relabel_vector(old_label, new_label) == VecSimRelabel_OK
+
+    index.wait_for_index()
```
dor-forer, line 649: Could we also test `get_vector` while a vector is still buffered, and for a multi-value label whose vectors are split across both tiers?
The current tiered tests call it only after `wait_for_index()`, so they exercise retrieval after ingestion completes. Controlling the ingestion workers would make those additional states deterministic. The assertion should verify that every expected vector is returned exactly once.

src/python_bindings/bindings.cpp
```diff
@@ -216,6 +215,10 @@ class PyVecSimIndex {
 
     void runGC() { VecSimTieredIndex_GC(index.get()); }
 
+    VecSimRelabelCode relabelVector(labelType old_label, labelType new_label) {
+        return VecSimIndex_RelabelVector(index.get(), old_label, new_label);
```
dor-forer, line 220: Do you think we need to find a way to synchronize the new `relabel_vector` binding with concurrent HNSW queries? knn_parallel releases the GIL and its workers hold `indexGuard`, but this call bypasses that guard. Relabeling writes `idToMetaData[id].label` while queries can read it through `getExternalLabel`, without a common lock.

tests/flow/test_hnsw_tiered.py
nonirosenfeldredis: tested it in C++ now

src/python_bindings/bindings.cpp
nonirosenfeldredis: decided to remove completly the parael tests on plain indexes. we have the concurrent tests on tiered instead

dor-forer (approved):

PR #1047, opened 2026-09-10
[MOD-18494] Stop a relabel mid-query from reporting one vector twice

nonirosenfeldredis:

A tiered query reads the flat buffer, releases `flatIndexGuard`, then reads the main index, and `merge_result_lists` collapses a vector that both tiers report by matching **labels**. A relabel landing in that window defeats the match: the flat half of the pair still says `old_label` while the main half now says `new_label`, so the merge keeps both and the caller gets **one vector twice**, under a pair of labels the index never held at the same time.

The label-keyed merge is load-bearing for the migration race it was written for — `TopKQuery_SameVectorDifferentScores_DueToQuantization` pins that — and it works there only because migration *preserves* the label. `relabelVector` (#1017) moves the very key the merge rests on.

**Replay, not just detection.** `recordRelabel` appends `{old, new, epoch}` under the guards that made the move. A query samples the epoch while the flat guard still covers its snapshot — so the sample names exactly the state its results came from — and after the main read replays every recorded move newer than that sample onto both lists. A flat result under a label that has since moved is rewritten to the label it moved to, and the label-keyed merge collapses it as it always did.

Details that matter:
- **Chains are followed.** A label that moved twice in one window (`A→B→C`) must land on `C`; stopping at `B` matches nothing in the main results and duplicates just the same. The walk is bounded by the number of moves, which also stops a cycle (`A→B` then `B→A`) spinning.
- **Re-sorted before merging.** Rewriting labels can break the by-id tie-break `merge_result_lists` assumes.
- **Fixed ring, so it never allocates and never grows.** Past capacity the oldest moves are dropped, and a query whose snapshot is no longer covered falls back to a pinned re-read (flat guard held across both reads, which excludes relabel outright). The epoch is the version stamp; the pin is the backstop.

**Why record rather than only detect.** A reader that cannot restart can still consult the log. That is the route to covering `TieredHNSW_BatchIterator`, which merges the tiers incrementally across many calls and which redo-on-detect cannot help. **Not wired up here** — the mechanism now exists for it.

**Rejected: keying the log off the flat index and reading it with the flat snapshot.** Both halves fail. At flat-read time the relabel has not happened yet, so the list comes back empty — the query needs to know about a move in its *future*. And a relabel touching only the main index (reachable when the snapshot predates migration) would never be recorded by a flat-scoped log at all.

**Rejected alternative.** Validating flat-sourced results against current label state after the main read also fixes this, but it additionally drops stale results for concurrently *deleted* labels — a behaviour change well outside relabel's scope.

**Not covered.** `TieredHNSW_BatchIterator` merges the tiers incrementally across many calls, so redo-on-detect does not map onto it; a relabel concurrent with a live batch iterator can still duplicate. Flagging rather than papering over it.

**Tests.** Two tests drive a relabel into the window via a new `BUILD_TESTS` hook, one for topK and one for range. Each asserts the vector is reported once *and* that the retry actually fired, so they fail loudly rather than passing on the fast path if the window moves. The hook deliberately does not fire on the pinned retry — there is no window there, and injecting one would self-deadlock against the guard the querying thread holds.

Verified by mutation, twice. With the replay disabled, all three duplicate tests report `Which is: 2` — the bug. With the chain walk cut to a single step, the chain test alone fails while the other two pass, so each test discriminates rather than merely passing.

Full suites green: `test_hnsw` 191/191, `test_hnsw_parallel` 4/4, `test_hnsw_sq8` 55/55, `test_svs` 376 passed / 101 skipped (pre-existing SVS gating) / 0 failed. `make check-format` clean.

1. MOD-18494
2. Found while reviewing the disk relabel implementation in RediSearchEnterprise#967 — the same defect, since vecsim_disk reuses this merge. Fixing the RAM path first; the disk index inherits the query shape.

1. `VecSimTieredIndex` — `relabelEpoch` + `bumpRelabelEpoch()`; `topKQuery`/`rangeQuery` sample the epoch and redo once; `topKQueryImp`/`rangeQueryImp` take `pin_flat`; `FlatGuardRelease` releases the pinned guard on the early-return paths.
2. `TieredHNSWIndex::relabelVector` — bumps the epoch under both guards.
3. `tests/unit/test_hnsw_tiered.cpp` — `relabelDuringQueryDoesNotDuplicate`, `relabelDuringRangeQueryDoesNotDuplicate`.

**Mark if applicable**

### DIFF
--- a/src/VecSim/algorithms/hnsw/hnsw_tiered.h
+++ b/src/VecSim/algorithms/hnsw/hnsw_tiered.h
@@ -1122,16 +1122,26 @@ VecSimQueryReply *TieredHNSWIndex<DataType, DistType>::TieredHNSW_BatchIterator:
         // the distances and access the BF index. We take the lock on this call.
         this->index->flatIndexGuard.lock_shared();
         auto cur_flat_results = this->flat_iterator->getNextResults(n_res, BY_SCORE_THEN_ID);
-        this->index->flatIndexGuard.unlock_shared();
         // This is also the only time `getNextResults` on the BF iterator can fail.
         if (VecSim_OK != cur_flat_results->code) {
+            this->index->flatIndexGuard.unlock_shared();
             return cur_flat_results;
         }
         this->flat_results.swap(cur_flat_results->results);
         VecSimQueryReply_Free(cur_flat_results);
+#ifdef BUILD_TESTS
+        if (this->index->testHookBetweenFlatAndMainRead) {
+            this->index->testHookBetweenFlatAndMainRead();
+        }
+#endif
         // We also take the lock on the main index on the first call to getNextResults, and we hold
-        // it until the iterator is depleted or freed.
+        // it until the iterator is depleted or freed. That is what keeps a relabel out for the
+        // rest of the iteration -- it needs this guard exclusively -- so the only window it could
+        // use is between the two acquisitions here. Taking the main guard *before* releasing the
+        // flat one closes it: a relabel needs both, and cannot have either. The flat->main order
+        // is the same one every other acquisition uses, so overlapping them cannot deadlock.
         this->index->mainIndexGuard.lock_shared();
+        this->index->flatIndexGuard.unlock_shared();
         this->hnsw_iterator = this->index->backendIndex->newBatchIterator(
             this->flat_iterator->getQueryBlob(), queryParams);
         auto cur_hnsw_results = this->hnsw_iterator->getNextResults(n_res, BY_SCORE_THEN_ID);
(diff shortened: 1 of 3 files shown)


nonirosenfeldredis:
Closing this without merging.

The duplicate is real — a cross-tier merge keyed on the label reports one vector twice when a relabel lands between the flat snapshot and the backend read — but it is the same exposure the delete-and-add path it replaces already has, so the relabel epoch and the replay log here buy correctness the surrounding code does not otherwise provide. Not worth their weight.

The work is preserved on `sharon-tiered-relabel-query-race` (3 commits) if the trade-off is revisited. Tracked in MOD-18494.

Worth knowing what closing this also drops: the last commit, *Close the batch iterator's relabel window by overlapping its guards*, is independent of the log. It takes `mainIndexGuard.lock_shared()` before releasing `flatIndexGuard` in `hnsw_tiered.h`'s batch iterator, which is a three-line fix for a window the iterator has regardless of how the merge dedups. That one could land on its own.

PR #1054, opened 2026-09-16
[MOD-17704] Add an updateVectors API and implement it for HNSW, flat and SVS

nonirosenfeldredis:
## What

```c
VecSimUpdateCode updateVectors(labelType label, const void *new_blobs, size_t n)
// VecSimUpdate_OK | VecSimUpdate_MultiNotSupported | VecSimUpdate_Unsupported | VecSimUpdate_Failed
```

Sets the vectors stored under `label` to the given ones: whatever it holds now is removed and these
take its place. This is what overwriting through `addVector` does, generalized to a label holding
several vectors - where `addVector` cannot express it, since in a multi-value index it appends. A
caller doing it by hand has to delete the label and then insert, which also means knowing that a
delete is needed at all.

- How many vectors the label ends up with is decided by the update, not by what was there.
- `n == 0` leaves the label holding nothing, i.e. `deleteVector`.
- A label the index does not hold yet is simply stored.
- A single-value index accepts only `n == 1` and reports `MultiNotSupported` otherwise - refused
  *before* anything is removed, rather than storing one vector and dropping the rest. Single vs
  multi is fixed when the index is created (it selects the single or multi class), so this is a
  property of the index the caller built, not of the call.
- `Failed` says the index tried and could not store them. Unlike the other two refusals it does
  *not* leave the index untouched: the removal precedes the insertion, so the label may hold
  nothing and the caller has to re-insert to reach a known state.

Follows `relabelVector`: an interface default reporting `Unsupported`, a C API
(`VecSimIndex_UpdateVectors`) and a python binding (`update_vectors`).

## Implementations

**One vector into a single-value label is an overwrite**, and every implementation says so by
delegating to `addVector` for that case - which matters because it is the case a caller hits most,
a hash document holding one vector per field. In the tiered indexes `addVector` serves it better
than a delete and an insert would: it writes over the buffered copy in place, keeping its id, and
only then marks the backend's copy deleted, so the label is never without a vector and no other
label's buffered vector is moved as collateral. The general path below is what runs for everything
else.

**HNSW and flat.** For a graph index this is a removal followed by insertions rather than a write
over the stored data: an element's neighbours, and the elements that chose it as one, were selected
from its old value, so overwriting the bytes would leave the graph describing vectors that are no
longer there. A flat index derives nothing from the values, so there a single-value label is
written over in place - it keeps its id, and the rest of the arrays are untouched.

**Tiered HNSW.** Both halves go through the index's own delete and add, which is what makes every
tier and write mode fall out for free: the delete drops the label's buffered copies and invalidates
their pending insert jobs, and marks its indexed elements deleted (or removes them in place in
`VecSim_WriteInPlace`); each insert then lands wherever a fresh vector would - the flat buffer, or
HNSW directly when that buffer is full. An indexed element leaves a tombstone for the usual repair
and swap jobs; a label whose vectors are all still buffered costs none.

**SVS and tiered SVS.** The plain index deletes the label and adds the given vectors as one batch,
which is the shape SVS wants anyway - `addVectorsImpl` preprocesses the whole buffer in one pass,
and a compressed index gets to see all of the vectors at once. The tiered one goes through its own
delete and add, so the journal that keeps an in-flight update job from carrying away a deleted
label, the training and update thresholds, and the write modes all apply as for any other write.

Nothing is compared against the stored vectors, so a quantized backend needs no special case.

Not atomic for the multi-value path: a query landing between the removal and the insertion sees the
label with none of its vectors - the same window the delete-then-insert it replaces would leave.
The single-value path has no such window, per the delegation above.

**Failures are reported, not thrown.** SVS signals a failure by throwing `ANNException`, and the
batch add has no status to return - it answers with the number of *new* labels. Since it is reached
across an `extern "C"` boundary, where an escaping exception is undefined behaviour, the SVS update
paths catch and report `Failed`. (`relabelVector` had the same problem and is fixed separately in
#1055, stacked on this branch, since it is code already on main.)

## Tests

- **plain HNSW and flat**, single and multi: a label replaced, grown, shrunk and emptied; the
  `MultiNotSupported` refusal leaving the index untouched; and the flat in-place path pinned by the
  id it keeps.
- **tiered, every ingestion phase**: still buffered, already indexed, split across the tiers, caught
  in the ingest window with a copy in each tier, a full flat buffer where inserts bypass the buffer,
  two updates before anything is ingested, in-place write mode, and SQ8 both during and after the
  accumulation phase (where the running sum that trains the quantizer has to stay right). The
  buffered one also pins the overwrite: the label's own id is kept and a second buffered label is
  not moved as collateral.
- **SVS**, plain single and multi and tiered: a label replaced, grown and emptied; the refusal for a
  single-value index; and for the tiered index the label buffered, then drained to the backend and
  updated there, then given several vectors at once where it is multi-value. Two SVS specifics the
  tests are written around: `indexSize` counts soft-deleted entries until a consolidation (so the
  counts subtract `getNumMarkedDeleted()`), and a compressed index clips a replacement outside its
  trained range (so the value-based assertions are gated on `isCompressed()`).
- **under load, per index**: `updateVectorsAsync` (tiered HNSW) replaces 500 labels 1000 times while
  workers ingest, repair and swap, for single and multi and for a swap threshold that keeps
  tombstones and one that disposes of them; `updateVectorsDuringUpdateJob` (tiered SVS) updates
  while the batch move and its journals are in flight, which is where a replaced vector could be
  carried into the backend behind its delete.
- **the three writes together** (`addRelabelAndUpdateTogetherUnderLoad`, one per tiered index): a
  fresh insert, a label moved onto a new doc id and a label's vectors replaced, interleaved on one
  index while the workers run - a label updated while its insert job is pending and then moved
  before that job runs, and the reverse order, both driven deliberately.
- **reading while updating** (`queryWhileUpdating`): searches queued into the same pool the writers
  use, which no previous test did for an update - only for insertion. It asserts no reply invents a
  label the index never held (what a use-after-free on a swapped id looks like), that the searches
  ran, and that the writes still landed; not *what* a search returns, since a query overlapping an
  update may see the old vectors, the new ones, or for a multi-value label neither.
- **the failure reporting** (`updateVectorsReportsAFailedWrite`, plain and tiered): a failed write
  reports `Failed` rather than a refusal that would claim the index was left alone, nothing escapes
  the API boundary, and the index still works afterwards. The tiered one runs in place and with more
  than one label in the backend - a multi-value update deletes before it adds, and with only that
  label present the backend would be left empty, after which the add buffers in the flat index and
  never reaches the backend the throw was armed on.
- **python flow tests**, compared against the add API: a tiered index updated *while ingestion is
  under way* must hold what the update put there, and match the recall of an index that was given
  those vectors to begin with (measured against the exact answer, so an approximation difference
  between the two graphs cannot fail it).

Two notes on how the tests assert, both learned the hard way here. Value checks ask the label for
its distance to the vector it should hold rather than running a k-nearest query: over a graph that
has just absorbed a few hundred deletes and moves, an approximate search does miss the vector
itself, which says nothing about whether the write landed. And the counts subtract what a soft
delete leaves behind, since `indexSize` counts those entries until a consolidation.

Mutation-verified: skipping the flat half of the tiered HNSW delete fails five tests, dropping the
flat index's in-place path fails its own, removing the single-value delegation fails the buffered
test on the collateral it would move, and leaving the replaced vectors behind in the tiered SVS
update fails both of its concurrent tests on the live count.

**A test-only seam.** `SVSIndexBase` gains `throwOnNextWriteForTest()` under `BUILD_TESTS` - beside
the `getLogger` already there for tests, a no-op in production builds. The reporting paths above are
otherwise unreachable: SVS throws only for the states the checks preceding those calls rule out, and
`add_points` grows its dataset rather than refusing a write.

## Coverage

Two added lines are uncovered, in `vec_sim_interface.h`: the `return VecSimUpdate_Unsupported;`
default. Every concrete index type in this repo overrides `updateVectors` and the rest of the
hierarchy is abstract, so no in-repo test can reach it; its consumer is out of repo - the disk index
in `vecsim_disk`, which inherits it until someone implements it. Covering it would mean a test-only
index type implementing the abstract base's whole pure-virtual surface to execute one `return`.

Local runs (ASAN): `test_hnsw` 365/365, `test_bruteforce` 71/71, `test_svs` 429 passed / 0 failed
(101 skipped - quantization variants this build does not include).

**The python flow tests are verified by CI rather than locally** - the bindings will not configure
on this machine (CMake 4.x has removed compatibility with the `cmake_minimum_required` of a vendored
dependency), which CI does not hit. The `basic tests` job runs `make flow_test`, and it passed on
532312b8; the non-contiguous case added since then gets its first run in this PR's latest CI.


src/python_bindings/bindings.cpp
```diff
@@ -220,6 +220,19 @@ class PyVecSimIndex {
...
+    // call with the whole new contents of the label.
+    VecSimUpdateCode updateVectors(labelType label, const py::object &input) {
+        py::array vectors(input);
+        // A single vector may be passed on its own rather than as a one-row array, so that the
+        // common case reads the way `add_vector` does.
+        const size_t n = vectors.ndim() > 1 ? vectors.shape(0) : (vectors.size() == 0 ? 0 : 1);
+        // An empty update has nothing to point at, and `data(0)` on an empty array is out of range.
+        const char *blobs = n == 0 ? nullptr : (const char *)vectors.data(0);
+        py::gil_scoped_release py_gil;
+        return VecSimIndex_UpdateVectors(index.get(), label, blobs, n);
```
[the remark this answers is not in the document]
nonirosenfeldredis: Fixed in c1cf9d93. `py::array::ensure(input, py::array::c_style)` copies the input only when it is not already packed, and leaves the dtype alone — `forcecast` would have been wrong here, since the blob's type has to keep matching the index's. A 3D array is now rejected as well, matching what `addVectorsParallel` does for non-2D input.

The single packed buffer is deliberate: the caller this API exists for holds its vectors that way (`fdata->vector` / `fdata->numVec` in RediSearch, already passed in that shape to `VectorIndex_HoldsVectors`), and SVS's batch add wants them packed too. So the per-row `data(ind)` approach the neighbouring helpers use isn't available here — hence the copy.

Regression case added in `test_update_vectors_multi`: a Fortran-ordered pair of vectors, asserting first that it really is not C-contiguous.

src/VecSim/vec_sim_common.h
```diff
@@ -142,6 +142,20 @@ typedef enum {
...
+// label and is the caller's doing, `Unsupported` says this index type does not implement the
+// operation at all - so the caller has to delete the label and insert the vectors itself - and
+// `Failed` is the index having tried and not managed it. The first two leave the index untouched;
+// `Failed` does not, since the label's vectors are removed before the new ones are stored, so the
+// label may be left holding nothing and the caller has to re-insert to get back to a known state.
+typedef enum {
+    VecSimUpdate_OK = VecSim_OK,    // for returning VecSim_OK as an enum value
+    VecSimUpdate_MultiNotSupported, // more than one vector was given to a single-value index
+    VecSimUpdate_Unsupported,       // this index type does not implement updating a label
+    VecSimUpdate_Failed // the index could not store them, and the label's contents are undefined
```
GuyAv46: What could cause this case? Will we handle it on Search side?

src/python_bindings/bindings.cpp
```diff
@@ -220,6 +220,31 @@ class PyVecSimIndex {
...
+        // would be read from the wrong offset. `addVector` and the parallel helpers get away
+        // without this because they index one row at a time, which numpy resolves through the
+        // array's own strides. The dtype is left alone: it has to keep matching the index's.
+        py::array vectors = py::array::ensure(input, py::array::c_style);
+        if (!vectors) {
+            throw std::runtime_error("Input vectors must be a numpy array");
+        }
+        if (vectors.ndim() > 2) {
+            throw std::runtime_error("Input vectors array must be 1D or 2D");
+        }
```
ofiryanai: should we check dimesions as well? not critical, I don't think the python bindings are of high usage

ofiryanai (approved):
LGTM
(didn't read all tests, good job covering all that. hope its not over extensive, lots LOC :) )

GuyAv46 (approved):

